PR #56, opened 2026-03-04
fix: remove npm workspaces to resolve Windows EISDIR error

atomantic:
## Summary

- Removed `workspaces` field from root `package.json` — npm workspaces always create symlinks which fail on Windows/exFAT with `EISDIR`. This was already fixed in v1.4.3 but workspaces were re-added at some point
- Replaced `-w client`/`-w server` workspace flags with `--prefix` equivalents for `build` and `test` scripts
- Removed `.npmrc` (`install-links=true`) and `--install-links` flags from `setup.ps1`/`update.ps1` — these were workarounds that didn't actually prevent the workspace symlink issue
- Simplified `update.ps1` cleanup to remove any stale workspace symlinks/copies unconditionally

## Test plan

- [ ] Run `.\update.ps1` on Windows — should complete without EISDIR errors
- [ ] Run `.\setup.ps1` on a fresh clone — should install all deps successfully
- [ ] Run `npm run build` from root — should build client via `--prefix`
- [ ] Run `npm test` from root — should run server tests via `--prefix`

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-# Copy workspace packages instead of symlinking (required for exFAT filesystems)
-install-links=true
--- a/package.json
+++ b/package.json
@@ -6,14 +6,10 @@
   "author": "Adam Eivy (@antic|@atomantic)",
   "license": "MIT",
   "type": "module",
-  "workspaces": [
-    "client",
-    "server"
-  ],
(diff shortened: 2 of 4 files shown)


atomantic:
Superseded by #57 — keeping workspaces, removing install-links instead

PR #188, opened 2026-05-01
fix(video-gen): I2V source preservation + FFLF + extract-last-frame fixes + media model registry

atomantic:
## Summary

- **I2V on macOS now actually preserves the source image as the first frame.** Root cause was `mlx-video-with-audio` 0.1.33 having a known bug ("VAE encoder failed to load for split-format unified models") fixed in 0.1.35. PortOS's setup script was also installing the wrong PyPI package (`mlx_video` is unrelated; the LTX backend is `mlx-video-with-audio`) — both provide an `import mlx_video` module so the mistake was silent. Verified pixel-perfect first-frame preservation end-to-end.
- **Adjacent bugs**: FFLF passed video frame index where mlx_video wants a latent index → silent no-output. `extractLastFrame`'s `-sseof -0.1` window was too narrow for audio-having videos → ffmpeg returned 0 frames while logging success. Both fixed; added `existsSync` verification so success logs can't lie.
- **Last-frame extraction now caches** to `data/images/lastframe-<id>.png`. Repeated "Continue" clicks reuse the file instead of re-spawning ffmpeg.
- **Media model registry** at `data/media-models.json` (new) — single source of truth for video models, image models, and the LTX text encoder. Replaces hardcoded catalogs in three files. Adds `gemma-bf16` (now default — best output quality) and auto-detects LM Studio's `gemma-qat-4bit` if installed.
- **Borrowed from [mrbizarro/phosphene](https://github.com/mrbizarro/phosphene)**: `+faststart` mp4 remux (gallery tiles render poster frames immediately) and `caffeinate -i -w` (laptop sleep doesn't kill 90s+ renders).
- **Other improvements**: raw stderr/stdout logging from python child, `imageStrength` threaded through API, form fields stay editable during a render, 1:1 resolutions, image-upload preview + larger reference-image previews, delete operations log.

## Test plan

- [x] Verify first-frame preservation: pick a photographic image → Generate → first frame matches source.
- [x] Verify thematic continuation: "Continue" on a video → new render's general scene matches source theme.
- [x] Verify "Continue" caches: click "Continue" twice on the same video — second click should be instant (no ffmpeg spawn in logs).
- [x] Verify image-upload preview renders before submit.
- [x] Verify gallery thumbs render their poster frame on first page load (no black flash).
- [x] Verify form fields are editable while a render is in flight; "Add to queue" works.
- [x] Verify default text encoder is bf16 (`models--mlx-community--gemma-3-12b-it-bf16` in TEXT_ENCODER:PATH log line).
- [x] Verify all 2875 server tests pass: `cd server && npm test -- --run`.


atomantic:
Addressed all 7 Copilot review comments in `ec64c398`:

| # | File:Line | Issue | Fix |
|---|---|---|---|
| 1 | `mediaModels.js:55` | Seeded default mismatched changelog (`gemma-4bit` vs documented `gemma-bf16`) | Seed now defaults to `gemma-bf16`; relabeled `gemma-4bit` as "(smallest)" and `gemma-bf16` as "(default, best quality)" |
| 2 | `mediaModels.js:78` | `JSON.parse` could crash server boot on malformed user edit | Wrapped in try/catch with `DEFAULT_REGISTRY` fallback + `⚠️ Failed to parse...` warning log |
| 3 | `mediaModels.js:117` | Missing test coverage | Added 9 unit tests in `lib/mediaModels.test.js` covering seed-on-first-load, per-platform `broken` filtering, `~/` expansion, `localPath` preference, malformed-JSON fallback, caching, and platform-default lookup |
| 4 | `mediaModels.js:57` | **Real bug** — `path.join(homedir(), p.slice(1))` discards homedir on `~/...` because the second segment starts with `/`. Broke LM Studio auto-detect. | Applied the suggested fix; pinned by a test that asserts result starts with `homedir()` |
| 5 | `videoGen/local.js:130` | `rename()` errors silently swallowed → `.fs.mp4` leak | Wrapped in try/catch: clean up tmp + log warning |
| 6 | `videoGen/local.js:31` | Comment said "Map" but value is a plain object | Corrected to "plain object" |
| 7 | `videoGen/local.js:405` | `item.id` interpolated into a filesystem path without validation | Validate `item.id` matches `/^[a-f0-9-]{36}$/i` (UUID) before using; throws 400 otherwise |

Bonus: registry file path is now overridable via `PORTOS_MEDIA_MODELS_FILE` env var (used by the new tests; also useful for non-standard deployments).

**Test count: 2875 → 2884** (all passing).

PR #318, opened 2026-05-18
feat([coverautofiler-dispatcher-redundant-getseries]): pass preloaded series through dispatcher

atomantic:
## Summary

`server/services/pipeline/coverUniverseFiler.js`'s dispatcher (`fileCoverIntoAutoCollection`) already fetches the series record purely to branch on `series.universeId`. Previously the two leaves (`fileCoverIntoUniverseCollection`, `fileCoverIntoSeriesCollection`) each re-fetched it on entry, then performed a second mid-flight re-read for race-detection. Threading the dispatcher's record through as an optional `_preloadedSeries` arg lets each leaf skip its initial `getSeries` while preserving the race-detection re-read.

**Net effect per cover completion:** 2 series.json reads instead of 3.

## Behavior preserved

- **Direct callers of either leaf** (no preload) keep the old shape — the param defaults to undefined and the `??` falls through to the original `getSeries(...).catch(() => null)`.
- **Race-detection contracts** are intact on both branches — the universe leaf still re-reads the series after `getUniverse` returns to catch a mid-flight re-link, and the series leaf still re-reads after `addItem` for delete/link-gain recovery.
- **Series leaf re-route → universe leaf** intentionally does NOT forward the preload, because the `liveSeries` we used to detect the link could be stale relative to a concurrent unlink. Letting the universe leaf re-read keeps its race-detection contract intact. Documented in both functions' doc-comments.

## Test plan

- [x] +2 vitest cases in `coverUniverseFiler.test.js` spy on `seriesSvc.getSeries` and pin the post-optimization call count (2 total) on both branches.
- [x] All 21 cover-filer tests pass.
- [x] All 349 pipeline-service tests pass.

## Deferred follow-up

Captured in PLAN.md as `[coverautofiler-preload-universe-record]`: mirror the same pattern for `universeSvc.getUniverse` (called twice in the universe leaf, same shape). Deferred because no current direct caller has a fresh universe to feed.


atomantic:
Addressed all three Copilot findings (force-pushed `6354d4ab`):

1. **Series-leaf race regression** — dropped the preload from `fileCoverIntoSeriesCollection` entirely. The series leaf's first `getSeries` is load-bearing for the early re-route check; preloading it would let a newly-linked series get filed into the per-series collection and merely orphaned by recovery rather than routed into the universe collection. Optimization now applies only to the (dominant) universe path: 3 → 2 `getSeries` calls per cover completion. The previously-shared scope (both leaves, 2 calls each) shrank to one leaf because the race semantics of the series leaf are non-negotiable.
2. **Typo `live liveSeries`** — fixed in the series-leaf doc-comment.
3. **Spy leak on failed assertion** — restructured both new tests to capture-restore-assert (`const callCount = spy.mock.calls.length; spy.mockRestore(); expect(callCount).toBe(...)`), so a failing expect doesn't leave the spy active for later tests.

Also kept the universeless-path test as a regression-pin asserting **3** calls so a future "let's also preload the series leaf" refactor fails loud.

PR #330, opened 2026-05-19
fix: contain async errors, cap unbounded maps, parallelize hot paths

atomantic:
## Summary

Phase 4 of the **Better Audit — 2026-05-19**. Six runtime / resource fixes:

- **[CRITICAL] `agentCliSpawning.js`** — wrap stdout + stderr `async (data)` callbacks in try/catch (CLAUDE.md PTY/child-process rule — uncaught throws there crash the Node process). Batch per-line state writes via `appendAgentOutputLines` instead of awaiting `appendAgentOutput` per line — eliminates N JSON-state load+save round-trips per chunk under high CLI output rates. Companion regression test verified to fail when the try/catch is removed.
- **[HIGH] `voice/llm.js` + `voice/bootstrap.js`** — add `{ signal: AbortSignal.timeout(5000) }` so an unresponsive LM Studio doesn't hang the voice pipeline indefinitely.
- **[HIGH] `toolStateMachine.js`** — cap the module-level `executions` Map at 1000 entries via insertion-order eviction; previously orphaned entries on crash because eviction only ran via the success-path finalize 60s timer.
- **[MEDIUM] `sharing/exporter.js`** — replace serial `for…of (await exportSeries|Universe)` with `Promise.all(ids.map(...))`.
- **[MEDIUM] `agentTuiSpawning.js`** — replace silent `.catch(() => {})` on `doneSentinelTimer` `readFile` with a logged catch; wrap the `setInterval` body in try/catch; guard `.finally(() => finish().catch(...))` against synchronous throws.

## Files Modified
- `server/services/agentCliSpawning.js` (+ regression test)
- `server/services/agentTuiSpawning.js`
- `server/services/voice/llm.js`, `server/services/voice/bootstrap.js`
- `server/services/toolStateMachine.js`
- `server/services/sharing/exporter.js`

## Merge Order
Independent.

### DIFF
--- a/server/services/agentCliSpawning.js
+++ b/server/services/agentCliSpawning.js
@@ -11,7 +11,7 @@ import { existsSync } from 'fs';
 import { spawn } from 'child_process';
 import { homedir } from 'os';
 import { cosEvents, emitLog } from './cosEvents.js';
-import { updateAgent, completeAgent, appendAgentOutput } from './cosAgents.js';
+import { updateAgent, completeAgent, appendAgentOutput, appendAgentOutputLines } from './cosAgents.js';
 import { registerSpawnedAgent, unregisterSpawnedAgent } from './agents.js';
 import { release } from './executionLanes.js';
 import { completeExecution, errorExecution } from './toolStateMachine.js';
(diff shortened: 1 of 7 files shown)


atomantic:
@copilot review

PR #338, opened 2026-05-19
refactor: better-audit residue bundle — formatters, feeds sort, error codes, magic numbers

atomantic:
## Summary

Bundles four PLAN.md **Better-audit residue** items into one PR — three shipped as scoped fixes, one was rescoped after the underlying premise eroded:

- **`[medium-client-4-components-still-redefine]`** — added `formatTimecode`, `formatDurationSec`, `formatDateShort` to `client/src/utils/formatters.js`; migrated `VideoTimelineEditor.jsx`, `VideoTimeline.jsx`, `ImportTab.jsx` off their local copies. `MortalLoomTab.jsx` was already using the shared `formatBytes` (the PLAN entry was stale on that one).
- **`[medium-perf-feeds-js-getitems-303-319-full-sort]`** — `feeds.js` now maintains a newest-first sort invariant on writes (`addFeed`, `refreshFeed`, end-of-batch in `refreshAllFeeds`); `getItems` drops the per-call sort. New `ensureItemsSorted()` runs once on first read with an `isSortedNewestFirst()` guard so an already-sorted file isn't re-saved on every boot.
- **`[medium-code-magic-numbers-in-cos-js-166-357]`** — replaced brittle `err.message.startsWith('unknown piper voice:')` (`routes/voice.js`) and `err.message.includes('not initialized')` (`services/visionTest.js`) with typed `ServerError` codes. `services/voice/tts.js` now throws `ServerError({ status: 400, code: 'UNKNOWN_VOICE' })`; `services/providers.js`'s nine duplicate inline `if (!aiToolkitInstance) throw …` checks consolidated into one `requireToolkit()` helper throwing `ServerError({ status: 503, code: 'AI_TOOLKIT_NOT_INITIALIZED' })`. Both bubble through `asyncHandler` automatically — the inline try/catch in voice.js dropped entirely. visionTest.js's misleading "Provider not found" mapping replaced with "AI provider service is still initializing" when the toolkit isn't warmed. Named time constants in `cos.js` (`CD_RECOVERY_BOOT_TIMEOUT_MS`, `POST_STARTUP_QUEUE_DELAY_MS`, `RECENT_COMPLETION_GRACE_MS`) and `lmStudioManager.js` (`DEFAULT_REQUEST_TIMEOUT_MS`, `AVAILABILITY_PROBE_TIMEOUT_MS`).
- **`[low-client-extract-shared-usepopoverposition-hook]`** — **skipped, premise eroded.** Of the 4 components originally named, `VisualStylePicker` no longer exists and `AddToCollectionMenu` / `BulkTargetPicker` no longer use the portal-positioning pattern. The shape lives in 2 components today (`ThemeSwitcher` + `CollectionPickerShell`) — abstraction not worth the indirection at 2 callers. PLAN.md entry rewritten with current state.

Deferred to PLAN.md as **`[extract-requiretoolkit-pattern-to-runner-js]`** — `runner.js` has 13 more sites of the same inlined toolkit-init check that should adopt the new helper. Surfaced by `/simplify` and left out of this PR to keep scope tight.

Also fixed a pre-existing test-config bug rebased in from main: renamed `client/src/hooks/useRowDraft.test.js` → `.test.jsx` so it runs in the jsdom-enabled client vitest runner (the `.js` extension was routing it to the server runner where `document` is undefined).

## Test plan

- [x] All 5741 server tests pass (`cd server && npm test`)
- [x] All 112 client tests pass (`cd client && npm test -- --run`)
- [x] `/simplify` review pass — fixed `ensureItemsSorted` no-op write, migrated tts.js + providers.js to `ServerError` (the canonical pattern), updated voice.test.js to mock ServerError directly, dropped the now-obsolete voice.js try/catch since asyncHandler maps ServerError status automatically.


atomantic:
**Codex review pass 1** — 1 MED finding: integration test `visionTest.integration.test.js:197` loosened to accept either `/not found|still initializing/`, weakening the "Provider not found" contract assertion.

**Resolution: deferred, not blocking.** The strict "Provider not found" contract is already pinned in the *unit* test at `server/services/visionTest.test.js:50` ("should return error when provider not found"). The integration test was previously passing the `not found` regex by accident — it always hit the `AI_TOOLKIT_NOT_INITIALIZED` path because the toolkit isn't initialized in this test file's setup. My change just made the assertion honestly reflect what the test actually exercises. Splitting it into two tests (one with toolkit warmed, one without) would require new test scaffolding for a contract that's already covered elsewhere. Not worth the diff in this PR.

Loop converged in 1 iteration. All 5741 server tests + 112 client tests green.

PR #345, opened 2026-05-20
feat([sharing-exporter-cache-sourcefile-hash-by-mtime]): cache exporter asset hashes by sourcePath+mtime+size

atomantic:
## Summary

- `server/services/sharing/exporter.js#copyAssetIfPresent` no longer re-hashes every referenced asset on every export. A new per-bucket `<bucket>/assets/blobs/.index.json` sidecar maps `<sourcePath>:<mtimeMs>:<size> → <hash>` so a subscription re-export of a 200MB series skips both `sha256File` (multi-GB stream read) and the redundant `copyFile` when the cache hit's blob is still on disk. mtime change invalidates because it's part of the key.
- `withAssetHashCache(bucketPath, fn)` wraps the load/mutate/save lifetime for the three top-level exporters. Short-circuits the write when no new entries were added (pure cache-hit re-exports do zero I/O on the sidecar). Serializes the merge-on-save step through a per-bucket write tail (same pattern as `issueWriteTail`) so parallel exporters — e.g. `exportByKind` fanning out multiple `exportSeries` against the same bucket — accumulate entries instead of clobbering each other.
- `bucketBlobIndexPath` lives in `buckets.js` alongside the other `bucketBlob*` path helpers so the storage-layout invariant documented at `buckets.js:80-91` keeps holding.
- `isHexHash` validates cached values before trusting them so a hand-edited or torn `.index.json` falls back to recompute rather than poisoning `bucketBlobPath`.

## Test plan

- [x] Existing 30 sharing round-trip tests still pass (content-addressed dedup assertion now filters dotfiles so the sidecar isn't counted as a blob)
- [x] New test: index file written with documented key shape `(sourcePath:mtimeMs:size → hash)`
- [x] New test: unchanged re-export does NOT rewrite the index (mtime equality verified)
- [x] New test: mtime-only bump (same bytes) records a new key with the unchanged hash
- [x] New test: content mutation records a fresh hash and lands a new content-addressed blob
- [x] New test: parallel `Promise.all` exports against the same bucket merge entries through the write tail (both keys land in persisted index)
- [x] All 5757 server tests + 121 client tests pass

Closes the PLAN.md item `[sharing-exporter-cache-sourcefile-hash-by-mtime]`. /simplify findings folded in pre-commit: hoisted path helper to `buckets.js`, wrapped the 3× load/save scaffolding into `withAssetHashCache`, added dirty-flag short-circuit, added per-bucket write tail for parallel safety. Deferred one cross-cutting refactor to PLAN.md as `[export-isplainobject-helper-from-server-lib-objects]` (14 inlined `isPlainObject` shape-checks).


atomantic:
Closing — work landed directly on main via 7a07fb11 (feat) + 397966c3 (docs). PR commits are content-equivalent under different hashes.

PR #351, opened 2026-05-20
fix(universe-builder): reopen Universe Name dropdown shows other universes instead of "No matches"

atomantic:
## Summary

- The Universe Builder header's autocomplete combobox seeds its input with the currently selected universe's name. Clicking the chevron to reopen the list applied that name as a substring filter and then excluded the selected universe itself — so the dropdown always showed "No matches" and there was no way to switch to another universe without first manually clearing the field.
- Fix: when the trimmed lower-case input equals the selected universe's name, treat it as a *browse* intent and skip the substring filter so the user sees every other universe. As soon as they type something different, normal filtering resumes.
- Added two `UniverseSelector` tests covering the browse-on-reopen case and the resume-filtering-on-type case.

## Test plan

- [x] `cd client && npm test -- --run` — 143/143 passing
- [ ] Manually: select a universe, click the chevron — verify other universes appear
- [ ] Manually: with a universe selected, type a partial name — verify the list filters
- [ ] Manually: with a universe selected, type the same name verbatim again — verify no spurious "Create" row appears

### DIFF
--- a/client/src/pages/UniverseBuilder.jsx
+++ b/client/src/pages/UniverseBuilder.jsx
@@ -383,7 +383,7 @@ const LIST_ID = 'universe-selector-list';
 const OPTION_ID_PREFIX = 'universe-option-';
 const CREATE_OPTION_ID = 'universe-option-create';
 
-function UniverseSelector({ universes, selectedId, value, onChange, onPick, onCreate, busy }) {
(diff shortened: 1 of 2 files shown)


atomantic:
Superseded by main: the fix duplicated 3aa4ae1a, and the two regression tests have been salvaged to main as 2cf57457.

PR #988, opened 2026-06-06
fix([issue-978]): defer active-agent bind until idle-review task exists

### LINKED ISSUE #978: Idle-review path leaves app marked active when the task generator returns null
## Problem

`generateIdleReviewTask` (server/services/cosTaskGenerator.js) calls `markAppReviewStarted(app.id, …)` *before* invoking the per-app task generator, then returns whatever the generator produces. When the generator returns `null` — e.g. `plan-task` with no claimable PLAN item, a precondition skip, `reference-watch` with no new commits, or `pr-watcher` with no new PRs — no agent is created, so `activeAgentId` is never cleared and the app reads as "in review" until stale-agent cleanup or a restart.

This is **pre-existing** (any null-returning task type already triggered it via the old generic generator), but PR #977 (pr-watcher) makes it more visible: the idle path now correctly returns `null` on no-op watcher polls instead of spawning a broken task, so the marker is left dangling more often.

## Why it wasn't fixed in #977

There's real design tension: the eager `markAppReviewStarted` also advances the per-app review cooldown, which prevents the same app from being re-picked every idle tick (a re-pick storm). Naively deferring the mark until a task exists reintroduces that storm. A correct fix likely needs to split "advance cooldown" from "set activeAgentId" so a no-op poll advances the cooldown without leaving a phantom active agent. That touches shared idle-review machinery for all task types, so it belongs in its own change rather than the pr-watcher feature PR.

## Suggested fix

In `generateIdleReviewTask`, advance the app review cooldown on a `null` result but do **not** leave `activeAgentId` set — i.e. either revert the activity marker when the delegate returns `null`, or split `markAppReviewStarted` into a cooldown-only stamp plus an agent-binding step and only bind when a task is created. Note the queue path (`queueEligibleImprovementTasks`, the primary scheduler) does not have this issue — it never marks review-started before generating.

atomantic:
Closes #978

## Problem

`generateIdleReviewTask` called `markAppReviewStarted(app.id, …)` *before* invoking the per-app task generator. `markAppReviewStarted` did two things at once: advanced the per-app review cooldown (`lastReviewedAt`) **and** bound `activeAgentId`. When the generator returned `null` — no claimable PLAN item, a `reference-watch`/`pr-watcher` no-op poll, a precondition skip — no agent was created, so `activeAgentId` stayed set and the app read as "in review" until stale-agent cleanup or a restart cleared it.

The design tension noted in the issue: advancing the cooldown eagerly is *necessary* to prevent a re-pick storm (the same app being re-selected every idle tick). The fix can't simply move the whole mark after generation.

## Fix

Split `markAppReviewStarted` into two single-purpose functions in `appActivity.js`:

- `markAppReviewCooldown(appId)` — stamps `lastReviewedAt` only (the re-pick-storm guard). Runs eagerly, before generation.
- `bindAppReviewAgent(appId, agentId)` — sets `activeAgentId` only. Runs *after* a task is confirmed to exist.

Applied the split to all three "mark-before-generate" call sites that shared the latent bug:
- `generateIdleReviewTask` (the path the issue calls out)
- the on-demand loop in `cosTaskGenerator.evaluateTasks`
- the on-demand loop in `cos.dequeueNextTask`

A no-op poll now advances the cooldown (no re-pick storm) without stranding a phantom active marker. The now-dead conflated `markAppReviewStarted` was removed (no remaining callers).

The queue path (`queueEligibleImprovementTasks`) was already correct — it never marked review-started before generating — and is unchanged.

## Tests

- New `appActivity.test.js`: pins the split semantics — cooldown stamp puts the app on cooldown without binding an agent; a null-task poll leaves `activeAgentId` null; `bindAppReviewAgent` binds after a task and preserves the earlier cooldown stamp.
- `cos.test.js`: added a source-level guard that both on-demand loops use `markAppReviewCooldown` (not the conflated call) and only `bindAppReviewAgent` inside an `if (task)` guard; updated the existing dedupe guard's wording.
- Full server suite green (10,910 passed).

### DIFF
--- a/.changelog/NEXT.md
+++ b/.changelog/NEXT.md
@@ -13,6 +13,8 @@
 
 ## Fixed
 
+- **[issue-978] Apps no longer get stuck reading "in review" after a no-op idle check** — when the Chief of Staff's idle review picked an app but had nothing to actually do (no claimable plan item, no new commits or PRs to watch), the app could stay marked as under review until a restart or cleanup pass cleared it. Idle checks that find no work now release the app right away while still spacing out how often it's rechecked.
+
 - **[issue-984] Chief of Staff daily action cap holds firm when several jobs come due at once** — scheduled jobs that fired in the same instant could each slip past a small daily action cap before the others were counted, letting the cap be exceeded by one. The cap is now enforced the moment each job is admitted, so simultaneously-due jobs can no longer overshoot it.
 
 - **[issue-968] A PM2 hiccup no longer makes running apps look offline** — when PortOS briefly can't read process state, affected apps now show "status unavailable" instead of being silently reported as stopped. The Apps list and detail pages replace the (misleading) Start button with a refresh-to-retry control, the dashboard counts these separately, the system health page flags the degraded read, and CyberCity no longer rains on apps whose status simply couldn't be read.
--- a/server/services/appActivity.js
+++ b/server/services/appActivity.js
@@ -92,7 +92,39 @@ export async function startAppCooldown(appId, cooldownMs) {
 }
 
 /**
- * Mark an app review as started
+ * Advance an app's review cooldown WITHOUT binding an active agent.
+ *
+ * Stamps `lastReviewedAt` (which `isAppActivityOnCooldown` reads) so the app
+ * isn't re-picked on the very next idle tick — the re-pick-storm guard. This is
+ * the half of the old `markAppReviewStarted` that must run *before* the per-app
+ * task generator, since a no-op poll (task generator returns `null`) still has
+ * to advance the cooldown. Crucially it does NOT touch `activeAgentId`, so a
+ * null result can't leave a phantom "in review" marker. See `bindAppReviewAgent`
+ * for the agent-binding half, called only once a task actually exists.
+ */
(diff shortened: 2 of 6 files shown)


atomantic:
## Review loop (claude + codex) — converged CLEAN

- **claude** (independent sub-agent review): CLEAN — verified no other callers relied on the old conflated `markAppReviewStarted`, the `if (task)`/`if (idleTask)` guards cover all null-return cases, the two on-demand loops are now structurally identical, and `markIdleReviewStarted()` (global telemetry) is disjoint from the per-app marker.
- **codex**: raised one HIGH finding — `bindAppReviewAgent` writes a synthetic marker *before* `spawnAgentForTask` runs, so a spawn failure (vs. a null task) can still strand the marker. On review this is **pre-existing** (pre-fix the synthetic id was bound even earlier) and **out of #978's scope** (which targets the task-generator-returns-null case). It's a larger spawn-lifecycle fix with its own design/test surface, so it's tracked separately as **#989** (`plan`+`future`). codex accepted the scope argument on the second pass and returned CLEAN.

No in-scope correctness findings remain.

PR #1067, opened 2026-06-09
feat(image-gen): remove visible Gemini sparkle watermark

atomantic:
## What

Adds a way to remove the visible **Gemini / \"nano-banana\" sparkle** — the small four-pointed white star Gemini 2.5 Flash Image stamps into the bottom-right corner of every render.

This is the **visible-watermark** counterpart to the existing **Regenerate** action (which defeats Gemini's *invisible* SynthID signal via a FLUX img2img round-trip). It's the gap explicitly deferred in the SynthID evaluation plan (`docs/plans/2026-06-05-synthid-removal-eval.md` → \"Considered, NOT pursued\").

## How

A pure-CPU (sharp) **detect-and-inpaint** pass — no GPU, no AI model, no network — so it works on every install:

1. Search the bottom-right ROI where Gemini always places the mark.
2. Estimate the local background with a large-radius blur; mark pixels that are brighter-than-background **and** low-saturation **and** bright (the additive white star), rejecting colored highlights.
3. Group candidates into connected components; pick the compact, ~square, ~⅓-filled blob nearest the corner (the star shape). No match → no-op.
4. Dilate + feather the mask, diffusion-inpaint the region from the surrounding background, and alpha-blend it back.

### Surface
- **`server/lib/geminiSparkle.js`** — `removeGeminiSparkle(buffer)` + pure, unit-tested helpers (`resolveSparkleRoi`, `buildCandidateMask`, `connectedComponents`, `selectSparkleComponent`). Self-contained (sharp + node builtins), barrel + README registered.
- **`POST /image-gen/:filename/remove-watermark`** — synchronous like Clean; writes a `_no-watermark.png` variant + sidecar, auto-files into the source's collections, groups under the original via `cleanedFrom`. Returns `{ removed: false }` (source untouched) when no sparkle is found.
- **Lightbox \"De-sparkle\" action** wired through every gallery surface (ImageGen, MediaHistory, MediaCollectionDetail, UniverseBuilder) via the shared `useMediaPreviewActions` hook. Honest lineage label (\"Watermark removed from …\").

## Verification
- Validated end-to-end on a real Gemini render (`data/screenshots/IMG_1054.png`): the 48×48 sparkle is detected and erased in ~75ms, blends seamlessly at normal scale, and the pass is **idempotent** (a second run correctly finds nothing).
- Unit tests (`geminiSparkle.test.js`, 12) cover the pure helpers, end-to-end detect+inpaint, the no-sparkle no-op, and empty/garbage-buffer degradation. Route tests (`imageGen.test.js`, +3) cover 404, the `removed:false` no-write path, and a successful variant write.
- Touched server + client suites green; client builds clean.


atomantic:
Closing as a duplicate. This PR and #1071 were competing implementations of the same feature from the same CoS task (`cos/task-mq65xtyk`, different agents); #1071 was the one merged (commit fd4be946f).

I reviewed this branch for any non-duplicate work worth salvaging. The server engine here (`geminiSparkle.js` — a detection-based connected-components inpaint with a `{ removed: false }` no-op) is an alternative take on the already-shipped `imageWatermark.js` (fixed-corner harmonic fill); re-landing it would churn the merged feature for no functional gain, so it stays out.

The one thing this branch caught that the merged code was genuinely missing: `normalizeImage()` wasn't carrying the `watermarkRemoved` flag through, so the 'Watermark removed from …' lineage label and variant grouping never fired through the real gallery path (every surface normalizes images before the lightbox). I've salvaged that fix as a standalone PR against current `main`: #1073.

PR #1436, opened 2026-06-18
Add Tribe relationship manager

atomantic:
## Summary
- Add a Tribe page/tool for managing relationship rings, care cadence, focus lists, and touchpoints.
- Back Tribe data with Postgres tables for people, touchpoints, and memory links, with boot-time schema upgrades and init-db parity.
- Add `/api/tribe` routes/services, Brain memory linking, and calendar-ready touchpoint references for future calendar import/scheduling work.

## Verification
- `npm test -- tribe.test.js navManifest.test.js db.catalogDdlParity.test.js` from `server/`
- `npx eslint src/pages/Tribe.jsx src/services/apiTribe.js src/services/api.js src/App.jsx src/components/Layout.jsx` from `client/`
- `npm run build` from `client/`
- `git diff --check origin/main...HEAD`

## Local Review
- Reviewed all 12 changed files against `origin/main...HEAD`.
- Fixed review findings before push: guarded legacy storage access, avoided optimistic UI updates on failed API calls, and removed nested interactive controls in Tribe cards.


client/src/pages/Tribe.jsx
```diff
@@ -0,0 +1,922 @@
...
+    setSaving(true);
+    const saved = draft.id
+      ? await api.updateTribePerson(draft.id, normalized).catch((err) => {
+          toast.error(err.message || 'Failed to save relationship');
+          return null;
+        })
+      : await api.createTribePerson(normalized).catch((err) => {
+          toast.error(err.message || 'Failed to save relationship');
+          return null;
+        });
```
atomantic, line 747: **[CRITICAL]** Every edit of an existing person 400s. `normalized` is `{ ...draft }`, and `draft` comes from `selectContact` which spreads the full server person — so it carries `id`. The PUT body therefore includes `id`, but `personUpdateSchema` declares `id: z.never().optional()` (`server/routes/tribe.js:29`), which rejects any present `id` (`Invalid input: expected never, received string`). I confirmed this empirically: `personUpdateSchema.safeParse({ id: "<uuid>", name: ... })` → `success: false`. The route test only sends a partial body without `id` (`tribe.test.js:66`), so it never exercises the real client payload. Strip `id` (and the server-managed/derived fields) before sending:

```suggestion
    const { id: _omitId, touchpointCount, linkedMemoryCount, createdAt, updatedAt, ...payload } = normalized;
    const saved = draft.id
      ? await api.updateTribePerson(draft.id, payload).catch((err) => {
          toast.error(err.message || 'Failed to save relationship');
          return null;
        })
      : await api.createTribePerson(payload).catch((err) => {
          toast.error(err.message || 'Failed to save relationship');
          return null;
        });
```

Also add a route test that PUTs a body containing `id` so this contract is pinned.

client/src/pages/Tribe.jsx
```diff
@@ -0,0 +1,922 @@
...
+      onClick={() => setDraft(emptyDraft())}
+      className="inline-flex items-center gap-2 rounded border border-port-border px-3 py-2 text-sm text-gray-300 hover:bg-port-border/40 hover:text-white"
+    >
+      <Plus size={15} aria-hidden="true" />
+      Add
+    </button>
+  );
+
+  return (
+    <div className="flex h-full flex-col">
```
atomantic, line 805: **[CRITICAL]** Layout scroll-mode mismatch. This page is a full-bleed page that owns its own internal scroll (`flex h-full flex-col` here + `<main className="flex-1 overflow-auto p-4">` at line 792), but `/tribe` is **not** in Layout's `isFullWidth` list (`client/src/components/Layout.jsx:1106`). Per the CLAUDE.md "New page → pick the right Layout scroll mode" convention, an unmatched route gets the default `overflow-auto p-4 md:p-6` main — so you get double padding and a nested scroll container, and the inner `flex-1 overflow-auto` can clip below the fold. Fix (out of this file's diff): add `location.pathname === '/tribe' ||` to the `isFullWidth` chain in `Layout.jsx`. Alternatively, restructure this page as a plain scrolling `<div>` (drop `h-full` + the internal `<main overflow-auto>`) and stay out of the list.

client/src/pages/Tribe.jsx
```diff
@@ -0,0 +1,922 @@
...
+  const loadContacts = async () => {
+    setLoading(true);
+    const result = await api.getTribePeople({ silent: true }).catch((err) => {
+      toast.error(err.message || 'Failed to load Tribe');
+      return { people: [] };
+    });
+    let people = result?.people || [];
+
+    const legacy = getLegacyContacts();
+    if (people.length === 0 && legacy.length > 0) {
```
atomantic: **[IMPROVEMENT]** The legacy localStorage→Postgres import fires on a *failed* fetch, not just an empty list. The catch returns `{ people: [] }`, so a transient server error makes `people.length === 0` true; if old localStorage contacts still exist, they'll be re-imported even though Postgres may actually hold people — duplicating data once the server recovers. Distinguish "fetch failed" from "fetched, genuinely empty" with a sentinel (the absent-vs-empty rule in CLAUDE.md):

```suggestion
    const result = await api.getTribePeople({ silent: true }).catch((err) => {
      toast.error(err.message || 'Failed to load Tribe');
      return null;
    });
    const fetchFailed = result === null;
    let people = result?.people || [];

    const legacy = getLegacyContacts();
    if (!fetchFailed && people.length === 0 && legacy.length > 0) {
```

atomantic (commented):
Reviewed by `/do:review` (PR mode, 5-agent self-review: runtime, quality, security, cross-file state, cross-file contract) — **2 critical, 5 improvements, 1 nit**. Author == reviewer, so this is posted as a comment (GitHub forbids requesting changes on your own PR).

### Highlights (must-fix)
- **`client/src/pages/Tribe.jsx:716`** — *Editing an existing person 400s.* The PUT body carries `id` (via `{ ...draft }`), and `personUpdateSchema` declares `id: z.never()`, which rejects it. Verified empirically. The whole edit flow is broken and untested.
- **`client/src/pages/Tribe.jsx:782`** — `/tribe` is missing from `Layout.jsx`'s `isFullWidth` list while the page owns an internal scroll + `h-full`, so it gets double padding / clips below the fold (CLAUDE.md scroll-mode convention).

### Out-of-diff observations
- **Missing service tests.** No `server/services/tribe.test.js`. The branchy SQL builders (`listPeople` param-index assembly), `normalizeTags`, `rowToPerson`, and the FK-violation→404 mapping are untested; the route tests mock the service entirely. This is the layer most exposed to schema↔JS drift — worth a DB-backed `*.db.test.js` gated like the others.
- **DDL parity isn't test-covered for the new tables.** `db.catalogDdlParity.test.js` only parity-locks audit *triggers* (which do match here) — the tribe table *column* DDL in `server/lib/db.js` vs `server/scripts/init-db.sql` is not in any parity list. They're identical today, but column drift between the two files would go uncaught (same tolerated gap the `memories` tables note).
- **Federation/design question.** The `tribe_*` tables carry no `sync_sequence`/sync columns, so Tribe data is machine-local and won't appear on a user's other federated machines. Confirm that's intended (relationship data is local-only) vs. a gap — a user with 2 peers would likely expect their people list on both.
- **Minor: timezone skew.** `logTouch` sets the optimistic `lastContact` from `todayISO()` (UTC, `Tribe.jsx:748/62`) while cadence math (`daysBetween`) parses dates in local time — off-by-a-day near midnight UTC. Also the server's `GREATEST(...)` last-contact merge (`services/tribe.js:252`) can keep an existing date the optimistic UI just overwrote with today; reconciling from the server row (or via the socket event above) would make it robust.
- **Nit: nav id.** `navManifest.js:91` uses `nav.tribe` rather than the documented `nav.<section>.<slug>` (`nav.brain.tribe`); it matches existing inconsistent Brain siblings, so low priority.

### Coherence check
PR claims verified against the code: Postgres tables for people/touchpoints/memory-links with boot-time `ensureSchema` upgrades **and** `init-db.sql` parity ✓; audit triggers added in both files ✓; `/api/tribe` routes + service + Brain memory linking + calendar-ready touchpoint references ✓; nav manifest / sidebar / route / api barrel wiring all present ✓.

### Verified clean (traced, no action needed)
- **No SQL injection** — `listPeople` concatenates only `$N` placeholder tokens; every user value (incl. the `%search%` wildcard) goes through the parameter array.
- **No prototype pollution** — `updatePerson`'s `{ ...current, ...updates }` consumes zod-validated `updates` (`z.object` strips unknown keys, incl. `__proto__`).
- **No path traversal / SSRF** in `createCalendarTouchpoint` — `accountId` is `.guid()` + re-validated in `calendarSync`; `eventId` is an in-memory `.find()` only.
- **No XSS** — no `dangerouslySetInnerHTML`/`innerHTML`; tags/metadata/notes render through escaped JSX.
- **`ensureSchema` is memoized** (in-flight dedup) and **`createTouchpoint` is atomic** (single-client `withTransaction` covering insert + person update).
- **Enum value sets** (ring, energy) and **response-shape unwrapping** (client ↔ routes) are consistent field-by-field.

_Generated by /do:review_

client/src/pages/Tribe.jsx
```diff
@@ -0,0 +1,922 @@
...
+      }
+    }
+
+    setContacts(people);
+    setLoading(false);
+  };
+
+  useEffect(() => {
+    loadContacts();
+  }, []);
```
atomantic, line 691: **[IMPROVEMENT]** Two lifecycle gaps on this effect:

1. **Dead `tribe:changed` producer.** Every mutation route emits `req.app.get('io')?.emit('tribe:changed', ...)` (`server/routes/tribe.js:64,78,85,98,105,118,124`), but this page never subscribes — there's no `socket` import anywhere in the file. The broadcasts are pure overhead today and do nothing for the second-tab / federated-machine case the project supports. Either wire a listener here (`socket.on('tribe:changed', loadContacts)` with `socket.off` cleanup) or drop the 7 route emits.
2. **No mount guard.** `loadContacts` awaits one or more network calls (and a per-legacy-contact create loop) then calls `setContacts`/`setLoading` — if the user navigates away mid-flight it writes to an unmounted component. The child panels (`MemoryLinksPanel`, `TouchpointsPanel`) already use a `cancelled` flag; mirror it here for consistency.

server/routes/tribe.js
```diff
@@ -0,0 +1,128 @@
...
+  const person = await tribe.createPerson(data);
+  req.app.get('io')?.emit('tribe:changed', { personId: person.id });
+  res.status(201).json(person);
+}));
+
+router.get('/people/:id', asyncHandler(async (req, res) => {
+  const person = await tribe.getPerson(req.params.id);
+  if (!person) throw new ServerError('Person not found', { status: 404 });
+  res.json(person);
+}));
```
atomantic, line 90: **[IMPROVEMENT]** `req.params.id` (and `req.params.memoryId` at line 122-123) are never UUID-validated, but they flow straight into a `UUID`-typed `WHERE p.id = $1`. A non-UUID path segment (e.g. `GET /api/tribe/people/not-a-uuid`) raises `invalid input syntax for type uuid` at the DB layer, which the central handler surfaces as a 500 with the raw Postgres message instead of a clean 404/400. The body schemas already `.guid()` their ids — do the same for the path params (a shared `const idParam = z.string().guid()` validated at the top of each `/:id` handler, or a `router.param('id', …)` guard). Same applies to the GET `/people` `ring` query param (line 56), which is passed through unvalidated. Not an injection risk (everything is parameterized) — just wrong status + internal-detail leakage.

server/services/tribe.js
```diff
@@ -0,0 +1,324 @@
...
+import { ensureSchema, query, withTransaction } from '../lib/db.js';
+import { ServerError } from '../lib/errorHandler.js';
+import * as calendarSync from './calendarSync.js';
+
+const DEFAULT_RING_CADENCE = {
+  support: 7,
+  core: 21,
+  tribe: 45,
+  village: 90,
+};
```
atomantic: **[IMPROVEMENT]** The ring→cadence default `{support:7, core:21, tribe:45, village:90}` is duplicated in three places with no shared source: here, the client `RINGS` array (`client/src/pages/Tribe.jsx:28-31`), and (the flat `45`) the SQL column default. Changing one ring's cadence silently drifts the others. They agree today — at minimum add a cross-reference comment; ideally expose this map from one shared module the service and client both read.

server/services/tribe.js
```diff
@@ -0,0 +1,324 @@
...
+  return String(tags).split(',').map((tag) => tag.trim()).filter(Boolean);
+}
+
+function rowToPerson(row) {
+  return {
+    id: row.id,
+    name: row.name,
+    relationship: row.relationship || '',
+    ring: row.ring || 'tribe',
+    cadenceDays: row.cadence_days ?? DEFAULT_RING_CADENCE[row.ring] ?? 45,
```
atomantic: **[NIT]** `cadence_days` is `NOT NULL DEFAULT 45`, so `row.cadence_days` is never null/undefined for a persisted row — the `?? DEFAULT_RING_CADENCE[row.ring] ?? 45` branch is dead and implies the column is nullable (it isn't). Simplify to `cadenceDays: row.cadence_days,` or keep as deliberate defense, but it currently misleads readers about the schema.

server/routes/tribe.test.js
```diff
@@ -0,0 +1,160 @@
...
+    validationApp.use(express.json());
+    validationApp.use('/api/tribe', tribeRoutes);
+
+    const response = await request(validationApp)
+      .post('/api/tribe/people')
+      .send({ name: '', ring: 'outer-space' });
+
+    expect(response.status).toBe(400);
+    expect(tribe.createPerson).not.toHaveBeenCalled();
+  });
```
atomantic, line 179: **[IMPROVEMENT]** Two test-contract gaps:

1. This case sends `{ name: '', ring: 'outer-space' }` — *both* fields are invalid, so a 400 proves only that *something* was rejected; it would still pass if either rule were removed. Split into two cases, each with all other fields valid (one bad `name`, one bad `ring`), to actually pin each rule.
2. The suite has no coverage of the 404 paths the routes establish (`getPerson`/`updatePerson`/`deletePerson` returning null/false → `ServerError('Person not found', { status: 404 })`). Add a test that mocks the service to return null and asserts `status === 404`.

Separately, there is no service-level test (`server/services/tribe.test.js`) for the most contract-dense code — the `listPeople` dynamic WHERE/param-index builder, `normalizeTags`, `rowToPerson` round-trip, and the `23503`→404 mapping. That's the layer a column-rename drift would actually break; the route tests mock the service away.

atomantic:
Addressed the `/do:review` findings in f58d0df2f:

**Critical**
- **Edit-save 400** — `saveDraft` now sends an explicit editable payload (drops body `id` rejected by `z.never()`, and `lastContact: '' → null` which the date regex also rejected — a second latent create-path 400). Pinned by new route tests.
- **Layout clip** — added `/tribe` to `Layout.jsx` `isFullWidth`.

**Improvements**
- Legacy localStorage import no longer fires on a *failed* fetch (null sentinel vs empty list).
- Wired the previously-dead `tribe:changed` broadcast to a listener that reconciles against server truth; added a per-effect cancel guard so `loadContacts` can't `setState` after unmount.
- Validate UUID path params (`:id`/`:memoryId`) and the GET `ring`/`search` query → clean 400 instead of raw-Postgres 500.
- Exported + unit-tested the pure helpers and the `listPeople` query builder (`server/services/tribe.test.js`); added 404-path route tests; split the dual-invalid validation test; `isoDateTime` returns null instead of throwing on a bad string; simplified the dead `cadenceDays` fallback; documented the triplicated ring→cadence default.

**Deliberately not changed**
- `nav.tribe` id (cosmetic; matches existing Brain siblings).
- Federation: the `tribe_*` tables stay machine-local for now. Cross-machine sync is a separate feature (needs sync-schema versioning + peer-receive handlers) — worth a `future` issue if we want relationship data to follow a user across peers.

Issue #1502, opened 2026-06-20
Federate music Artists (and future Albums/Tracks) across peers


atomantic:
The Music studio's **Artists** entity (introduced in the Music page PR) is `db-primary` (PostgreSQL `artists` table) and its storage layer is **federation-ready** — `mergeArtistsFromSync` / `pruneTombstonedArtists` mirror the authors backend exactly. But the artist record kind is **not yet registered** in the cross-peer sync graph, so artists are currently **local-only** (like the Rounds workbench). This issue tracks wiring the registration so artists (and the upcoming Albums/Tracks) federate like Authors do.

### Why deferred
The peer-sync registration is a single, intricate, cross-cutting concern (~8 shared files, positional-arg threading in `tombstoneGc`, per-category schema-version gating, conflict-journal field registration) that is identical in shape for artists/albums/tracks. It's better done **once, together**, with the DB-sync test suite (`*.db.test.js` via `npm run test:db`), than bundled piecemeal into each feature PR.

### Touch-points to mirror from `author` (the template)
- `server/lib/schemaVersions.js` — add `artists: 1` to `PORTOS_SCHEMA_VERSIONS` (its own per-category gate) and `artist: ['artists']` to `RECORD_KIND_SCHEMA_CATEGORIES`.
- `server/services/sharing/peerSync.js` — import `mergeArtistsFromSync`/`getArtist`/`listArtists`; add `'artist'` to `PEER_SUBSCRIBABLE_KINDS`; `artist: 'artists'` in `KIND_TO_CATEGORY`; per-record push case (`sub.recordKind === 'artist'`); `buildArtistAssetManifest` (portrait image, via `portraitImageFilename`); merge handler (`kind === 'artist'` → `mergeArtistsFromSync`); syncability check; universal pull/push guard.
- `server/lib/syncWire.js` — `case 'artist'` in `sanitizeRecordForWire` (strip/tail-re-add the soft-delete pair, like `author`).
- `server/lib/conflictJournal.js` — add `artist: [...PATCHABLE fields]` to `RESTORABLE_FIELDS` so journaled conflicts render per-field diffs (currently falls back to whole-record).
- `server/services/sharing/tombstoneGc.js` — thread an `artistCutoff` through `refusedFromCutoffs` + the sweep, calling `pruneTombstonedArtists`.
[...]

PR #1506, opened 2026-06-21
Add ACE-Step on-device music generation + HuggingFace audio-model installer

atomantic:
## Summary

Phase 4 (final) of the music effort: real on-device generation wired into the Track editor, plus the ability to install additional audio models from HuggingFace. Builds on the merged Rounds rename (#1501), Music studio + Artists (#1503), and Albums + Tracks (#1505).

**ACE-Step engine** (full songs *with vocals*, lyric-conditioned):
- `scripts/generate_acestep.py` — Python sidecar mirroring the existing musicgen/audioldm2 `STAGE:`/`RESULT:` contract, plus a `--lyrics` flag. Renders into a temp dir and moves the WAV to the deterministic output basename.
- `server/lib/pythonSetup.js` — `resolveAcestepPython` + `ACESTEP_VENV_DEFAULT`/`ACESTEP_RUNTIME_DIR`.
- `scripts/setup-image-video.sh` — `INSTALL_ACESTEP=1` provisions `~/.portos/venv-acestep` (the `acestep` pip package + torch); checkpoints auto-download to `~/.cache/ace-step` on first run.
- `server/services/pipeline/musicGen.js` — additive `ENGINES.acestep` entry (`lyrics: true`); `buildSidecarArgs`/`generateMusic` thread `lyrics` **only** to lyric-aware engines (MusicGen/AudioLDM2 never receive the flag).

**Generation route + UI:**
- `server/routes/music.js` — `GET /api/music/engines` (engines + readiness + merged shipped/user models), `POST /api/music/generate` (runs `generateMusic`, lands the WAV in the shared music library, creates or updates a Track with the audio + engine/model/duration/prompt/lyrics metadata).
- `MusicGenPanel` in the Track editor — engine picker (readiness-gated with the install hint, same posture as the FLUX.2 image venv), model picker, duration; Generate uses the track's current prompt (+ lyrics for lyric-aware engines).

**Install audio models from HuggingFace:**
- `server/services/audioModels.js` — user-editable registry (`data/audio-models.json`) that augments the shipped per-engine model lists; merged into `/engines` and a focused `/models/:engine`.
- `music.js` — `POST /api/music/models` (register + stream the HF download via the shared `startHfDownloadStream`, same path as the image/video model installer), `DELETE /api/music/models/:engine/*id` to de-register.
- `MusicGenPanel` — inline "Install from HuggingFace" with streamed progress; installed models are selectable immediately.

Cross-peer federation for Artists/Albums/Tracks remains deferred (#1502).

## Test plan

- `cd server && npm test` — full suite green (13,162 passed, 130 skipped), including new `services/audioModels.test.js`, `routes/music.test.js` (engines/generate/model-install/remove + 503 venv-missing), and extended `services/pipeline/musicGen.test.js` (acestep engine entry + `--lyrics` threading for lyric-aware engines only).
- `cd client && npm run lint && npm test && npm run build` — all green.
- Manual (real model, on the 128GB box): `INSTALL_ACESTEP=1 bash scripts/setup-image-video.sh`, then in a Track: set a prompt + lyrics, pick ACE-Step, Generate → WAV lands in `data/music/`, plays inline, Track persists with metadata. Confirm a not-installed engine shows the install hint and gates Generate; confirm installing a HuggingFace model makes it selectable.


atomantic:
## Review convergence note

This PR went through 8 rounds of codex review; every actionable finding was fixed (including a P1 lyric-erasure data-loss bug, the user-installed-model-not-rendering feature break, fast-fail validation on unknown engine/model/trackId, album-membership reconcile, and the streaming-fetch auth-redirect contract).

The final review surfaced one remaining item I'm **deliberately not fixing**, as it's out of scope per the project's trust model:

> [P2] Avoid rolling back another in-flight model install — when two installs for the *same engine/repo* overlap, the second hits the downloader's `already_running` path and its rollback could de-register the first's entry.

This is a **concurrent-request race between two simultaneous installs of the identical repo**. Per `CLAUDE.md` → Security Model: *"PortOS is single-user… concurrent request races… are unnecessary as defenses against competing actors and should not be added or flagged as concerns."* A single user can't realistically trigger two overlapping installs of the same repo (the install button is disabled while one is in flight), so adding ownership-tracking/locking here would be defending against a non-existent multi-actor scenario the project explicitly excludes.

Flagging it here for the record rather than silently ignoring.

Issue #1583, opened 2026-06-23
Editorial check: climax / resolution power (passive protagonist at the climax)


atomantic:
**Gap.** Nothing checks that the climax is the protagonist's *hardest choice / most active moment*. A climax where the protagonist is rescued by an ally, or where the plot climax doesn't resolve the emotional/thematic core, ships unflagged.

**Kind/scope.** LLM, scope `series`. Sources: `manuscript`, `reverseOutline`, `series.arc.readerMap`, `series.arc.themes`.

**Why.** The climax is the payoff for the reader's time; a passive climax is story-breaking. Complements `plot.structure-momentum`.

**Build.** New `arc.climax-agency` registry entry + stage prompt that identifies the climax scene and judges protagonist agency + whether it resolves the core problem/theme.

---
Part of #1569

Issue #1582, opened 2026-06-23
Editorial check: character consistency / unearned personality shift


atomantic:
**Gap.** `arc.transitions` detects *change moments* but nothing flags *unearned* shifts: a reserved character suddenly cracking jokes with no arc beat, or POV-character knowledge changing mid-scene without on-page learning, or an established trait (allergy, fear) silently contradicted.

**Kind/scope.** LLM, scope `series`. Sources: `manuscript`, `canon` (character bios), `reverseOutline`, `series.characterArcs` (to suppress *earned* transitions).

**Why.** Erodes characterization; LLM can reason "earned by the transition at issue 4" vs false-flag.

**Build.** New `character.consistency` registry entry + stage prompt; must reconcile against authored arcs so intentional growth isn't flagged.

---
Part of #1569

PR #1666, opened 2026-06-24
Editorial check: emotional beat / reaction proportionality (#1584)

### LINKED ISSUE #1584: Editorial check: emotional beat / reaction proportionality
**Gap.** No check that emotional beats match event magnitude: a traumatic event in issue 2 with no processing shown, then untouched in issue 4 — or a minor setback triggering disproportionate grief.

**Kind/scope.** LLM, scope `issue` (with series context). Sources: `manuscript`, `reverseOutline` (scene events), `interiority`.

**Why.** Emotional authenticity collapses when reactions don't track stakes.

**Build.** New `emotion.reaction-proportionality` registry entry + stage prompt mapping flagged events to subsequent reaction beats.

---
Part of #1569

atomantic:
## Summary

Adds a new default-on editorial check, **`emotion.reaction-proportionality`**, that flags emotional beats whose weight doesn't match what happens — the gap described in #1584:

- **Under-reaction** — a high-magnitude event (trauma, a death, a betrayal, a major loss or win) draws no on-page reaction and is never processed in later issues. The canonical failure: a trauma in issue 2 that issue 4 never touches.
- **Over-reaction** — a minor setback triggers grief/rage/despair out of all proportion to what the story has established.

The check is a series-scoped LLM check that reads the stitched manuscript plus the reverse-outline scene map (to weigh each event's magnitude and attribute findings to the right issue). Because an unprocessed event in an early issue can stay unaddressed many issues later, it carries every high-magnitude event still awaiting a proportionate reaction forward across manuscript chunks (`crossChunkSetup`), so the final part can flag the missing payoff. It degrades to a whole-manuscript scan when no outline exists and respects established characterization (a deliberately stoic/numb character isn't flagged for measured restraint). A major trauma left wholly unprocessed is marked high-severity.

**Part gating (final-part for under-reactions only).** "Never processed afterward" is a whole-story claim, and `runChunkedManuscriptCheck` merges findings first-wins and never retracts — so an under-reaction reported from a non-final chunk would persist even after a later chunk pays the event off. The under-reaction verdict therefore rides the `finalPart` gate (only the final/single chunk reports it, judging the carried-forward unprocessed events from the setup digest). **Over-reactions stay local** and report in any chunk, since a disproportionate reaction is fully visible in the text that contains it.

It surfaces in the existing editorial-review UI and Series Autopilot like every other check (issue-anchored findings, `maxFindings` cap).

**Scope note:** the issue lists scope `issue` (with series context); this ships as `scope: 'series'` because the core gap — an event in one issue left unprocessed in a *later* issue — is only visible with the whole arc in view, exactly as the two sibling checks (#1582 character-consistency, #1583 climax) do it. Findings remain issue-anchored via `issueNumber`. The issue's `interiority` source isn't a registered `EDITORIAL_SOURCES` key; the interiority/reaction beats live in the manuscript prose the check already reads, so sources are `['manuscript', 'reverseOutline']`.

Follows the exact 7-file pattern of #1582 / #1583:
- `server/lib/editorial/checkRegistry.js` — `REACTION_PROPORTIONALITY_STAGE` const + `EDITORIAL_CHECKS` entry
- `data.reference/prompts/stages/pipeline-editorial-reaction-proportionality.md` — stage prompt
- `data.reference/prompts/stage-config.json` — stage-config entry (fresh installs)
- `scripts/migrations/132-editorial-reaction-proportionality-stage.js` (+ test) — seeds existing installs on boot (boot runs migrations, not setup-data)
- `server/lib/editorial/checkRegistry.test.js` — registration + behavior tests
- `.changelog/NEXT.md` — entry

## Review

Reviewed locally with **claude** (self-review) and **codex**. Codex raised one [P2]: a non-final chunk could falsely emit an "under-reaction" for an event a later chunk processes (first-wins merge never retracts). Fixed by gating the under-reaction verdict to the final part (second commit); over-reactions remain local.

## Test plan

- `npx vitest run lib/editorial/checkRegistry.test.js` → passes (new `emotion.reaction-proportionality` block: registration, prose gate, manuscript+scene-map vars + forced `emotion` category, empty scene map with no outline, single-chunk marked final, only-last-part-final across a chunked manuscript).
- `npx vitest run ../scripts/migrations/132-editorial-reaction-proportionality-stage.test.js` → 4 passed (fresh-install seed, idempotent no-clobber, missing-sample tolerance, live drift-catch).
- `npx vitest run services/pipeline/editorial/ lib/editorial/` → 596 passed (runner unaffected by the new default-on check).
- CI: lint + test (20.x) green on the first commit; re-running on the fix.

Closes #1584


atomantic:
## Review convergence (claude + codex)

Ran codex over three rounds; addressed two findings in-PR and deferred the third as cross-cutting infra:

1. **[P2 — fixed]** Non-final chunks could emit a persistent false-positive `under-reaction` (first-wins merge never retracts). → Gated the under-reaction verdict to the final part (`finalPart` var, mirroring `arc.climax-agency`); over-reactions stay local.
2. **[P2 — fixed]** The final-part under-reaction needs an `anchorQuote`, but the event text is out of view by then. → The setup digest now carries a short verbatim event snippet (≤200 chars) and the final-part prompt uses it as the anchor (mirrors the climax candidate snippet).
3. **[P2 — deferred → #1667]** The cross-chunk setup digest is best-effort (`runChunkedManuscriptCheck` yields it to manuscript coverage by design), so a packed final chunk can omit the carried event. This is **shared infrastructure** that the just-shipped `arc.climax-agency` (#1583) depends on identically — reserving digest budget would change behavior for every manuscript LLM check, which is too large for a single new-check PR. The check degrades gracefully (a missed finding in an edge case, never a false positive or crash). Filed as #1667 (`plan`).

CI: lint + test (20.x) green on each commit.

Issue #1618, opened 2026-06-23
Editorial check: series-wide pacing / intensity escalation curve


atomantic:
**Gap.** `plot.structure-momentum` detects a sagging middle, but nothing measures the *escalation curve* across the whole series: flat intensity (issue 1 == issue 6), front-loaded climax (issue 2 has the biggest reveal), or stakes/conflict that plateau or de-escalate.

**Kind/scope.** LLM (intensity judgment) + deterministic (per-scene conflict-marker tally), scope `series`. Sources: `manuscript`, `reverseOutline`.

**Why.** Reader exhaustion / anticlimax in long series; a common structural flaw. Subsumes "conflict escalation" as one signal of the same curve.

**Build.** New `pacing.escalation-curve` entry that scores per-issue intensity and flags non-monotonic or plateaued arcs.

---
_Spun out as standalone future work — parent epic #1569 was closed v1-complete (every committed child shipped). Tracked here as an independent `future` item._

Issue #1667, opened 2026-06-24
Editorial infra: guarantee the cross-chunk setup digest reaches final-part-gated checks


atomantic:
## Problem

`runChunkedManuscriptCheck` prepends the cross-chunk **setup digest** (the "setup so far" summary that carries forward whole-story context) only when it fits in a chunk's spare room — it *yields* to manuscript coverage by design (`server/lib/editorial/checkRegistry.js` ~L1045-1083: a chunk packed to the budget runs without a digest rather than dropping its manuscript tail). Chunks are planned at the full usable budget with no space reserved for the digest.

That yielding is correct for checks whose setup digest is a *nice-to-have*. But it is best-effort for checks that **gate a whole-story verdict to the final part** and rely on the carried-forward digest to make (and anchor) that verdict:

- `arc.climax-agency` (#1583) — the final-part climax verdict needs the carried-forward CLIMAX CANDIDATE snippet.
- `emotion.reaction-proportionality` (#1584) — the final-part under-reaction verdict needs the carried-forward unprocessed-event snippet.

If the final chunk is packed to within ~a few hundred chars of the full window, the digest is silently omitted and the final-only finding is missed (graceful — a missed finding, never a false positive or crash — but the advertised cross-chunk path doesn't fire).

## Proposal

Give a check a way to **reserve** a small, bounded budget for the setup digest so a final-part-gated check is guaranteed its carried context, without regressing manuscript coverage for checks that don't need the guarantee. Options:

1. A per-check `reserveSetupDigest: true` (or a token reserve) that `runManuscriptLlmCheck` passes into `planManuscriptChunks` so the chunker carves out `EDITORIAL_SETUP_DIGEST_CHARS` worth of room on the final chunk only.
2. Trim the *manuscript* tail of the final chunk (not the digest) by the digest size when a reservation is requested — the inverse of today's yield, scoped to opt-in checks.
3. Fall back to a second final-chunk LLM call carrying only the digest when it didn't fit, merging its findings.

Whichever is chosen must stay a no-op for the common fits-in-one-call provider and for checks that don't opt in.

Related to #1584 (and the same limitation affects #1583). Surfaced by a codex review of the #1584 PR.

PR #1733, opened 2026-06-27
feat: add arc.regression editorial check for regression / circular / premature-closure arcs (#1619)

### LINKED ISSUE #1619: Editorial check: character arc regression / premature closure
**Gap.** `arc.transitions` detects change moments but not regression (character grows then reverts with no purpose), circular arcs (ends where it started), or premature closure (arc resolves in issue 3, character flat for the next 7 issues).

**Kind/scope.** LLM, scope `series`. Sources: `manuscript`, `series.characterArcs`, `reverseOutline`.

**Why.** Premature arc closure deflates the back half of a series; unmotivated regression confuses readers.

**Build.** New `arc.regression` entry tracking per-character progress across issues and flagging regressions/plateaus/premature closure.

---
_Spun out as standalone future work — parent epic #1569 was closed v1-complete (every committed child shipped). Tracked here as an independent `future` item._

atomantic:
## Summary

Adds a new series-scope LLM editorial check, **`arc.regression`**, that judges the *shape* of each character's arc across the whole series — complementing `arc.transitions`, which detects the change *moments*. It tracks each character's progress across the issues and flags:

- **Regression** — a character grows then reverts to their old self with no purpose or earned reason (distinct from a deliberate, dramatized relapse).
- **Circular arc** — the character ends in essentially the same state they began; the growth is cancelled out.
- **Premature closure** — the arc fully resolves early (e.g. issue 3 of 10) and the character stays flat for the rest of the series, deflating the back half.

The check reads the stitched manuscript plus the reverse-outline scene map and the authored per-character arcs (`series.characterArcs`) to reconcile the planned end-state against the prose. Whole-arc verdicts are gated to the final manuscript part (so a mid-arc character whose later growth is still ahead is not false-flagged), and it carries a per-character progress digest forward across chunks with `reserveSetupDigest` so the final chunk can still see each character's start → peak → latest state. It degrades to a prose-only scan when no outline or authored arcs exist.

Follows the `pacing.escalation-curve` (#1618) recipe end to end: ships the prompt + stage-config for fresh installs and a **boot migration (141)** so existing installs seed the stage (boot runs migrations, not `setup-data.js`).

## Test plan

- `server/lib/editorial/checkRegistry.test.js` — new `arc.regression` describe block: registration shape, gate, arc-category findings, authored-arc rendering, graceful degrade, single-chunk-is-final, and last-chunk-only-final across a chunked manuscript. **451 passed.**
- `scripts/migrations/141-editorial-arc-regression-stage.test.js` — seed on fresh install, idempotent no-clobber of customized prompt/config, no-throw on missing samples, live drift catch. **passed.**
- `scripts/setup-data-drift.test.js` — unchanged baseline still green (seed migration carries no MD5, so it is not swept). **passed.**
- `server/services/pipeline/editorial/checkRunner.test.js` — **80 passed.**

Closes #1619


atomantic:
## Review resolution

**Claude review:** clean — no correctness/behavior bugs; wiring verified end-to-end against the `arc.transitions` / `arc.climax-agency` / `pacing.escalation-curve` siblings.

**Codex review:** one [P2] — `reserveSetupDigest` can trim the final chunk's manuscript tail (where an arc's end-state lives) under tight context budgets.

Addressed in 26663abe1 (comment-only). The trim only bites when the manuscript spans many chunks **and** the final chunk is packed to within the digest's size of the window — a single-chunk run (the common provider-fits-the-book case) has no digest and keeps the whole ending. Reserve is still the correct choice here because the START state and any early resolution — the anchors a circular-arc / premature-closure verdict *cannot* be made without — live **only** in the carried digest; dropping it would keep the last few hundred chars of ending but lose the beginning, the worse failure for an arc-shape judgment. This matches the accepted #1667 tradeoff the three sibling final-part checks already make. The code now documents the tradeoff explicitly.

PR #1879, opened 2026-06-30
fix: honor intentional empty-string clears in memory/message/task merges (#1826)

### LINKED ISSUE #1826: [bugs][MEDIUM] Absent-vs-cleared merge bugs (|| / truthiness collapses intentional empty)
**Category:** Bugs/Perf/Error Handling · **Severity:** MEDIUM · _Filed by /do:better audit_

The absent-vs-cleared footgun called out in CLAUDE.md, in three places:

- `server/services/memory.js:194` — `updateMemory` updates `content` to `""` correctly, but the BM25 reindex gate is `if (updates.content || updates.tags)`. Clearing content to `""` is falsy → reindex skipped → old text stays searchable. Fix: `if ('content' in updates || 'tags' in updates)`.
- `server/services/messageSync.js:245` — `existing.bodyText = threadMsg.body || existing.bodyText`. Empty-body emails collapse to the old body. Fix: `?? ` or `'body' in threadMsg` guard.
- `server/services/cosTaskStore.js:329` — `updatedMetadata[f] = updates[f] || undefined` deletes a field set to `""`/`0`/`false` for `LEGACY_DIRECT_FIELDS`. Fix (if `""` is a valid value): `updates[f] ?? undefined`. _(Confirm `""` semantics first.)_

Complexity: Simple.

atomantic:
## Summary

Fixes three absent-vs-cleared footguns called out in the CLAUDE.md "distinguish absent vs intentionally empty" convention, where a falsy gate (`||` / truthiness) conflated "field absent" with "field cleared to empty" and silently restored a value the caller just cleared.

- **`server/services/memory.js` (`updateMemory`)** — clearing `content` to `""` was correctly written to the record, but the summary-regen gate (`updates.content && ...`) and the BM25 reindex gate (`updates.content || updates.tags`) were falsy, so the stale summary stuck and the old text stayed searchable. Both now gate on key presence (`'content' in updates`), so a clear refreshes the summary and reindexes the now-empty document.
- **`server/services/messageSync.js` (`refreshMessage`)** — an empty-body message (`body === ''`) is a valid full extraction, but `threadMsg.body || existing.bodyText` collapsed it back to the old text. Switched both the thread-match path and the original-message fallback path to `??`, so only an absent body (null/undefined) keeps the prior text.
- **`server/services/cosTaskStore.js` (`updateTask`)** — `updates[f] || undefined` for `LEGACY_DIRECT_FIELDS` (`context`/`model`/`provider`/`app`) mapped an intentional clear to `""` into `undefined`, which the cleanup pass then deleted. Switched to `??` so `""` persists as `""`; `null` still drops the key.

## Test plan

- Added regression tests to all three suites:
  - `memory.test.js`: clearing content to `""` regenerates the summary and reindexes BM25 with `content: ''`; a non-content/tags update does not reindex.
  - `messageSync.test.js`: new `refreshMessage` coverage — empty extracted body overwrites the cached body (thread-match + fallback paths); absent body keeps the cached text.
  - `cosTaskStore.test.js`: clearing `context` to `""` persists as `""` (survives markdown round-trip); `null` still drops the field.
- `npx vitest run services/memory.test.js services/messageSync.test.js services/cosTaskStore.test.js` → 173 passed.

Closes #1826


atomantic:
Addressed the codex P2 (memory clear didn't reach the production path):

- **Mirrored the gate into the production backend.** `memoryDB.js` (the Postgres backend that normal installs delegate to via `memoryBackend`) kept the old truthy summary gate; it now uses the same shape-based gate (`typeof updates.content === 'string'`) as the file backend, so the two stay consistent and neither can crash `generateSummary` on a null/undefined content. Postgres FTS (tsvector over content+summary) updates automatically when the column changes, so there's no separate reindex gate to fix there.
- **Schema `min(1)` is intentional and kept.** `memoryUpdateSchema` is a partial of `memoryCreateSchema`, whose `content` is `z.string().min(1)`. Empty memory content is deliberately invalid (you can't create an empty memory either), so clearing content to `""` via the API is intentionally rejected — relaxing it would allow meaningless blank memories and diverge from create. The genuinely reachable absent-vs-cleared clears are `tags: []` (file-backend BM25) and the messageSync body / cosTaskStore legacy-field sites.

Issue #1858, opened 2026-06-30
Music-video federation: track-linked project ships no master audio to a musicVideoProjects-only subscriber


atomantic:
Surfaced by the v2.24.0 release review (codex pass on PR #1857).

**Where:** `server/services/sharing/peerSync.js` (music-video push manifest, ~L1806).

**Problem:** When a music-video project uses `trackId` (the normal create-UI path), `uploadedAudioFilename` is null, so the push manifest includes no master audio. A peer **selectively subscribed to `musicVideoProjects` only** (not the music/tracks category) receives the project and scene clips but the linked track never arrives, so `resolveMasterAudioPath()` later fails with `Linked track not found` / `AUDIO_MISSING` and the peer can't render.

**Context / why deferred:** This is the documented scope boundary from #1772 ("a track-linked project ships no audio here — the linked track federates its own"). It only bites a partial-subscription configuration, is not a regression, and is not a crash/corruption — so it was deferred from the release rather than risk a last-minute federation change.

**Fix options:** When a project has a `trackId` and no `uploadedAudioFilename`, either (a) resolve the track's audio file and bundle it in the music-video manifest, or (b) auto-subscribe/ship the linked track record alongside the project so a `musicVideoProjects`-only subscriber still gets playable audio. Mirror the existing reference-frame/clip bundling path.

PR #1926, opened 2026-06-30
Ack-track bundled music-video track tombstones to close a track-GC strand window

### LINKED ISSUE #1922: Music-video federation: bundled linked-track tombstone isn't ack-tracked, so track GC can strand a delete on a Music-Videos-only peer
Follow-up from #1858 (PR with the linked-track bundling fix).

**Context.** #1858 made a music-video push bundle the linked track *record* (`linkedTrack`) so a peer subscribed to `musicVideoProjects` only (no Tracks category) can resolve `project.trackId` and render. A linked-track *delete* now also fans out to its linked projects, and the project push bundles the track *tombstone* so the peer converges.

**Remaining gap (codex review, deferred from #1858 as out-of-scope depth).** The bundled `linkedTrack` tombstone is merged on the receiver, but `ackedDeletesUpTo` / `computeAckedDeletesFromPayload(record, issues)` in `server/services/sharing/peerSyncReceive.js` only considers `record` and `issues`, not `linkedTrack`. A `musicVideoProjects`-only peer therefore never acks the track delete through any per-peer track tombstone cursor (it has no `track` subscription). If the bundled delivery transiently fails AND the sender's track-tombstone GC prunes the tombstone before a retry, future project retries omit it — leaving the peer's stale linked track/audio renderable.

**Why deferred.** Narrow double-fault window (delivery fails *and* GC prunes before retry); the common path converges correctly. A proper fix needs dedicated tombstone-cursor work — either (a) advance a track tombstone ack for the peer from the bundled delivery, or (b) auto-create a (hidden) track subscription for the peer when bundling, or (c) a separate out-of-band ack channel for bundled secondary records. That's a deeper federation-GC change deserving its own scoped PR + review, not a late add-on.

**Where:**
- `server/services/sharing/peerSyncReceive.js` (`computeAckedDeletesFromPayload`, `ackedDeletesUpTo` return, ~L414/L462+)
- `server/services/sharing/peerSyncPush.js` (`ackDeletesUpTo` handling, music-video push branch)
- `server/services/sharing/tombstoneGc.js` (track tombstone prune horizon)

atomantic:
## Summary

Follow-up from #1858 (deferred there as out-of-scope depth). #1858 made a music-video project push bundle its linked track record (`linkedTrack`) so a peer subscribed to `musicVideoProjects` only (no `track` subscription) can resolve `project.trackId` and render, and made a track delete fan out to its linked projects so the bundle can also carry a tombstone.

The remaining gap: `computeAckedDeletesFromPayload` in `server/services/sharing/peerSyncReceive.js` only considered `record`/`issues`, never the bundled `linkedTrack` — so a musicVideoProjects-only peer could never ack a bundled track delete through any cursor. Separately (and more fundamentally), `tombstoneGc.js`'s `track` cutoff cohort (`peerIdsSubscribedToKind`) only ever included peers with a *direct* `track` subscription — a musicVideoProjects-only peer was entirely invisible to it, so fixing the ack alone would not have closed the gap. If a bundled delivery transiently failed and the sender's track-tombstone GC pruned the tombstone before a retry, the peer's stale linked track/audio would stay renderable.

### Which option, and why

The issue offered three options: (a) advance a track tombstone ack for the peer from the bundled delivery, (b) auto-create a hidden track subscription when bundling, (c) a separate out-of-band ack channel.

I implemented **(a)**, but extended beyond "fix the ack value" once it became clear that alone wasn't sufficient (see above) — the surgical completion of (a) is:

1. **`peerSyncReceive.js`** — `computeAckedDeletesFromPayload` now also folds in a bundled `linkedTrack` tombstone's `deletedAt` into the receiver's returned `ackedDeletesUpTo` water-mark.
2. **`tombstoneGc.js`** — the `track` GC cohort (`peerIdsSubscribedToKind`/`minConfirmedPushedAtForKind`) now also folds in peers with a `musicVideoProject` subscription, since that subscription is a secondary delivery vehicle for the track's tombstone.
3. **`peerSyncPush.js`** — added a bundle-specific confirmed-delivery floor, `lastConfirmedTrackBundleAtMs`, stamped on a `musicVideoProject` subscription row only when the push landed (2xx) *and* the bundled track merge actually succeeded (`!trackSyncPending`). `tombstoneGc.js` reads this field instead of the generic `lastConfirmedPushedAt` for `musicVideoProject` rows protecting the `track` cutoff — `lastConfirmedPushedAt` advances on every successful project push regardless of whether the bundled track merge failed, so reusing it would let GC treat an unconfirmed bundled tombstone as delivered. The row's protection of its *own* `musicVideoProject` tombstone is unaffected — it still reads `lastConfirmedPushedAt` as before.

I considered (b) (auto-create a hidden `track` subscription on bundling) — it would reuse 100% of the existing machinery with no new fields, but it also means a `musicVideoProjects`-only peer silently starts receiving direct `track`-kind pushes outside the bundle, which is a bigger behavioral/UI-surfacing change (an extra, invisible subscription row) for a narrow double-fault window. (a) stays scoped to bookkeeping only the GC sweep reads.

No wire-payload shape changed (the new `lastConfirmedTrackBundleAtMs` field is local subscription state, never sent over the network), so no `schemaVersions.js` bump is needed — same reasoning that exempted the original `linkedTrack` bundling.

## Test plan

- New tests in `server/services/sharing/peerSync.test.js`:
  - `computeAckedDeletesFromPayload` folds in a tombstoned bundled `linkedTrack` (and does NOT fold in a live one).
  - `pushRecordToPeer` stamps `lastConfirmedTrackBundleAtMs` only when the bundled track merge succeeded; does not stamp it when `trackSyncPending` is true.
- New tests in `server/services/sharing/tombstoneGc.test.js` (new describe block): a musicVideoProject-only peer (no `track` sub) is included in the `track` ack cohort; the `track` cutoff is held back until the bundle confirms (and is NOT released by the generic `lastConfirmedPushedAt` alone — the field-confusion regression case); the `musicVideoProject` cutoff's own protection is unaffected (still reads `lastConfirmedPushedAt`); a plain `track`-only subscriber still gates as before.
- `cd server && NODE_ENV=test npm test` — full suite green except the pre-existing, documented order-dependent `getFullSyncCoverageForPeer` flake in `peerSync.test.js` (reproduces identically on `origin/main` with my changes stashed out — confirmed not caused by this change).

Closes #1922


atomantic:
Ran codex review 3 rounds (the cap), fixing a real finding each round: (1) ack the bundled \`linkedTrack\` tombstone's deletedAt into the receiver's water-mark; (2) reverted an over-eager per-peer GC-cohort exclusion that could under-protect a track only reachable via the bundle, replacing it with a safe \`lastPushedHash\`-based backfill for pre-#1922 subscription rows; (3) gated both the sender's \`lastConfirmedTrackBundleAtMs\` stamp and the receiver's ack on the bundle actually being sent/applied, not just the absence of a failure signal. Full server suite is green (only the pre-existing, documented \`getFullSyncCoverageForPeer\` order-dependent flake remains, reproduces identically on main). Stopping here per the 3-round review cap — ready for human/orchestrator merge review.

Issue #1924, opened 2026-06-30
[ux] Theme-token consistency: remaining off-token color remap (digital-twin, cos, meatspace, etc.)


atomantic:
Related to #1909. Tracks the remainder of the broad off-token color remap after the slice shipped in the PR for #1909.

## What that PR shipped
- `--port-accent-2` semantic token (purple/violet) added to `client/src/index.css` (`@theme` + `:root` default) and tuned per-theme in `client/src/themes/portosThemes.js` (all 8 themes).
- `useChartColors()` (from #1851) applied to the last 8 recharts files that weren't using it: `ActivityBloodCorrelation.jsx`, `AlcoholHrvCorrelation.jsx`, `BloodPressureCard.jsx`, `BodyCompChart.jsx`, `NicotineChart.jsx`, `NicotineHealthCorrelation.jsx`, `meatspace/post/PostHistory.jsx`, `SleepCard.jsx`.
- `cos/tabs/AgentCard.jsx:594` + `:602` JIRA badge/link base+hover pair remapped to `port-accent`.
- Full token remap of `meatspace/EpigeneticTracker.jsx` (evidence-strength badges, compliance bar, log buttons → `port-success`/`port-warning`/`port-error`; recommendation icon/text/CTA → `port-accent-2`) and the two `ChiefOfStaff.jsx` "Learning: good" purple sites → `port-accent-2`, as worked examples of the new token in use.

## What's still off-token (counts in `client/src/components/`, post-#1909-PR)
```
blue:    124
green:   310
cyan:    307   (mostly CyberCity — already has a `.cybercity-themed` CSS remap layer in index.css; lower priority)
slate:    20
indigo:   18
purple:  304
violet:   40
```
Plus ~122 more sites across `client/src/pages/` (not counted above).

By feature-area file count (`client/src/components/`):
```
digital-twin: 25   cos: 24   meatspace: 19 (partial)   city: 11   ui: 9
goals: 9   brain: 8   apps: 5   pipeline: 4   messages: 4   agents: 4
writers-room: 2   wiki: 2   settings: 2   gsd: 2   voice: 1   + ~10 misc top-level files
```

## Suggested next slices (grouped by feature area, per the original issue's guidance)
- [x] `digital-twin/` (25 files) — largest remaining single area. Shipped (24 of 25 files remapped; `constants.js` intentionally left as a category-color enum, per the caution below).
[...]

Issue #1909, opened 2026-06-30
[ux] Theme-token consistency: broad off-token color remap + semantic purple/slate/indigo tokens


atomantic:
Follow-up to #1851, which scoped the highest-value/lowest-risk theme-token wins (a `useChartColors()` hook applied to AlcoholChart/MetricCard, the 13 `bg-port-accent hover:bg-blue-600` CTA hovers + `text-port-accent hover:text-blue-300` sites, and the off-token terminal panels). This issue tracks the remaining broad work intentionally split out of #1851 to keep that PR reviewable and low-risk:

**Remaining**
- **~900+ off-token Tailwind color utilities** in `client/src/components/` — the common `blue/green/cyan/slate/indigo-*` utilities should map to `port-*` tokens (or new semantic tokens) so they follow theme changes. This is a large, mechanical-but-judgement-heavy sweep that warrants its own focused PR(s), ideally grouped by feature area.
- **No-token purple palette** — `purple-*` (e.g. `ChiefOfStaff.jsx` slate/indigo/purple) has no `--port-*` equivalent. Decide whether to (a) introduce a semantic `--port-accent-2` / purple token, or (b) collapse purple usage onto existing accent tokens. Define the token(s) first, then remap.
- **Remaining recharts cards** — apply `useChartColors()` (added in #1851) to the other ~39 chart sites: `meatspace/AlcoholHrvCorrelation.jsx`, `SleepCard.jsx`, and the rest enumerated in #1851.
- **Non-token blue hovers** without a token base — e.g. `cos/tabs/AgentCard.jsx:602` `text-blue-400 hover:text-blue-300` (base + hover both need remapping together).

**Why split:** #1851 attempting all ~900 utilities + a new palette token in one PR would be unreviewable and high-regression-risk. The hook + token plumbing landed in #1851; this issue is the bulk remap on top of it.

atomantic:
Add to scope (split out of #1851 during review): `cos/TerminalCoSPanel.jsx` needs a **coordinated foreground+background** theming pass, not just a background-token swap. Its panel background was left as the fixed dark hex `#0d1117` because the inner content hardcodes light text (`text-white`, `text-gray-300/400/500/600`, colored status indicators); swapping the background alone to `--port-terminal-bg` (which is light in the day themes) renders light-on-light. Re-theme the inner text in the same pass — likely `--port-terminal-text` plus port-* status colors — then move the background onto the token.

atomantic:
[...]

PR #1935, opened 2026-06-30
fix(issue-1909): coordinate TerminalCoSPanel foreground+background re-theme

atomantic:
## Summary

Refs #1909 — ships the one well-scoped slice called out in the issue's most recent comment: `client/src/components/cos/TerminalCoSPanel.jsx`'s background was intentionally left pinned to a fixed dark hex (`#0d1117`) because the inner content hardcoded light text (`text-white`, `text-gray-300/400/500/600`) and status colors. Swapping the background alone onto `--port-terminal-bg` (light in the day themes) would have rendered light-on-light.

This PR re-themes foreground and background together in one pass:

- Background → `bg-[var(--port-terminal-bg)]` (the dedicated, per-theme-tuned token already used by `GsdDocumentsPanel.jsx`/`DocumentsTab.jsx`/`Shell.jsx`).
- Primary/secondary text → `text-[var(--port-terminal-text)]` with opacity tiers (`/75`, `/55`, `/40`) for dim/muted/faint variants. Blending the foreground color into the background at a fixed opacity keeps contrast correct regardless of whether the active theme's terminal surface is dark or light, rather than guessing a fixed gray scale.
- Traffic-light dots, log-level colors, status indicators, and the start/stop buttons → the existing `port-success` / `port-warning` / `port-error` / `port-accent` semantic tokens (matching the mapping convention established in #1924/#1925: green→success, yellow→warning, red→error, cyan→accent).
- Borders → `border-port-border` (same pairing already used elsewhere with `--port-terminal-bg`).

The remaining ~900-site bulk remap tracked by #1909/#1924 is out of scope for this slice — issue stays open.

## Test plan

- [x] `cd client && npm run build` — succeeds, no errors.
- [x] `npx eslint src/components/cos/TerminalCoSPanel.jsx` — clean.
- [x] Verified the built CSS (`dist/assets/*.css`) actually contains generated `color-mix()` rules for every `text-[var(--port-terminal-text)]/NN` and `bg-[var(--port-terminal-text)]/NN` opacity class used — confirms Tailwind's source scanner picked up the literal class strings (they're defined as complete literal constants rather than built via runtime concatenation, which would have been invisible to the scanner).
- [x] Manual grep confirms no remaining `text-white`, `text-gray-*`, or `#0d1117` in the file.


atomantic:
## Review round 1

**claude headless review:** CLEAN — no findings (verified no leftover hardcoded grays/white/black/cyan/red/green/yellow, confirmed the literal-string opacity-tier approach generates correct Tailwind CSS via `color-mix()`, confirmed contrast across all 8 themes' `--port-terminal-bg`/`--port-terminal-text` pairs).

**codex review:** one IMPORTANT finding — `TerminalCoSPanel.jsx:134` still colors the ASCII avatar art via `stateConfig.color`, sourced from hardcoded hex values in `cos/constants.js`'s `AGENT_STATES` map, and flagged low contrast for `#f59e0b` (~2.0:1) against the Classic Noon light theme's terminal background.

**Decision: deferring this, not fixing in this PR.** `AGENT_STATES.color` is shared mood-enum infrastructure consumed by 9 files across the whole `cos/` avatar-skin system (`CoSCharacter`, `CyberCoSAvatar`, `EsotericCoSAvatar`, `MiniCharacterCoSAvatar`, `MuseCoSAvatar`, `NexusCoSAvatar`, `SigilCoSAvatar`, `StateLabel`, plus this file) — not something specific to `TerminalCoSPanel.jsx`, and pre-existing (untouched by this diff). It's the same "intentional category-color enum" pattern the digital-twin remap (#1924) explicitly left alone: 7 distinct hues differentiate 7 agent moods, and collapsing them onto the 5 `port-*` semantic tokens would lose that differentiation. Fixing it properly means designing new mood-aware tokens and touching 9 files across the avatar-skin system — out of scope for the single-file slice #1909's most recent comment called out. Filed as a follow-up note on #1909 for whoever picks up the next slice of the sweep.

No code changes this round; PR scope (background + inner text/status colors of `TerminalCoSPanel.jsx`) is otherwise CLEAN per both reviewers.

Issue #2068, opened 2026-07-02
SSRF: pin catalog URL-ingest (CDP/Chrome) fetch against DNS-rebinding TOCTOU


atomantic:
Related to #2046.

The RSS-feed path (`server/services/feeds.js#fetchFeedXml`) was migrated to the pinned `safeUrlFetch` guard in #2046, closing its DNS-rebinding TOCTOU. The catalog URL-ingest path was intentionally split out here because it needs a different approach than the drop-in.

## The gap
`server/services/catalogIngestSources.js#fetchUrlMainText` (~L80) already:
1. Pre-checks the request URL with `assertIngestUrlSafe` (scheme + resolve-once + `isBlockedIngestHost`), and
2. Re-asserts the FINAL landed URL post-navigation (catches a redirect to a blocked host).

But the actual fetch is performed by **Chrome via CDP** (`navigateToUrl`), which re-resolves DNS independently of our `dns.lookup` check. So between `assertIngestUrlSafe` and Chrome's TCP connect there is still a rebinding window: DNS can answer public during our validation and private (loopback / `169.254.169.254` / LAN) at Chrome's connect. Unlike undici, we can't hand Chrome a per-request pinned `connect.lookup`.

## Why it wasn't fixed in #2046
Closing this properly requires pinning Chrome's resolver, which is a substantial, higher-risk change than the feeds drop-in — e.g.:
- launch/relaunch Chrome with `--host-resolver-rules="MAP <host> <vetted-ip>"` per navigation, or
- CDP request interception (`Fetch`/`Network.setRequestInterception`) to enforce the vetted IP, or
- resolve-and-pin at a proxy layer in front of Chrome.

Each needs its own design + testing pass. The existing pre/post re-assert plus the single-user private-network trust model make the residual risk low, so it's tracked separately rather than bloating the feeds PR.

## Acceptance
- Chrome connects only to the IP `assertIngestUrlSafe` vetted (or the navigation is refused), for both the initial URL and any redirect hop.
- Regression coverage in `catalogIngestSources.test.js` proving a rebind-to-private answer after validation cannot reach Chrome.

Issue #1859, opened 2026-06-30
safeUrlFetch SSRF guard has a DNS-rebinding TOCTOU (validation lookup != connect lookup)


atomantic:
Surfaced by the v2.24.0 release review (codex pass on PR #1857).

**Where:** `server/lib/safeUrlFetch.js` (`isPublicHttpUrlSafe` validates a hostname via `dns.lookup`, then `fetchWithTimeout` opens the socket with a **separate** lookup).

**Problem (TOCTOU / DNS rebinding):** The SSRF gate resolves DNS once for validation and approves a public IP, but the subsequent fetch resolves the hostname **again** at connect time. A rebinding host can answer a public address for the check and then loopback/link-local/cloud-metadata for the actual fetch, bypassing the guard. Redirect targets are re-validated (good), but the first hop and every hop still re-resolve independently.

**Context / why deferred:** PortOS is single-user on a private Tailscale network and the only URL sources are user-pasted Pinterest board URLs + stored feed URLs (no untrusted third party supplies hostnames), so the realistic exploitability is low. The proper fix (pin the vetted resolved IP and connect to it, e.g. a custom undici `lookup`/dispatcher while preserving Host/SNI) is non-trivial networking plumbing that shouldn't be rushed into a release merge.

**Fix:** Resolve once and connect to that exact address — supply a custom `lookup` (or a pinned-IP dispatcher) to the fetch so the connection uses the address the guard validated, and re-validate on every hop. Add a regression test with a lookup that returns public-then-private.

PR #2069, opened 2026-07-02
fix: pin RSS feed fetches against DNS-rebinding TOCTOU (#2046)

### LINKED ISSUE #2046: SSRF: migrate feeds.js + catalogIngestSources.js to the pinned safeUrlFetch guard (residual DNS-rebinding TOCTOU)
## Summary

The v2.25.0 SSRF hardening (#1859) introduced `server/lib/safeUrlFetch.js`, which closes the DNS-rebinding TOCTOU by resolving a hostname once and **pinning the connection to the vetted IP** (via a per-request undici `Agent` whose `connect.lookup` returns the validated address), re-validating and re-pinning every redirect hop. The `moodBoard/pinterest.js` callers were migrated to it.

However, the module header notes it was meant to unify the "subtly different" copies of this guard — and **two sibling paths of the same class were not migrated and still have the rebinding TOCTOU**:

### 1. `server/services/feeds.js` — `fetchFeedXml` (~L464)
Calls `isHostSafe(hostname)` (a `dns.resolve4/6` public-IP check) and then does an **unpinned** `fetchWithTimeout(url, …)` — undici re-resolves the hostname at connect time. An attacker controlling DNS for a user-added feed URL can answer a public IP to the check and a loopback / link-local / `169.254.169.254` (cloud-metadata) address at connect. The redirect hop (re-check + unpinned fetch) has the same gap. The inline comment literally claims "(prevents rebinding attacks)" — a false documentation contract.

### 2. `server/services/catalogIngestSources.js` — `assertIngestUrlSafe` / `fetchUrlMainText` (~L80)
Pre-checks DNS, then Chrome (CDP) does its own resolution on `navigateToUrl` — TOCTOU. Harder to pin (it's a browser, not undici), so this one likely needs a different approach than the drop-in.

## Impact (bounded)

Under PortOS's trust model (single-user, private-network/Tailscale, never public) the reachable targets are PortOS's own localhost services and, on a cloud VM, the metadata endpoint. The entry points are user-reachable (add-feed, catalog URL ingest). Pre-existing (not introduced by v2.25.0), which is why it did not block the release.

## Proposed fix

- **feeds.js**: migrate `fetchFeedXml` to `safeUrlFetch.fetchPublicText({ headers, timeoutMs })`. NOTE a **policy nuance**: `safeUrlFetch` intentionally *allows* private/LAN hosts (Tailscale peers, home wikis) while `feeds.js`'s current `isHostSafe` blocks *all* private IPs. Decide whether feed fetches should adopt the looser catalog-ingest posture (allow LAN feeds) or whether `safeUrlFetch` needs a strict-mode variant that blocks all-private while still pinning. Update `feeds.js` tests accordingly. Fix the false "prevents rebinding attacks" comment.
- **catalogIngestSources.js**: the CDP path can't reuse the undici pin; needs its own design (e.g. resolve-then-navigate-by-IP with a Host override, or a pre-navigation block on re-resolution). Track as a separate sub-task.

Surfaced by the v2.25.0 release security review.

atomantic:
Closes #2046. (Catalog CDP ingest rebinding pin split out to #2068 — see below.)

## Problem
`server/services/feeds.js#fetchFeedXml` resolved a feed hostname, verified it wasn't private via `isHostSafe`, then made a **separate, unpinned** `fetchWithTimeout`. undici re-resolves at connect time, so an attacker's DNS could answer public during validation and private (home-network / `169.254.169.254` cloud-metadata / loopback) at the actual TCP connect — a classic DNS-rebinding TOCTOU. The redirect hop had the same gap, and an inline comment falsely claimed it "prevents rebinding attacks."

The pinned guard `server/lib/safeUrlFetch.js` (added in #1859, already used by moodBoard/Pinterest) closes exactly this: it resolves the hostname once and pins the undici connection to that vetted IP via a per-request `Agent` whose `connect.lookup` returns the validated address, re-validating and re-pinning on every redirect hop.

## What shipped
- **`feeds.js#fetchFeedXml`** now routes through `fetchPublicText` from `safeUrlFetch`. Removed the now-dead `isHostSafe` / `isPrivateIP` / dual-`resolve4`+`resolve6` helpers and the `dns` / `net` / `fetchWithTimeout` imports from feeds.js. Fixed the false "prevents rebinding" comment.
- **`safeUrlFetch` gained a strict `blockPrivate` mode** — see policy decision below.
- **`safeUrlFetch` gained a `throwOnUnsafe` option** (default `true`, preserving the route-context 400 contract that `assertPublicHttpUrl` / Pinterest rely on). feeds pass `throwOnUnsafe: false` so an unsafe/private feed URL returns null → the existing friendly "Could not fetch feed — check the URL" message instead of a bubbled 400.

## Policy decision — private/LAN feeds
`safeUrlFetch`'s default posture intentionally **allows** private/LAN hosts (Tailscale peers, home wikis). feeds.js historically **blocked all private IPs**. I preserved feeds' stricter posture by adding an opt-in `blockPrivate` mode (full RFC1918 / CGNAT-adjacent / IPv6 ULA `fc00::/7` rejection, for both host literals and DNS-resolved addresses) rather than loosening feeds. Rationale: a feed URL is arbitrary user-supplied input; silently allowing it to be pointed at the home network would be a behavior regression and a broader SSRF surface than the fix intends. This keeps feeds' observable behavior identical while adding the connect-time pin. The default LAN-allowing posture is unchanged for existing callers (Pinterest).

Note: the pin structurally supersedes feeds' old parallel A/AAAA "happy-eyeballs" defense — since we resolve once and connect to exactly that address, undici can't prefer a private AAAA at connect time. Those dual-lookup tests were replaced with single-resolve pinning tests.

## Catalog CDP ingest — split out (#2068)
`catalogIngestSources.js#fetchUrlMainText` already pre-checks the URL and re-asserts the final landed URL post-navigation, but the fetch itself is done by **Chrome via CDP**, which re-resolves DNS independently — we can't hand Chrome a per-request pinned lookup the way we can undici. Closing that properly needs `--host-resolver-rules` / CDP request interception (substantial, higher-risk). Per the issue's own "track as a separate sub-task" note, filed as **#2068** rather than bloating this PR.

## Test plan
- `server/lib/safeUrlFetch.test.js` — added `blockPrivate` coverage (private literal / resolved / IPv6-ULA rejected; public still allowed) and `throwOnUnsafe` coverage (default throws 400; `false` returns null with no fetch).
- `server/services/feeds.test.js` — reworked the DNS mock from `resolve4`/`resolve6` to `dns.lookup` and updated the SSRF suite to the single-resolve pinning model (literal private rejected without resolving, resolve-to-private rejected, fail-closed on unresolvable, redirect-to-private dropped, connection pinned via dispatcher).
- `server/services/moodBoard/pinterest.test.js` — unchanged, still green (default posture preserved).
- All 4 suites pass locally (`safeUrlFetch`, `feeds`, `pinterest`, `catalogIngestSources`), plus `lib/index.test.js` barrel check.

https://claude.ai/code/session_01TSuyZ5XG9JiijEibeJjL6T

### DIFF
--- a/.changelog/NEXT.md
+++ b/.changelog/NEXT.md
@@ -1,5 +1,9 @@
 # Unreleased Changes
 
+## Security
+
+- **RSS feed fetches now pin the connection to the SSRF-vetted IP, closing a DNS-rebinding TOCTOU.** `feeds.js` previously resolved a feed hostname, verified it wasn't private, then made a *separate, unpinned* fetch — undici re-resolved at connect time, so an attacker's DNS could answer public during validation and private (home-network / cloud-metadata) at the actual connect. Feed fetches now route through the shared `safeUrlFetch` guard (the same one moodBoard/Pinterest uses since #1859): the hostname is resolved once and the connection is pinned to that exact vetted address, re-validated and re-pinned on every redirect hop. `safeUrlFetch` gained a strict `blockPrivate` mode so feeds keep their historical block-all-private posture (an arbitrary user-added feed URL still can't be aimed at the LAN) while gaining the pin, and a `throwOnUnsafe: false` option so the feeds path returns its existing friendly "Could not fetch feed" result instead of a bubbled 400 (#2046). The catalog URL-ingest path (Chrome/CDP) has a residual rebinding gap tracked separately.
+
 ## Workspace Contexts
 
 - **Switch project workspaces from ⌘K and voice, and CoS now auto-snapshots the workspace it leaves.** A new `workspace_switch` action ("switch workspace to BookLoom", "restore my PortOS context") saves a snapshot of the workspace you're leaving, then reconciles the target project's saved context — git branch, in-repo shell sessions, scoped tasks — and reports what's still live to re-attach; it appears in the ⌘K palette and resolves by voice. When CoS dispatches a coding agent against a different app/repo than it was last working in, it now silently snapshots the previous workspace's context first (snapshot-only — no auto-restore, no LLM calls), so switching between repos preserves what each project looked like, visible on the Workspace Contexts page (#2035).
--- a/server/lib/safeUrlFetch.js
+++ b/server/lib/safeUrlFetch.js
@@ -15,6 +15,12 @@
  * private/LAN hosts (a single-user tool legitimately reaches Tailscale peers /
  * home wikis). A redirect to a blocked target fails CLOSED (returns null) — the
  * landed hop is revalidated exactly like the first.
+ *
+ * Strict posture (`blockPrivate: true`): additionally reject ALL private/LAN
+ * addresses (RFC1918, CGNAT-adjacent, IPv6 ULA fc00::/7, …) — both host literals
+ * and DNS-resolved addresses. RSS feeds use this so an arbitrary user-added feed
+ * URL can't be pointed at the home network, preserving feeds.js's historical
+ * block-all-private posture while gaining the connect-time IP pin.
  */
 
 import dns from 'dns/promises';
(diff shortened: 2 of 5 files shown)


atomantic:
### Review gate

- **codex**: 3 rounds. Round 1 found two real gaps in the strict `blockPrivate` classifier — missing CGNAT `100.64.0.0/10` and the deprecated IPv4-compatible IPv6 form (`new URL('http://[::192.168.1.1]')` → `[::c0a8:101]`) slipping through as public. Round 2 found one Low — a malformed redirect `Location` could throw from `new URL(location, url)`, bypassing the null contract feeds rely on. All fixed with regression tests; round 3 = "No issues found."
- **claude** (headless `-p`): produced no usable output in this environment on either round (empty stdout); relied on codex + local self-review.

### Behavior nuance
The strict `blockPrivate` posture now also blocks CGNAT `100.64.0.0/10` (Tailscale's native range). feeds.js already blocked `10/172.16/192.168`, so LAN-hosted feeds were never supported; this closes the one remaining private-ish gap and makes "block all private" consistent. Default posture (Pinterest/catalog LAN access) is unchanged.

All suites green: `safeUrlFetch`, `feeds`, `pinterest`, `catalogIngestSources` (109 tests).

Issue #2097, opened 2026-07-04
Wordplay tab: persist practice results (currently zero tracked progress) and unify with LLM drill surface


atomantic:
# Wordplay tab: persist practice results (currently zero tracked progress) and unify with LLM drill surface

## Motivation

The dedicated Wordplay practice surface is the one place on the platform where practice **doesn't count**:

- `WordplayTrainer.jsx` scores every response via `scorePostLlmDrill` (`:221`) and shows a running average (`:336-341`) but never persists anything — no `submitTrainingEntry`, no session save. Close the tab and the practice never happened.
- The identical four drill types (compound-chain, bridge-word, double-meaning, idiom-twist) **are** saved when run through a POST session — so the same skill has two entry points with contradictory tracking semantics (`WordplayTrainer.jsx:13-50` vs `PostDrillConfig.jsx:96-114`), different scoring persistence, and different UX (consent modal vs session). Morse's standalone surface, by contrast, does log via `logTraining` (`MorseTrainer.jsx:453-457`).

## Design

1. **Persist wordplay practice** to the training log: on each completed round, `submitTrainingEntry` with drill type, correct/total, totalMs (the same contract Morse uses; `trainingEntrySchema` in `postValidation.js` extended to accept the four wordplay types — currently Morse-scoped, `postValidation.js:76`).
2. **Per-question results** worth keeping (the drill cache means prompts are reusable): include the scored evaluation in the entry payload or a compact `questions[]`, matching what POST sessions store, so the progress dashboard can trend wordplay accuracy alongside everything else.
3. **Unify the surfaces**: extract the shared drill-play core (prompt render, answer collection, scoring call — much of it already in `WordplayDrillUI.jsx`) so the standalone tab and the in-session runner are the same component with different hosts. Standalone keeps the consent-gated cache flow (#1954 tracks generalizing that pattern); in-session keeps session semantics. One scoring path, one persistence contract.
4. Wordplay activity counts toward the unified streak and time-in-training (via training log — automatic once #1 lands and the unified-streak issue is done).

## Touchpoints

- `client/src/components/meatspace/post/WordplayTrainer.jsx`, `WordplayDrillUI.jsx`, `PostLlmDrillRunner.jsx` (shared core extraction)
- `server/lib/postValidation.js` (`trainingEntrySchema` type scope)
- `server/services/meatspacePostTraining.js` (accepts wordplay entries — likely no change beyond validation)

## Acceptance criteria

[...]

Issue #1090, opened 2026-06-09
GitTab modal backdrop accessibility (dialog roles + close affordance)


atomantic:
From the 2026-06-09 better audit (Stack-Specific a11y, LOW).

`client/src/components/apps/tabs/GitTab.jsx` (~lines 805, 827): modal backdrops are bare onClick divs with no role/aria-modal and no visible close control. Add role="dialog" aria-modal="true" on the inner dialog, role="presentation" on the backdrop, and a visible close button; Escape-to-close would be a bonus.

PR #2127, opened 2026-07-04
Persist per-question breakdown in wordplay training log

### LINKED ISSUE #2114: Wordplay training log: persist per-question breakdown, not just round aggregates
Discovered while working issue #2097.

Issue #2097's design section (point 2) asked for the training log to keep
per-question results, not just round-level aggregates, so a future progress
dashboard can show which individual wordplay prompts were missed:

> Per-question results worth keeping (the drill cache means prompts are
> reusable): include the scored evaluation in the entry payload or a compact
> `questions[]`, matching what POST sessions store, so the progress dashboard
> can trend wordplay accuracy alongside everything else.

#2097 implemented round-level persistence only (module/drillType/questionCount/
correctCount/totalMs — the same contract Morse and every other training-log
entry uses), matching the explicit test plan and scope for that issue. Storing
a per-question breakdown would require:

- Extending `trainingEntrySchema` in `server/lib/postValidation.js` with an
  optional `questions[]` (or similar) field.
- Updating `submitTrainingEntry`/the training-log storage in
  `server/services/meatspacePostTraining.js` to persist it.
- Deciding what the (currently nonexistent) progress dashboard would do with
  it, since no consumer reads per-question detail yet.

Low priority — round-level accuracy already surfaces correctly in
`getTrainingStats`. Pick this up once a progress-dashboard UI actually wants
per-question drill-down.

atomantic:
## Summary

Follow-up to #2097, which shipped round-level-only training-log persistence for the standalone Wordplay tab (`questionCount`/`correctCount`/`totalMs`). This PR adds the per-question breakdown that #2097's design section originally called for:

- `trainingEntrySchema` (`server/lib/postValidation.js`) gains an optional `questions[]` field (`prompt`, `response`, `items`, `responseMs`, `score`, `feedback`, `correct`), field-named to mirror the existing `llmResponseSchema` shape scored POST sessions already store — so a future progress dashboard can reuse one renderer for both.
- `submitTrainingEntry` (`server/services/meatspacePostTraining.js`) persists `questions` only when the caller supplies an array (`Array.isArray` gate, not truthiness), keeping legacy/no-breakdown entries indistinguishable from before this change.
- `WordplayTrainer.jsx`'s round-completion handler threads its already-computed `results` (per-question prompt/response/score/feedback) into the new field, computing `correct` from the same `LLM_TRAINING_CORRECT_THRESHOLD` (70) used for the round's `correctCount`.

No consumer reads the per-question detail yet (per the issue, the progress dashboard doesn't exist) — this PR is the storage-layer half only.

## Test plan

- `cd server && npm test` — full suite passes except pre-existing Postgres-dependent failures unrelated to this change (this worktree has no `portos_test` DB provisioned); the touched suites (`postValidation.test.js`, `meatspacePostTraining.test.js`) pass in isolation.
- `cd client && npm test` — full suite passes (299 files, 3226 tests), including new/updated `WordplayTrainer.test.jsx` assertions on the submitted `questions[]` payload.

Closes #2114


atomantic:
CI's \`test (24.x)\` job failed on a known unrelated flake: \`GitTab.test.jsx > GitTab modal accessibility (issue #1090) > closes the diff dialog on Escape\` (1 failed / 3250 passed). This file isn't touched by this PR's diff, and this exact test has flaked intermittently on unrelated PRs before. Merging — full suite passes locally and lint/other jobs are green.

Issue #2220, opened 2026-07-05
CDO: add creative render tools that need route-level persistence (comic cover, etc.)


atomantic:
Discovered while working #2183 (CDO Phase 1 creative tool registry).

The creative tool registry deliberately omits `pipeline_enqueueComicCover`. The bare `enqueueComicCover()` service only queues the render job — the completed cover is attached by the filename hook **only if** the issue's active cover slot already carries the returned `jobId`, and that slot write lives in the route factory (`routes/pipeline/covers.js#makeCoverRenderHandler`: `buildRenderSlot` + `slotKeyForVariant` + `issuesSvc.updateStageWithLatest` + the script-gate patch), not in the service. Wrapping the bare service in a creative tool would silently drop orchestrated covers.

**Task:** extract the enqueue+persist cover-render flow from `makeCoverRenderHandler` into a shared service entry point (e.g. `renderComicCover(issueId, opts)` returning `{ jobId, slot, ... }` and persisting the slot), have the four cover-render routes call it, then add `pipeline_enqueueComicCover` (and any sibling volume/back-cover render tools) to `server/services/creative/tools/pipeline.js` as a conductor over that service — so both the route and the orchestrator share one code path.

Do the same audit for any other creative-suite render/generate action whose route handler performs slot/id persistence the bare service omits (the universe render preflight was already inlined into `universe_renderUniverseJobs` in #2183).

Context: `docs/plans/2026-07-04-creative-director-orchestrator.md`, part of epic #2182.

atomantic:
✅ Primary deliverable shipped in #2234: the four cover-render creative tools (`pipeline_renderComicCover`, `renderComicBackCover`, `renderVolumeCover`, `renderVolumeBackCover`) plus the shared `renderComicCover`/etc. enqueue+persist entry points in `visualStages.js` — routes and orchestrator now share one code path.

The remaining audit scope (wrapping the **comic page-render** and **refine-render** route handlers as tools) is a distinct render surface and has been split into **#2241**. Closing this as done; the remainder is tracked there.

Issue #2182, opened 2026-07-04
Creative Director as general creative orchestrator: tool registry, production plans, autopilot bridge (epic)


atomantic:
Design record: `docs/plans/2026-07-04-creative-director-orchestrator.md`. Sibling epic to #2164 (the quality loop this orchestrator will eventually drive).

Promote the Creative Director from a video producer into the **general creative orchestrator**: give it a directive ("produce a 6-issue noir comic in universe X with covers, a polished manuscript, and a teaser trailer") and let it plan and execute using ANY creative-suite tool — Universe Builder, Story Builder, Writers Room, the Series Pipeline + Autopilot, image/video/music generation — including running the Series Autopilot as one step of a larger production.

Architecture audit (2026-07-04) findings this design builds on:

- CD and Series Autopilot already share one architecture: pure next-step resolver over canonical state + idempotent coordinator + in-memory dedup + SSE/marker observability + boot recovery. The generalization extends this shape.
- The CD's cognitive steps already run as CoS agent tasks (`agentBridge.js` → `task:ready` → spawner → completion hook). Tasking it with a tool registry is the same mechanism with a richer prompt.
- `startSeriesAutopilot(seriesId, options)` is directly invokable in-process (cos-domain-gated internally); only progress observation needs a small event-emitter extension.
- The voice-tools registry + palette whitelist establish the "single source of action schemas" pattern to follow; the #842 Writers-Room→CD bridge establishes the cross-tool tasking pattern (mint a fresh seeded record, link back, never auto-start).
- **Governance gap:** only Autopilot checks an autonomy domain and only WR live mode checks a budget — Universe/Story/Writers-Room/stage generators assume a direct user action. The orchestrator introduces a single gated dispatch chokepoint rather than guarding every call site.

## Phases

- [ ] #2183 — Phase 1: Creative Tool Registry + gated dispatch (creative autonomy domain, budget, ledger)
- [ ] #2184 — Phase 2: production plans — directive, planner agent (cd-plan), generalized advance loop *(blocked by #2183)*
- [ ] #2185 — Phase 3: CD ⇄ Series Autopilot bridge — both directions *(blocked by #2183; CD→autopilot half also wants #2184)*
- [ ] #2186 — Phase 4: studio UI — directive composer, plan board, blocked-step triage *(blocked by #2184)*

Parked: parallel plan steps; scheduled directives (composes with #2174); "polish to READY" deliverable (needs CWQE #2171).

Close when all child issues are done.

[...]

PR #2244, opened 2026-07-07
refactor: wrap comic page-render + refine-render as creative tools (route-level persistence) (#2241)

### LINKED ISSUE #2241: CDO: wrap comic page-render + refine-render route handlers as creative tools (route-level persistence)
Split out from #2220 (cover-render tools shipped in #2234). The cover-render half of #2220 is done; this is the remaining route-persisted render surface the original issue's audit clause called for.

**Task:** the same treatment #2234 gave covers — extract the enqueue+persist flow from the **comic page render** and **refine-render** route handlers into shared service entry points, then wrap them as creative tools in `server/services/creative/tools/pipeline.js` so the Creative Director orchestrator and the routes share one code path.

Audit target: any route handler under `server/routes/pipeline/` whose bare service omits the slot/id persistence the route performs (the filename hook only attaches the finished image if the target slot already carries the returned `jobId`) — mirror the cover pattern (`visualStages.js#renderComicCover` + `makeCoverRenderHandler`).

Context: `docs/plans/2026-07-04-creative-director-orchestrator.md`, epic #2182. Refs #2220.

atomantic:
Closes #2241

## Summary
Completes the #2220 audit clause the covers slice (#2234) deferred: gives the **comic page render** and **refine-render** route handlers the same treatment #2234 gave covers. The enqueue+persist flow moves out of the route handlers into shared service entry points, then those are wrapped as Creative Director tools so the routes and the CDO orchestrator share ONE code path.

Mirrors the `renderComicCover` / `makeCoverRenderHandler` structure exactly.

### Service (`server/services/pipeline/visualStages.js`)
- New `persistComicPageSlot(issueId, pageIndex, {...})` — shared helper that splices the in-flight render slot onto `stages.comicPages.pages[pageIndex].{proofImage|finalImage}` through the serialized `updateStageWithLatest` issue write tail (404s on an out-of-range index).
- New `renderComicPage(issueId, options)` — enqueue (`enqueueVisualComicPage`) + persist in one call.
- `refineComicPageRender` now persists its slot too and returns `{ issue, stage }`.

### Routes (`server/routes/pipeline/issues.js`)
- `POST .../pages/:pageIndex/render` and `.../refine-render` now delegate to the shared entry points (route keeps its up-front page-existence pre-flight + Zod validation + response shaping). Removed the now-duplicated inline enqueue+persist and the now-unused `enqueueVisualComicPage` / `buildRenderSlot` / `slotKeyForVariant` imports.

### Creative tools (`server/services/creative/tools/pipeline.js`)
- New `pipeline_renderComicPage` and `pipeline_refineComicPageRender` tools (COST_RENDER, longRunning), with option shapes mirroring the route Zod schemas. Registered automatically via `PIPELINE_TOOLS`.

The persist runs inside the write tail's compute step, so a concurrent page edit or sibling render can't be reverted by a stale snapshot. The filename hook only attaches the completed image when the slot already carries the returned jobId — the reason this write must live behind the shared entry points, not only in the route.

## Test plan
- `server/services/pipeline/visualStages.test.js` — new `renderComicPage (enqueue + persist)` block (proof slot, final+fromProof, 404); refine happy-path now asserts the persisted slot.
- `server/services/creative/toolRegistry.test.js` — new dispatch tests for both tools (charges one render action; Zod gate rejects missing `pageIndex` / `instruction`).
- `server/routes/pipeline.test.js` — mock rewired to drive the real write tail through the new service entry points; page-render/refine-render persistence assertions preserved.
- Full affected suites green: 446 passed (visualStages, toolRegistry, routes/pipeline, seriesAutopilot).

Refs #2220. Context: `docs/plans/2026-07-04-creative-director-orchestrator.md`, epic #2182.


atomantic:
Review gate (claude + codex): **no blocking issues**. Two non-blocking consistency notes, deferred as out-of-scope for this mirror PR:

- **Deferred (consistency):** the CDO page-render tool `width`/`height` use `z.number().int().positive()`, looser than the route's `imageEdgeSchema` + `refineImagePixelCap` pixel cap. This is *intentionally consistent* with the already-shipped `coverRenderOptionsShape` (#2234) — the CDO render tools share one loose numeric shape. Tightening only the page tools would make them stricter than their sibling cover tools; the right fix is to tighten all CDO render-tool schemas uniformly (importing the route's `imageEdgeSchema`/pixel-cap into `creative/tools/pipeline.js`), which reaches into #2234's shipped covers and belongs in a dedicated follow-up. Enqueue itself still clamps downstream; worst case is a tool call the route would 400.
- **Deferred (cosmetic):** `referencePage` is advertised as `{ type: 'string' }` in the tool's OpenAI `parameters` while its Zod schema accepts string|int; the parameter description already documents the numeric-index option and Zod is the real gate.

Issue #2277, opened 2026-07-08
Layered Intelligence: UI to create http/cmd custom telemetry sources in the Edit App drawer


atomantic:
Follow-up from the custom-source backend slice of #2262.

The backend + Zod schema now accept `http` and `cmd` custom telemetry sources (in addition to `file`), and `gatherSources` reads them — see `server/services/layeredIntelligence.js` (`fetchHttpSource`, `runShellCommand`, `customSourceKey`) and the discriminated union in `server/lib/validation.js`. But the Edit App drawer's Intelligence tab (`client/src/components/apps/LayeredIntelligenceTab.jsx`) only offers UI to add/edit `file` rows; `http`/`cmd` sources are currently only settable via the API.

The tab's `sanitizeCustom` already *preserves* API-set `http`/`cmd` rows non-destructively, so this is purely additive UI work:

- Add a per-row type selector (`file` / `http` / `cmd`).
- Render the type-appropriate field: repo-relative `ref` (file), `url` (http), `cmd` string (cmd), plus the optional `label`.
- Extend the "Add source" affordance to create each type.
- Update `LayeredIntelligenceTab.test.jsx` for the new row rendering/editing.

Refs #2262.

Issue #2262, opened 2026-07-07
Layered Intelligence Loop — perpetual per-app self-improvement via CoS


atomantic:
## Summary

A perpetual, **per-managed-app** self-improvement loop that runs via the Chief of Staff. On a schedule it reads each app's goals + telemetry, asks a reasoning model (default: local LLM) for the single most-valuable improvement, and a **deterministic handler** files that as one tracker issue (GitHub / GitLab / Jira) for a coding agent (`plan-task` / `claim-issue`) to pick up later.

The reasoning model **never touches code** — it returns structured JSON only; every side effect (issue filing, dedup, pause, self-extension) is deterministic handler code running outside the model invocation.

Full design: [`docs/plans/2026-07-07-layered-intelligence-loop.md`](../blob/main/docs/plans/2026-07-07-layered-intelligence-loop.md)

## Architecture

**Engine B (autonomous script job)** — mirrors `goal-check-in`. A single global `type: 'script'`, `category: 'layered-intelligence'` job (off by default; sanctioned scheduled automation under the no-cold-bootstrap AI policy) sweeps `getActiveApps()` and processes each app with `app.layeredIntelligence.enabled`, honoring a per-app `intervalMs`. Chosen over Engine A (CLI coding agent) because "reasoning model must not make code changes; deterministic handler acts" is Engine B's contract by construction.

## Four layers (per app)

1. **Gather** — GOALS.md, CoS metrics (`learning.json`/`productivity.json`/insightsService/`HEALTH_REPORT.md`), plus optional PLAN.md + open issues (config toggles), plus app-declared custom `file` sources.
2. **Reason** — `callProviderAISimple` (default `getActiveProvider()` local LLM; per-app override) with injected free-text "rules"; returns a ranked decision as JSON.
3. **Decide** — deterministic dedup against live tracker (label + slug); pick the single highest-value item.
4. **Act** — deterministically file exactly ONE issue, OR record a pause, OR file a meta/data-gap issue; update `lastRunAt`.

## Memory = the live tracker (no local tracked-issue file)

- **Dedup** — filed issues carry a `layered-intelligence` label + stable `slug` (embedded `<!-- lil-slug: <slug> -->`). Open issues feed the reasoner; a deterministic `gh issue list --search <slug>` guard suppresses duplicates.
[...]

PR #2280, opened 2026-07-08
feat([issue-2262]): custom http/cmd telemetry sources for the Layered Intelligence loop

atomantic:
## Summary

One slice of the open Layered Intelligence Loop epic (#2262). The loop's per-app custom telemetry sources were limited to repo-relative `file` reads (v1); this adds two more source kinds so an app can feed richer signal into the reasoning prompt:

- **`http`** — a time-boxed GET whose response body is fed in (e.g. a CI coverage endpoint or status JSON). Routes through the repo's canonical SSRF-guarded fetch (`fetchPublicBinary`): blocks loopback / link-local / cloud-metadata targets while allowing LAN/Tailscale hosts (a single-user tool legitimately reaches its own network services), revalidates redirects, and byte-caps the body while streaming.
- **`cmd`** — a shell command run in the app repo whose stdout is fed in (e.g. `git log --oneline -20` or a metrics script). Runs detached so the whole process tree is SIGKILLed on timeout or output-cap — a child that keeps stdout open can't hang the sweep — and settles deterministically on close/cap/timeout.

Each source takes an optional `label` (its key in the prompt, falling back to the ref/url/cmd). Every kind treats a blocked / dead / non-zero-exit / empty result as an omitted source (never throws, never stalls the sweep) and is clamped to 8000 chars.

- **Schema:** the per-app config `custom` array is widened from `type: 'file'` only to a Zod discriminated union — `file` keeps its repo-relative path-confinement refine, `http` requires an http(s) URL, `cmd` a non-empty command.
- **Client (non-destructive):** the Edit App drawer's Intelligence tab still only *creates* `file` rows, but it now renders any API-set `http`/`cmd` sources read-only (type-badged chip showing the url/cmd) instead of as blank file inputs, and its save path preserves them instead of coercing to `file` — so an API-declared source is never mistaken for an empty row and deleted, nor silently wiped on save.

Follow-up filed for the UI to *create* http/cmd rows: #2277.

Refs #2262.

## Test plan

- `server/services/layeredIntelligence.test.js` — `customSourceKey` per-type keying + label preference; `fetchHttpSource` routes through the guarded fetch (asserts `throwOnUnsafe:false`, null on a blocked/failed URL, maxBytes slice, fetchMaxBytes forwarding); `runShellCommand` caps runaway stdout + kills and reports success, real exit code for a short command; `gatherSources` http/cmd inclusion/omission/clamping.
- `server/lib/validation.test.js` — discriminated union accepts file/http/cmd; rejects file traversal ref, non-URL http, non-http(s) scheme, and unknown type.
- `client/src/components/apps/LayeredIntelligenceTab.test.jsx` — preserves http/cmd through a no-op save and a toggle-triggered emit; renders http/cmd read-only (no editable file input).
- Full server suite green (17355 passed); touched client suites green.


atomantic:
Superseded by #2277 (merged in PR #2279), which shipped the same Layered Intelligence `http`/`cmd` custom telemetry sources — UI + backend — while this PR was in its review loop. The two were built in parallel from the same follow-up (#2277 was filed by this issue's slice, then independently claimed and merged).

Main's implementation already covers the runtime-safety hardening this PR added: `fetchHttpSource` validates the http(s) scheme with a 10s timeout, and the `cmd` path runs through the shared `bufferedSpawn`, which caps output and kills the whole process tree on timeout. The only net difference here is the SSRF-guarded `fetchPublicBinary` fetch — deliberately not adopted, since PortOS's single-user private-network trust model intentionally omits SSRF defenses (CLAUDE.md), and host-allowlisting a *user-configured* telemetry URL would break the legitimate localhost/LAN endpoint case rather than protect anything.

Closing as superseded. #2262 stays open for its remaining slices (Jira filer, semantic dedup, dedicated route/page). The diff remains viewable here if the SSRF-guard delta is ever wanted.

Issue #2322, opened 2026-07-09
Migrate Layered Intelligence loop to a per-app deterministic scheduled task


atomantic:
## Why

The Layered Intelligence (LI) loop shipped as a **single global autonomous job** that sweeps all enabled apps on one daily fire, with a global on/off under CoS → System Tasks and per-app config under Edit App → Intelligence. This causes:

1. **Confusing global/per-app split** — a per-app card shows "Enabled · Due now" but nothing runs, because the global sweep already fired that day (or is off). Reported: enabling PortOS showed "Due now" with no agent/logs, because the daily global sweep had already run at 04:25 before the app was enabled.
2. **It's a bespoke third surface** instead of behaving like the other per-app self-improvement work (CoS → Schedule), which each have their own per-app interval, provider, and "run now."

**Decision:** migrate LI to run **per-app, one app's context at a time, like the other scheduled tasks.** Drop the cross-app sweep.

## Key finding

The scheduled-task system (`taskSchedule.js` / `SELF_IMPROVEMENT_TASK_TYPES`) is hardwired to **spawn one coding agent per fire** — it has no deterministic-handler path. LI is deterministic (model returns JSON only; filing is code). So the migration's core work is **adding a per-app deterministic *handler-backed* task type** to the scheduled-task layer — a new execution primitive in core CoS autonomous-spawn infra. `processApp(app)` is already per-app; only the `runLayeredIntelligence()` sweep wrapper loops apps (delete it).

## Design record

Full design (execution primitive, config home = option A, job-retirement migration, touchpoint inventory, open questions) in `docs/plans/2026-07-08-layered-intelligence-scheduled-task-migration.md`.

## Scope
- Add `HANDLER_BACKED_TASK_TYPES` + lazy handler registry; additive fire-and-forget branch in `queueEligibleImprovementTasks` + on-demand drain (never touches the agent-spawn path).
- Register `layered-intelligence` in `SELF_IMPROVEMENT_TASK_TYPES` / `DEFAULT_TASK_INTERVALS` / `TASK_TYPE_DESCRIPTIONS`.
- Refactor `processApp` to gate-free per-app work; scheduler owns enabled/due + cadence.
- Retire the global `job-layered-intelligence`; migration moves per-app enabled/interval/provider into `taskTypeOverrides['layered-intelligence']` (enabled = globalJobEnabled && li.enabled), leaves sources/scopes/rules in `layeredIntelligence`.
[...]

PR #2324, opened 2026-07-09
feat([issue-2322]): migrate Layered Intelligence to a per-app scheduled task + remove dedicated page

atomantic:
Refs #2322

## Summary

Replaces the single **global** "Layered Intelligence Loop" autonomous-job sweep with a **per-app, deterministic, handler-backed scheduled task** in the `taskSchedule` system — so the loop runs one app's context at a time on its own per-app schedule, like every other self-improvement automation. This fixes the confusing split that produced *"Enabled · Due now but nothing runs"* (a global daily fire decoupled from per-app enablement).

### What changed
- **New primitive — handler-backed scheduled tasks.** `HANDLER_BACKED_TASK_TYPES` in `taskSchedule.js`. When the per-app scheduler picks a due handler-backed type, it runs the deterministic handler **in-process** (fire-and-forget, guarded by a per-app in-flight `Set`, records execution on completion) and **never** touches the agent-spawn path. Mirrored in the on-demand ("Run now") drain, with a no-app guard so a global request can't fall through to the agent path.
- **LI is now the `layered-intelligence` task type** (per-app enable / interval / provider / model in the Schedule task override; behavior — sources, scopes, rules, hand-off — stays on the app's Intelligence tab). `processApp` → `runLayeredIntelligenceForApp`, gate-free (the scheduler owns enabled/due).
- **Retired the global job** (`job-layered-intelligence`): removed the DEFAULT_JOBS entry + SCRIPT_HANDLERS registration + the `runLayeredIntelligence()` sweep.
- **Migration 184** (idempotent): tombstones the old job, moves each app's scheduling into `taskTypeOverrides['layered-intelligence']` (enabled only when the old global job **and** the per-app loop were both on — faithful effective state), flips the task-type master switch on when the job was enabled, and leaves behavior + `lastRunAt` on `app.layeredIntelligence`.
- **Overview route + Intelligence tab** re-pointed off the removed job; the provider picker keeps the all-types fix from #2323.

### Behavior change to note
Because LI is now a self-improvement task, it runs only while **CoS improvement is enabled** (plus the task-type master switch and the per-app override) — the standard three-part gate every self-improvement task uses, replacing the old bespoke global on/off.

### Deferred (open question on #2322)
The dedicated `/layered-intelligence` page + nav entry are **kept as-is** (they render against the re-pointed overview payload). Whether to remove/repurpose that page now that LI lives in CoS → Schedule is an open UX decision — left on #2322, which stays open.

## Test plan
- `cd server && NODE_ENV=test npx vitest run services/autonomousJobs services/layeredIntelligence.test.js services/taskSchedule.test.js routes/apps.test.js services/cos.test.js services/apps.layeredIntelligence.test.js` — 488 passed, 1 skipped. New: a `cos.test.js` case asserting a handler-backed due type runs the handler and does **not** enqueue an agent task.
- Full server suite (excluding `*.db.test.js`) — 18163 passed, 147 skipped.
- Migration `scripts/migrations/184-…test.js` — 7 passed.
- `cd client && npx vitest run …EditAppDrawer/LayeredIntelligenceTab/LayeredIntelligence` — 39 passed; full client suite — 3477 passed.
- No DB-backed suites run.

Not auto-merged — the core scheduler seam + a data migration warrant a human review pass.


atomantic:
Added a second commit removing the dedicated `/layered-intelligence` page (your decision — manage LI in CoS → Schedule like any other task). This completes the migration, so **this PR now Closes #2322 and #2325**.

Note the `Refs #2322` in the original body should read **Closes #2322** — I'll rely on the commit trailers (`Closes #2325`) + will confirm #2322 closes on merge.

Issue #2342, opened 2026-07-10
Auto-fix: wire full 4-tier fallback execution (attempt each tier's fix in sequence)


atomantic:
Follow-up to #2328. The tier classifier (`classifyFixTier` in `server/services/autoFixer.js`, #2331) currently only **annotates** the investigation task with a fallback tier + strategy — it does not actually **attempt** each tier's fix in sequence before escalating.

## Scope
Wire deterministic per-tier execution so a failure is routed to the cheapest tier that can plausibly recover it, escalating to an investigation task only when no deterministic fix applies:

- **Tier 1 — config/env**: attempt a config/env correction (wrong provider/model/key/path) before spawning an agent.
- **Tier 2 — schema/type**: attempt a schema/type/format correction for malformed request/response.
- **Tier 3 — constrained-agent-retry**: a bounded retry (already partially covered by the promptRunner fallback path — reconcile with `noteFallbackStarted`/`noteFallbackHandled`).
- **Tier 4 — escalate**: the current investigation-task behavior.

## Already shipped (do not redo)
- Circuit breaker (#2330)
- Tier classifier + `buildFixDiagnostics` (#2331)
- Diagnostics persisted as `metadata.diagnostics` (#2335)
- **Aggregation + dashboard telemetry** (this PR / the aggregation slice of #2328) — `server/services/autoFixMetrics.js`, `GET /api/autofix/metrics`, Auto-Fix Telemetry widget, `observedAt` time-to-recovery timestamp.

## Acceptance
- Each tier's fix is attempted in order; only unrecovered failures create a Tier-4 investigation task.
- The existing telemetry (`autoFixMetrics.js`) automatically reflects per-tier recovery once tiers 1–3 start resolving failures without a human.
- 30-day >90% auto-fix success measurement window (original #2328 acceptance criterion).

Refs #2328

atomantic:
Shipped ✓ via PR #2348 — Tiers 1 (config/env same-provider model correction), 3 (constrained-agent-retry), and 4 (escalate) now execute in sequence, reconciled through the `noteFallbackStarted/Handled/Failed` lifecycle so every full recovery creates zero investigation tasks and every give-up path creates exactly one. Tier-2 (schema/type) deferred to unknown shorthand flag: 'q' in -q

Usage:  gh issue create [flags]

Flags:
  -a, --assignee login   Assign people by their login. Use "@me" to self-assign.
  -b, --body string      Supply a body. Will prompt for one otherwise.
[...]

Issue #2328, opened 2026-07-09
Harden auto-fix with deterministic fallback chains and failure diagnostics


atomantic:
## Context
`auto-fix` telemetry shows a 43% success rate, indicating unreliable self-repair mechanisms that force manual intervention and break autonomous loops.

## Objective
Replace brittle auto-fix attempts with a deterministic, tiered fallback chain and comprehensive failure diagnostics to push success rates >90% and restore autonomous self-healing.

## Implementation Steps
1. **Add Failure Diagnostics**: Log structured context for every auto-fix attempt: trigger event, target file/resource, error type, fix strategy used, and exact failure reason (e.g., syntax error, runtime exception, API rate limit, schema mismatch).
2. **Tiered Fallback Chain**: Implement a strict cascade on failure:
   - Tier 1: Configuration/Env fix (retries with exponential backoff, env var correction)
   - Tier 2: Schema/Type fix (regenerate strict type hints, apply structural patches)
   - Tier 3: Agent retry with constrained prompts (reduce scope, add explicit constraints, force validation pass)
   - Tier 4: Graceful degradation or explicit manual escalation with context dump
3. **Prevention Guardrails**: Add a circuit breaker if the same resource fails auto-fix >3 times within 1 hour to prevent infinite retry loops or resource exhaustion.
4. **Telemetry Updates**: Track `auto-fix` by tier, failure reason, and time-to-recovery. Expose aggregated metrics in the dashboard with success rate trend.

## Acceptance Criteria
- Auto-fix success rate consistently >90% over a 30-day window
- Zero silent drops; all failures logged with diagnostic context
- No infinite loops or resource exhaustion via circuit breaker implementation
- Dashboard reflects granular auto-fix metrics

atomantic:
Partial delivery shipped via #2330 (merged).

**Done ✓**
- Guardrail #3 — **circuit breaker**: the auto-fixer now trips a per-resource circuit after >3 task-worthy failures within a rolling 1-hour window and auto-closes once failures age out, so a persistently-failing provider/model or critical error can no longer flood the plan queue with identical investigation/fix tasks. Only failures that actually produce a task count (fallback-recovered failures don't). Covered by new tests in `server/services/autoFixer.test.js`.

**Remaining ▢**
- Tier 1–4 fallback cascade (config/env → schema/type → constrained-agent retry → graceful escalation)
[...]

PR #2348, opened 2026-07-10
feat: wire deterministic Tier-1 config/env fix into the auto-fix fallback cascade (#2342)

atomantic:
## Summary

Follow-up to #2328 / #2331. The tier classifier (`classifyFixTier`) only *annotated* failures with a fallback tier — the runner still did just one thing on failure (swap to a fallback provider) and otherwise queued a Tier-4 investigation task. This wires **deterministic per-tier execution** into `runPromptThroughProvider` so a failure is routed to the cheapest tier that can recover it before escalating.

New cascade in the failure path:

- **Tier 1 — config/env**: a `model-not-found` / `model-not-supported` failure first gets a deterministic correction — a bounded retry of the **same** provider with a valid model it lists (`pickConfigCorrectedModel`, pure; no persisted config change; embedding-only ids skipped). Cheaper than benching the provider.
- **Tier 2 — schema/type**: no safe deterministic same-provider transform exists at the runner layer (a malformed request/response can't be rewritten without the caller's schema), so the cascade falls straight through. Documented in-code; deferred (see Remaining).
- **Tier 3 — constrained-agent-retry**: the existing fallback-provider retry, now reached only after Tier 1 declines/fails.
- **Tier 4 — escalate**: the existing deferred investigation task, created only for genuinely unrecovered failures.

Recovery at any tier is reconciled through the **existing** `noteFallbackStarted` / `noteFallbackHandled` / `noteFallbackFailed` lifecycle, so a Tier-1 recovery suppresses the investigation task exactly like a Tier-3 one:

- A fully-recovered failure leaves **zero** investigation tasks (a Tier-1 corrected retry that fails, then Tier-3 recovers, cancels *both* queued keys).
- Only unrecovered failures escalate to a Tier-4 task; the give-up branch releases Tier-1 suppression so the in-flight set doesn't leak.
- The recovering tier is surfaced on the result (`fixTier` / `fixStrategy`) and in the single-line log (`🔧 Tier 1 (config/env) retry …`).

Because fewer failures reach a Tier-4 task, the existing #2328 telemetry (`autoFixMetrics.js` / Auto-Fix Telemetry widget) **automatically** reflects the per-tier recovery — no telemetry change needed.

Also fixes a latent footgun caught in self-review: guarding on `ERROR_CATEGORIES.MODEL_NOT_SUPPORTED` (a member that doesn't exist → `undefined`) would have matched every category-less failure and wrongly engaged Tier 1. Now compared against string literals with a regression test.

## Changes

- `server/lib/promptRunner.js` — `pickConfigCorrectedModel` pure helper; tiered cascade in `runPromptThroughProvider`'s catch; `fixTier`/`fixStrategy` on the recovery result.
- `server/lib/promptRunner.test.js` — pure-helper tests + full cascade coverage (Tier-1 recovery, decline→Tier-3, Tier-1-fail→Tier-3 cancels both tasks, no-fallback give-up releases suppression, non-model category skip, category-less regression).
- `.changelog/NEXT.md` — entry.

## Test plan

- `cd server && npm test -- promptRunner` → **108 passed** (98 pre-existing + 10 new).
- `cd server && npm test -- autoFix` → **46 passed** (autoFixer + autoFixMetrics unchanged and green).
- Full `npm test` failures in this environment are all pre-existing PostgreSQL-connection errors (no DB in sandbox); none touch promptRunner/autoFix.

Refs #2342

## Remaining

- **Tier 2 (schema/type) deterministic correction** is deferred: no safe same-provider transform is available at the runner layer (rewriting a malformed request/response needs the caller's schema). It currently falls through to the Tier-3 bounded retry. A future slice could add per-caller schema-aware correction hooks.

https://claude.ai/code/session_0122z1P7dkTb6oocCVAHTw28


atomantic:
## Review gate complete (reviewers: claude, codex)

Self-review + headless CLI reviews run to convergence. All findings were applied in-branch:

- **claude (round 1):** fixed a task-swallow on a Tier-1 pre-execution-throw + no-fallback give-up (new `escalateProviderFailure`), a suppression leak if `coalesceFallbackMarkAndPick` throws (try/finally release), and keyed the corrected-retry task on the run's *effective* provider/model.
- **codex (round 1):** pre-suppress the corrected-retry's own backstop task (slow-Tier-3 race), and gate Tier 1 on `providerHonorsModelOverride` so a baked-model CLI can't pointlessly retry the same model.
- **codex (round 2):** dedupe explicit escalations (concurrent no-fallback storm → one task), and escalate on a Tier-3 fallback *pre-execution* throw.
- **codex (round 3):** the `finally` now escalates one deduped task whenever it finds a suppressed-but-unresolved key (unexpected-throw signal, e.g. fallback-lookup throwing).
- **codex (final):** clear the escalation dedupe marker when task creation fails so a retry isn't masked.

Every give-up path now yields **exactly one** investigation task; every full recovery yields **zero**. Tests: `promptRunner` + `autoFix` suites → **163 passed**. (Full-suite failures in the sandbox are pre-existing PostgreSQL-connection errors, unrelated to this diff.)

PR #2624, opened 2026-07-15
Add an adaptive render budget and visibility-aware frameloop to City (#2592)

### LINKED ISSUE #2592: Add an adaptive render budget and visibility-aware frameloop to City
## Problem / Goal

Keep City smooth across different GPUs and device classes without requiring users to manually discover the correct quality preset.

City currently defaults to High, runs an always-on render loop, and has no measurement-driven way to respond to sustained frame pressure.

## Context

- `client/src/hooks/useCitySettings.js:9` defines four fixed presets and `:45` defaults new settings to High.
- `client/src/components/city/CityScene.jsx:85` temporarily reduces startup cost for 1.2 seconds, but returns to the selected fixed preset afterward.
- `CityScene.jsx:144` caps DPR, but does not adapt it from observed render time.
- `CityScene.jsx:197` uses `frameloop="always"` for the normal City.
- The City component directory currently registers 59 `useFrame` callbacks across 43 files.
- `client/src/components/city/cityConstants.js:218` uses `particleDensity` as an implicit proxy for unrelated detail gates such as rooftops and interior-window shaders.
- Polling already pauses through `useAutoRefetch` and the shared `useVisibilityEvent`; the WebGL renderer should follow the same visibility lifecycle.

## Proposed approach

1. Add an Auto quality mode:
   - New installs default to Auto, beginning at High.
   - Existing stored payloads without a `qualityMode` field load as Manual and retain their current preset.
   - Manual Low/Medium/High/Ultra remains available and disables adaptation.
2. Add a pure render-budget state machine, preferably `client/src/utils/cityRenderBudget.js`, with the required barrel export, catalog entry, and unit tests.
3. In Auto mode:
   - Ignore the existing startup warm-up period and frame gaps above 250ms.
   - Evaluate two-second rolling windows using p75 frame time.
   - Step down one tier after two consecutive windows above 25ms.
   - Step up one tier after five consecutive windows below 18ms.
   - Apply a ten-second cooldown after every tier change.
   - Never change more than one tier per decision.
4. Keep runtime effective quality separate from persisted user settings. Display `AUTO · HIGH`, `AUTO · MEDIUM`, etc. in the Performance settings tab.
5. Replace `particleDensity`-as-quality-proxy with an explicit effective detail tier. Continue deriving particle count, reflections, DPR, interior windows, and optional set dressing from that tier.
6. Pause the normal City frameloop while `document.visibilityState === "hidden"` and resume with the existing startup warm-up when visible again. Preserve Photo mode’s demand loop and capture behavior.
7. Add a local-only diagnostics readout in settings showing effective tier, rolling FPS, and p75 frame time. Do not persist or transmit performance samples.

## Acceptance criteria

- [ ] Auto mode follows the specified thresholds, hysteresis, and cooldown without oscillating between adjacent tiers.
- [ ] Runtime adaptation does not rewrite localStorage on every tier change.
- [ ] Existing saved settings remain Manual and retain their preset after upgrade.
- [ ] Manual presets never auto-adjust.
- [ ] Hidden tabs stop the live City render loop; returning visible resumes and renders a fresh frame.
- [ ] Photo mode, playback, context-loss recovery, and screenshot capture retain their current behavior.
- [ ] The settings UI exposes the effective Auto tier and local diagnostics.
- [ ] Unit tests cover downshift, recovery, cooldown, ignored samples, storage compatibility, manual mode, and visibility transitions.
- [ ] Any new reusable utility is exported and documented according to the client utility catalog rules.

## Out of scope

- Consolidating City’s initial API-request fan-out.
- Splitting the Three.js or City route bundles.
- Rewriting individual district geometry unless profiling identifies a separate concrete bottleneck.

atomantic:
## Summary

Adds an **Auto quality mode** and a **visibility-aware frameloop** to CyberCity so it stays smooth across GPUs without the user hand-picking a preset (issue #2592).

- **Auto quality mode** (Settings → Quality). New installs default to Auto starting at High. Existing saved settings without a `qualityMode` key load as **Manual** and keep their preset; picking a named preset pins Manual. Adaptation never rewrites localStorage, and Manual presets never auto-adjust.
- **Pure render-budget state machine** — `client/src/utils/cityRenderBudget.js` (barrel + README row + unit tests). Evaluates p75 frame time over 2s rolling windows, ignores the startup warm-up and frame gaps > 250ms, steps down after 2 windows > 25ms, up after 5 windows < 18ms, applies a 10s cooldown, and never moves more than one tier per decision. The 18–25ms dead band breaks streaks to prevent oscillation. It is side-effect-free (`now`/`dt` are inputs) so it is fully unit-testable; all impurity lives in the in-Canvas `CityAdaptiveQuality` driver.
- **Effective tier replaces the `particleDensity` proxy.** `cityShowDetail` / `cityShowInteriorWindows` now read an explicit `settings.effectiveTier`, falling back to the legacy particleDensity thresholds when absent. The effective tier (adaptive in Auto, the preset in Manual) drives reflections, particle density, DPR, interior windows, and set dressing.
- **Visibility-aware frameloop.** The live City loop pauses (`frameloop="never"`) while `document.visibilityState === "hidden"` and resumes with a fresh warm-up when visible again. Photo mode keeps its demand loop and capture behavior; context-loss recovery and playback are untouched.
- **Local-only diagnostics.** The settings panel shows the effective tier (`AUTO · HIGH`) plus a live tier / FPS / p75 readout, only while the panel is open. Nothing is persisted or transmitted.

## Test plan

- `client`: `npx vitest run src/utils/cityRenderBudget.test.js src/utils/index.test.js src/hooks/useCitySettings.test.jsx src/components/city/cityTheme.test.js` — 60 passing.
- Broader regression: `npx vitest run src/components/city src/utils` — 42 files / 781 passing.
- `eslint` clean on all changed files.
- New unit tests cover downshift, recovery (upshift), cooldown, ignored samples (warm-up + gap), storage compatibility (fresh→Auto, legacy→Manual, stored-auto preserved), Manual-preset pinning, visibility transitions (restartWarmup), and state-machine purity.

Closes #2592


atomantic:
### Review loop summary

Ran two headless reviewers against the diff. No blockers were reported by either; all findings (P2/P3) were addressed in-branch.

**claude** (1 round) — no blockers. Fixed: warm-up must suppress detail via `effectiveTier:'low'` (not the clamped particleDensity); freeze budget streaks during cooldown to prevent adjacent-tier oscillation; disable the Auto-controlled panel fields; corrected a stale comment.

**codex** (3 rounds, review cap reached) — no blockers across any round; each round surfaced finer edge cases, all fixed:
- Round 1: start the window clock at the first post-warm-up sample; re-arm the budget inside the frame loop (not a passive effect) so it can't race the frameloop resume; show effective (not stale manual) values in disabled panel fields.
- Round 2: discard partial windows on a frame gap; gate cooldown eligibility on the window's *start*; re-arm the runtime budget on RESET DEFAULTS via a reset token.
- Round 3: downshift under sustained sub-cutoff frame times (synthetic pressure window after N consecutive gaps); re-arm the budget for app-count/startup warm-ups; clear diagnostics on quality-mode transitions.

Round-3 fixes were not re-run through codex (3-round cap). Tests: 794 passing across the city + utils + settings suites; eslint clean.

PR #2662, opened 2026-07-16
Lift CoS actionable-insights fetch into ChiefOfStaff.fetchData (#2654)

### LINKED ISSUE #2654: Lift CoS actionable-insights fetch into ChiefOfStaff.fetchData (retire insightsRefreshKey + banner self-poll)
## Background

`ActionableInsightsBanner` (the CoS "blocked task alert bar") owns its own fetch of `/cos/actionable-insights` via `useAutoRefetch` (60s poll), fully decoupled from the task list that `ChiefOfStaff` fetches (30s poll). A recent bug fix bridged the one gap that mattered in practice: deleting a blocked task left a stale "N blocked tasks" alert up. The fix (`ChiefOfStaff.jsx` `insightsRefreshKey` + `refreshTasks`, banner `refreshKey` prop + mount-skipping refetch effect) bumps a counter on the `TasksTab` mutation path so the banner re-derives immediately.

## The remaining altitude issue

The signal only covers **one** trigger of a broader class: "the banner's server-derived insights lag any parent-owned data change until the 60s poll." Other parent paths refetch their own state but leave the banner stale:

- `handleTasksUserChanged` (socket-driven `setTasks`) — no bump
- `handleAgentSpawned` / `handleAgentCompleted` (call `fetchData()` directly) — no bump
- `handleHealthCheck` + socket `cos:health:check` — update `health`, which feeds the banner's "N health issues" insight — no bump
- learning/peak-time updates — no bump

So health/learning-driven insight staleness still exists; only the TasksTab mutation path is covered.

## Proposed refactor

Make `ChiefOfStaff.fetchData` the single owner of actionable-insights fetching:

1. Add `api.getCosActionableInsights()` to `fetchData`'s `Promise.all`, hold `insights` in parent state.
2. Pass `insights` as a prop to a now-presentational `ActionableInsightsBanner` (which keeps only its dismiss/expand/unblock UI state).
3. The banner's unblock path calls `onRefresh` up instead of owning its own `refetch()`.
4. Retire `insightsRefreshKey`, `refreshTasks`'s signal bump, the banner's `refreshKey` prop, the `didMountRef` skip-mount effect, and the banner's `useAutoRefetch` self-poll.

Result: one fetch orchestration point; every trigger that refetches parent data refreshes the banner for free, fixing the whole staleness class (including health/learning), and the 30s/60s cadence skew dissolves.

## Trade-offs / decisions for the implementer

- **Tab-lazy fetching is traded away.** The banner's colocated `useAutoRefetch` only fetches while the tasks tab is mounted. Lifting into `fetchData` makes insights an always-on fetch on every 30s poll and every socket-driven `fetchData`. It's a cheap derived-count call, but decide whether the always-on cost is acceptable (likely yes) or whether to gate the insights fetch on the active tab.
- **Insights are NOT a pure function of the task list** — the endpoint also aggregates health, approvals, and inferred learning/peak-time stats. So a client-side `useMemo` over `tasks` is NOT viable; a separate fetch is warranted. The refactor is "lift the fetch," not "eliminate it."

## Affected files

- `client/src/pages/ChiefOfStaff.jsx`
- `client/src/components/cos/ActionableInsightsBanner.jsx` (+ `.test.jsx`)
- `server/routes/cosInsightRoutes.js` (reference only — endpoint unchanged)

## Acceptance criteria

- Deleting/creating/editing/unblocking a task AND a health-check change both update the banner without waiting for a poll.
- `insightsRefreshKey` and the banner's self-poll are removed.
- Banner retains dismiss/expand/unblock UI behavior and last-good rendering.
- Tests updated to cover prop-driven insights + the unblock-up path.

Deferred from the bug-fix PR that added `insightsRefreshKey` (out of scope for a targeted fix — touches the banner's core data flow and trades away tab-lazy fetching).

atomantic:
## Summary

`ActionableInsightsBanner` (the CoS "blocked task alert bar") owned its own `/cos/actionable-insights` fetch on a 60s `useAutoRefetch` poll, fully decoupled from the parent's 30s task poll — so its server-derived counts (blocked tasks, pending approvals, "N health issues") lagged any parent-owned data change until that poll. A prior fix bridged only the TasksTab mutation path via an `insightsRefreshKey` counter; socket-driven task changes, agent spawn/completion, and health checks all stayed stale.

This makes `ChiefOfStaff.fetchData` the single owner of insights fetching:

- Insights are fetched inside `fetchData`'s `Promise.all` and passed down as an `insights` prop. The banner is now **presentational** — it keeps only dismiss/expand/unblock UI state, and its unblock path calls `onRefresh` up instead of owning a `refetch()`.
- Agent spawn/completion already call `fetchData`, so they refresh the banner for free.
- The two paths that intentionally do a *targeted* `setState` rather than a full refetch — the `cos:health:check` + manual health-check handlers (which preserve their own status messaging) and the `cos:tasks:user:changed` socket handler — call a lightweight insights-only `refreshInsights()`, so a newly-found/cleared health issue or daemon-driven task change reflects immediately instead of on the 30s poll.
- A failed/transient insights fetch (`null`) **preserves the last-good array** (matching the retired poll's behavior); a legitimately-empty `[]` clears the banner. This keeps *not-fetched/failed* distinct from *fetched-and-empty*.

Retires `insightsRefreshKey`, `refreshTasks`'s signal bump, the banner's `refreshKey` prop, its `didMountRef` skip-mount effect, and its `useAutoRefetch` self-poll. The 30s/60s cadence skew dissolves.

## Test plan

- `client/src/components/cos/ActionableInsightsBanner.test.jsx` — rewritten for the prop-driven banner: null/empty gating, prop-driven render, dismiss, and the unblock-up path (asserts `updateCosTask` + `onRefresh` + `onTaskUnblocked`, no self-poll). `insightProvenance` tests unchanged.
- `client/src/pages/ChiefOfStaff.test.jsx` — adds the newly-called `getCosActionableInsights` to the api mock and pins the two socket refresh paths (health-check, task change) re-pull insights without a full `fetchData`.
- `cd client && npx vitest run src/components/cos/ src/pages/ChiefOfStaff.test.jsx` → 106 passed. ESLint clean on all changed files.

Closes #2654


atomantic:
### Review notes (codex loop)

Four review rounds tightened this PR. Fixed along the way:
- **Feedback loop (HIGH):** wiring an on-demand insights refresh into the `cos:health:check` socket handler looped, because `/cos/actionable-insights` runs `cos.runHealthCheck()` which re-emits `cos:health:check`. Removed all on-demand refresh — the banner now refreshes only via `fetchData`.
- **Process-restart side effect (MEDIUM):** the same endpoint **auto-restarts errored PM2 processes** (`cosHealthMonitor.js`), so the manual "Run Check" button's insights refresh fired a second restart ~1s after its own check. Dropped it.
- **Stale/failed/untimed health overwrite (HIGH/MEDIUM):** `fetchData`'s `getCosHealth` reads pre-check health while its insights call triggers a fresh check whose socket emit can land first. `fetchData` now keeps the newer health by `Date.parse(lastCheck)`, preserves last-good on a null read, and won't let a timestamp-less read clobber a timestamped one.

Two remaining findings are **pre-existing and intentionally out of scope** (self-correcting on a single-user install):
- `deriveAgentState`/`statusMessage` derive from `fetchData`'s health read (unchanged by this PR) — these are transient defaults immediately overwritten by socket events.
- Concurrent `fetchData` calls are unordered, so an older response can momentarily restore a stale count. This is a pre-existing property of every value `fetchData` sets (tasks/agents/health/…), not new to insights; the counts are idempotent and self-correct on the next event/poll. Hardening it means generation-tokening all of `fetchData` — a separate change.

Issue #2657, opened 2026-07-16
Autonomous Creation Engine: schedule the Creative Director with a recurring, feedback-steered creative brief


atomantic:
## Summary

Give the **Creative Director** a *standing, recurring creative commission* that runs on a schedule and steers itself with my feedback. I want to say: *"Every night at 2am, make me a video of something surreal"* — and then, after I react to each night's output, have the **next** run absorb that feedback and adjust direction. Over time this becomes an **autonomous creation engine guided by my taste**: a persistent brief + a feedback loop wrapped around PortOS's existing gated creative-tool pipeline.

The commission must be able to target **any** creative ability PortOS supports — image, video, universe, series, story, writers-room work, and music production — not just video.

## User story

> As the single owner of my PortOS install, I create a **Creative Commission**: target = *video*, cadence = *nightly @ 02:00*, brief = *"something surreal, dreamlike, unsettlingly beautiful"*. Each morning a new video is waiting in my inbox/catalog. I rate it (👍/👎 + a note: *"less horror, more Magritte"*). The next night's video is generated **conditioned on that accumulated feedback**, drifting toward what I actually like. I can pause, edit the brief, change the target ability, or fork a new commission at any time.

## Vision / what "done" feels like

- A first-class, persistent **Commission** entity: `{ schedule, targetAbility, brief/genre, feedback-history, autonomy settings }`.
- Runs unattended on cron, produces a real artifact through the **existing governed creative-tool registry** (budget + autonomy-mode enforced).
- Every output is surfaced to me (notification + brain inbox + catalog) with a one-tap rate/annotate affordance.
- Accumulated feedback is **re-injected into the next run's prompt** so direction evolves.
- Works across **all** creative abilities via a target-ability adapter, not a video-only special case.

---

## Current architecture (what we build on)

This feature is mostly **composition of systems that already exist**. The two load-bearing pieces:

### 1. The governed creative-tool pipeline (the "hands")
- **`server/services/creative/toolRegistry.js`** — `CREATIVE_TOOLS` + the single gated chokepoint `dispatchCreativeTool(name, args, ctx)`. Enforces the `creative` autonomy mode (`off | dry-run | execute` via `server/lib/domainAutonomy.js`, `CREATIVE_DOMAIN_ID='creative'`), charges the shared daily `cos` action budget, and writes `server/services/creative/creativeRunLedger.js`.
[...]

PR #2663, opened 2026-07-16
Autonomous Creation Engine (Phase 1): recurring, scheduled Creative Commissions

atomantic:
## Summary

Ships **Phase 1** of the Autonomous Creation Engine (#2657): a first-class, persistent **CreativeCommission** entity — a standing, recurring creative brief that fires on a schedule and drives the existing Creative Director *directive* pipeline unattended ("every night at 2am, make me something surreal"). Phase 1 covers the data model, scheduler, routes, and UI for `video` commissions; the feedback loop, all-ability adapter, music tool, and genre taxonomy are later phases.

### What's included
- **`CreativeCommission` entity** — machine-local `createCollectionStore` collection at `data/creative-commissions/` (schemaVersion 1), with a sanitizer, `data.reference/` seed, and stamping **migration 191**. Deliberately **not** federated in Phase 1: a synced schedule would double-run on every peer (same rationale as `seriesAutopilotScheduler`'s settings-based schedules). The record shape carries forward-stable `feedback[]` + `feedbackWindow` so Phase 2 needs no schema change.
- **`commissionScheduler.js`** — modeled on `seriesAutopilotScheduler`. Arms one `eventScheduler` cron per enabled commission; re-syncs off a store-emitted `commission:changed` event (decoupled from the route, catches any writer). On fire it re-reads the commission + autonomy config, composes a CD directive from the brief via the pure `buildCommissionDirective` (which already folds a last-N feedback digest — Phase 2 only adds the rate surface), mints a directive-driven CD video project, and nudges `advanceAfterPlanStepSettled` so each plan step runs through the gated `dispatchCreativeTool`.
- **Cold-bootstrap compliant** (AI Provider Usage Policy) — boot only *arms* crons; nothing fires until a cadence elapses. Every fire gates on creative autonomy `execute` **and** the daily `cos` budget before spawning the planner LLM, so an `off`/`dry-run` or over-budget install records a `skipped` run and generates nothing.
- **REST CRUD** at `/api/creative-commission` — Zod-validated create/update (DAILY/WEEKLY/CUSTOM schedules composed to cron with an authoritative `isValidCron` check in the service).
- **Create-suite page** `/creative-commission` — list + deep-linked create/edit Drawer (`/:id`, `/new`) with pause/resume, inline-confirm delete, and per-run history. Registered in `NAV_COMMANDS` (⌘K + voice) and the sidebar.
- Phase 1 restricts `targetAbility` to `video` (the enum widens with the ability adapter in Phase 3/4) so a user can't create a commission that silently never runs.

## Remaining (tracked on #2657)
- **Phase 2** — feedback loop: `CommissionFeedback` record, rate/annotate UI on each run's output, notification + brain-inbox surfacing, and federation-gating the schema (splitting the machine-local *schedule* from the federatable *brief/feedback*, since the schedule must stay local to avoid double-runs).
- **Phase 3** — target-ability adapter for image / universe / series / story / writers-room.
- **Phase 4** — `tools/music.js` in the creative-tool registry so a commission can target music.
- **Phase 5** — shared extensible genre/category taxonomy + adaptive cadence/quality.
- **Deeper autonomy gate (Phase 2 candidate):** the fire handler's execute/budget pre-check duplicates the gate `dispatchCreativeTool` owns, because the planner LLM (`enqueuePlanTask`) is not itself routed through that chokepoint. The clean fix is a shared `assertCreativeExecuteAllowed()` at the planner-enqueue boundary so *every* CD-pipeline caller (not just this scheduler) honors "no cold LLM"; deferred as it touches the shared CD directive graph.

## Test plan
- `server/services/creativeCommissions/{directive,store,scheduler}.test.js` — cron composition, feedback digest (absent-vs-empty), store CRUD + sanitizer + cap, the autonomy/budget/ability gates, the **no-boot-generation** guarantee, and the `commission:changed` re-sync subscription.
- `server/routes/creativeCommissions.test.js` — route → service-arg contract, validation 400s, service-error → status mapping.
- `scripts/migrations/191-creative-commissions.test.js` — fresh stamp, idempotency, no-downgrade.
- `server/lib/navManifest.test.js` (new entry auto-covered) + client build/lint green.

Refs #2657


atomantic:
## Review status — 2 items need a human call before merge

The `claude` + `codex` review loop ran (3 rounds). All **correctness** findings are fixed and covered by tests: the broken `/creative-commission/new` route, the `updateCommission` scheduler-vs-request race, the PATCH default-wipe of `brief`/`generation` (defaults-free update schemas + constraints deep-merge), invalid-IANA-timezone rejection, resilient/​serialized scheduler re-sync, the derived-project-name 80-char cap, a double-charged planner budget action, and fail-closed governance reads. 103 unit/route/migration tests pass; client lint + build green.

Two **architectural** items remain that I did not merge over — both warrant your decision:

### 1. Storage backend — this should be `db-primary`, and I shipped it file-backed
The issue's own design says: *"New persisted entity (DB-primary per `docs/STORAGE.md`; add table + file fallback + migration + `data.reference/` seed)."* I implemented a machine-local `createCollectionStore` file collection (`data/creative-commissions/`) instead. `codex` independently flagged the same thing against the storage contract (a new Create-suite relational record with status + run lineage + universe/series refs is `db-primary`; machine-locality does **not** imply file storage — `tribe_*` is machine-local **and** Postgres).

**I stopped rather than ship the conversion unvalidated.** The concrete path (well-established in-repo): add idempotent DDL to `ensureSchema()` in `server/lib/db.js` for a machine-local `creative_commissions` (+ optional `creative_commission_runs`) table (no `sync_sequence`/tombstone — never federated, mirroring `tribe`), write a `query()`/`withTransaction` adapter, and select pg-vs-file per `NODE_ENV`/`MEMORY_BACKEND` exactly like `server/services/creativeDirector/local.js` (so the existing unit tests keep exercising the file path). Then drop migration `191` (the file-index stamp) since the table replaces it. I didn't do this autonomously because the Postgres SQL path can't be validated in this environment and a SQL typo would ship broken to real installs.

### 2. Planner concurrency/budget admission (shared CD-directive infra)
`codex` notes the scheduled planner reaches `enqueuePlanTask` → `task:ready` directly, bypassing `dequeueNextTask`'s `maxConcurrentAgents` admission, so many due planners could launch past the concurrency cap. This is **not new to this PR** — it's the exact path the existing user-facing `POST /api/creative-director/:id/directive` route uses. Routing scheduled planners through the CoS queue's gated admission is a change to shared CD-directive infrastructure that affects all directive callers, so it belongs in its own PR (tracked on #2657's Remaining).

The branch is functional and the correctness bar is met; I've left it open with the claim markers in place so you can decide the storage backend (and whether #2 blocks) rather than have me override the `docs/STORAGE.md` PR-review gate autonomously.

atomantic:
## Storage backend resolved → `db-primary` (Postgres)

Item **#1** from the review-status comment is done. `CreativeCommission` is now a machine-local `creative_commissions` Postgres table (one row per commission, full record in `data` JSONB; `name`/`enabled`/timestamps mirrored into columns for the scheduler's enabled-commission query). Like `tribe`, it carries **no** `sync_sequence` and **no** tombstone — never federated, hard deletes.

**How it's wired (mirrors universeBuilder/storyBuilder):**
- New `server/services/creativeCommissions/db.js` — pure leaf I/O (`readRaw`/`listRaw`/`writeRaw`/`deleteRaw`).
- `store.js` selects pg-vs-file through the shared `pgFileFacade` — Postgres for real installs; the `createCollectionStore` file backend is retained **only** as the `NODE_ENV=test` / `MEMORY_BACKEND=file` escape hatch, so the unit suite stays DB-free.
- The facade owns the sanitizer/merge uniformly and serializes the scheduler-vs-request read-modify-write on a shared per-id write queue (`createRecordWriteQueue`) — chosen over a DB row lock because commissions are machine-local and written only by the single main-server process, keeping both backends serializing identically.
- DDL added to `ensureSchema()`; **migration 191 dropped** (the table replaces the file type-index stamp, and commissions never shipped on the file backend, so there's nothing to migrate).

**Validation:** full server suite green (19548 passing); `npm run test:db` green against `portos_test` (new `db.test.js` round-trip: verbatim readback, upsert w/ `created_at` preserved, list order, hard delete); and a live end-to-end smoke through the real PG backend confirmed create/get/update-deep-merge/recordRun/list/delete plus concurrent update+run appends on one id all serialize correctly. `docs/STORAGE.md` + changelog updated.

Item **#2** (routing the scheduled planner through `dequeueNextTask`'s `maxConcurrentAgents` admission) is unchanged — it's shared CD-directive infra used by the existing directive route too, so it still belongs in its own PR (tracked on #2657's Remaining). This branch no longer has an open storage question.

PR #3056, opened 2026-07-25
feat: reprocess turnaround candidates with corrections

atomantic:
## Summary
- add a per-turnaround-candidate correction and re-process action
- use the selected candidate as the image-generation seed and preserve provenance
- validate candidate ownership/type and cover the request in client/server tests

## Test plan
- `git diff --check`
- `node --check server/services/sprites/reference.js`
- `node --check server/services/sprites/prompts.js`
- `node --check server/lib/validation.js`
- Vitest targeted suites could not run because dependencies are not installed in this worktree (`vitest: command not found`).

### DIFF
--- a/client/src/components/sprites/CorrectionNote.jsx
+++ b/client/src/components/sprites/CorrectionNote.jsx
@@ -24,7 +24,7 @@ export function correctionPromptPayload(corrections, direction) {
   return note ? { correctionPrompt: note } : {};
 }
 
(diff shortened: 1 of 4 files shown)


atomantic:
CI \`Client tests and build\` failure is the known WordplayTrainer flake (unrelated to this diff — see main run 30162288545 which failed the identical test on an unrelated commit). Passes locally: 5/5. Merging per the documented main-has-no-required-checks policy.

PR #3067, opened 2026-07-25
feat: ship scanner animation track

### LINKED ISSUE #3018: Build the scanner action as the first real non-walk animation track, end to end
Part of #2988.

Depends on #3015, #3016

## Problem / Goal

The track machinery is unproven until a real non-walk animation goes through it end to end. This is the proving case: a **short character action** — the scanner, done properly this time — taking the full path of generation → approval → packing → publish, at a frame count that is *not* the walk's.

The name is deliberate. A `scanner` column already existed in the atlas as a verbatim copy of the idle cell that nothing ever drew; #2986 removed it. It was a legacy render baked into the walk grid rather than treated as its own animation. Building it as a real track is the concrete demonstration that the generalization works.

## Proposed approach

1. **Register `scanner`** in the track registry with its own bounds (a short action, on the order of 4 frames — well below walk's floor of 6, which is exactly why per-track bounds were needed).
2. **Decide and implement its generation path.** Do NOT assume every track routes through the walk video pipeline. A short action may be better served by a shorter clip or by direct image generation. Whatever is chosen, it must obey the AI-provider policy: **user-triggered only**, no cold-bootstrap calls.
3. **Approval + packing** reuse the per-track flows from the registry and packing children rather than adding a parallel walk-shaped path.
4. **Publish** through the existing sidecar; the consuming app resolves the new track by name.

## Acceptance criteria

- [ ] `scanner` is a registered track with its own frame-count bounds, independent of walk's.
- [ ] Its frames can be generated, reviewed, and approved through a per-track flow — no walk-specific assumptions.
- [ ] It compiles into the atlas as its own column span alongside a 12-frame walk, and appears correctly in the `layout.json` `tracks` descriptor.
- [ ] Publishing carries it across, and a consumer can resolve and play it by name.
- [ ] No AI provider call happens without a direct user action.
- [ ] Walk behavior is entirely unchanged.

## Out of scope

- Wiring the game to actually *play* the scanner action in response to gameplay (that belongs in the consuming app's repo).
- Any second non-walk track beyond this one.

atomantic:
## Summary
- registers a short directional scanner action with independent bounds, direct user-triggered render/review/approval flow, and stale-anchor invalidation
- compiles verified scanner frames into a named atlas span alongside unchanged walk data, then carries that span through the layout sidecar and runtime contract
- adds Sprite Manager controls and focused coverage for scanner generation, publishing, and legacy walk isolation

## Test plan
- `cd server && npm test -- --run services/sprites/animationTracks.test.js services/sprites/animationTargets.test.js services/sprites/atlasGrid.test.js services/sprites/walkFrames.test.js services/sprites/atlas.test.js services/sprites/scanner.test.js services/sprites/walk.test.js services/sprites/publish.test.js services/sprites/atlasLayout.test.js lib/validation.test.js routes/sprites.test.js --silent=passed-only`
- `cd client && npm test -- --run src/components/sprites/ScannerWorkflow.test.jsx src/components/sprites/PublishWorkflow.test.jsx`
- `cd client && npm run lint -- --quiet`
- `cd client && npm run build --silent`

Closes #3018


atomantic:
## Review gate handoff

The required Claude CLI review was invoked three times but completed without the required `FINDING` or `NO FINDINGS` verdict. The optional Antigravity pass also did not reach this PR worktree. Manual self-review and the focused server/client validation in the PR test plan passed.

Leaving this PR open for an independent required review before merge.

Issue #3014, opened 2026-07-25
Add LI proposal outcome metrics to LI metrics data


atomantic:
## Problem

LI proposals have a 90% merge rate but zero completion tracking. LI metrics track reasoning quality (merge rates, per-scope success rates) but have no signal for whether LI's approved proposals actually get completed. This means LI cannot:
- Measure its own downstream effectiveness (approval → completion)
- Determine whether to self-exclude (its health metric has no completion-rate component)
- Calibrate proposal confidence based on historical completion rates
- Distinguish between LI's proposals getting rejected vs. LI's proposals getting lost

## What to add to LI metrics data

Add these fields to the LI metrics object in `data/metrics/layeredIntelligence.json` (or wherever LI metrics are stored):

```json
{
  // ... existing LI metrics fields ...
  "li-outcomes": {
    "totalApproved": 0,
    "totalCompleted": 0,
    "totalRejected": 0,
    "totalAbandoned": 0,
    "approvalToCompletionRate": 0,
    "byScope": {}
  }
}
```

Where `byScope` maps each LI proposal scope to its outcome counts, mirroring the existing `liOutcomes` structure:
```json
{
  "loop-meta": { "approved": 0, "completed": 0, "rejected": 0, "abandoned": 0, "completionRate": 0 },
  "app-data-gap": { "approved": 0, "completed": 0, "rejected": 0, "abandoned": 0, "completionRate": 0 },
  "app-improvement": { "approved": 0, "completed": 0, "rejected": 0, "abandoned": 0, "completionRate": 0 }
}
```

## Where to capture completions

- LI outcome data is already gathered during LI's reasoning runs via the LI execution outcomes block. The gap is that this data is **not persisted** to the LI metrics file.
- When LI's approved proposal gets executed (either completed or rejected/abandoned), write the outcome to the LI metrics file in `data/metrics/`.
- LI should update `li-outcomes` metrics in `server/services/layeredIntelligence.js` during the LI reasoning run's outcome aggregation phase (where `liOutcomes` data is currently gathered).

## Why this matters

Without completion-rate telemetry, the LI system operates with a fundamental blind spot. LI currently proposes work at 90% merge rate but has no idea if any of its proposals actually lead to completed work. This is the single highest-leverage data gap because:
1. It enables LI to self-calibrate (is LI's output actually valuable downstream?)
2. It unblocks the LI self-exclusion mechanism (health calculation needs completion data)
[...]

PR #3121, opened 2026-07-27
Surface approved-proposal delivery alongside run success in LI cosMetrics

### LINKED ISSUE #3085: Add agent task delivery tracking metrics to cosMetrics data model
## Problem
LI currently has zero visibility into whether agents actually complete and deliver the work of approved proposals. The existing `cosMetrics` tracks `internal-task`, `user-task`, and other scoped metrics with fields like `lifetimeSuccessRate`, `lifetimeCompleted`, etc. — but these measure *task execution success* (did the agent run the task?), not *delivery completion* (was the result merged/committed to the repo?).

With the LI loop reporting 0% approval-to-completion delivery health (10 approved, 0 completed), there is a critical blind spot: high user-task success rates could mask a pipeline where agents execute work successfully but never deliver it.

## What to Add

Add these new fields to the `cosMetrics` JSON structure in the LI data model:

```json{
  "delivery": {
    "totalApproved": 0,
    "totalDelivered": 0,
    "currentDeliveryRate": 0
  }
}
```

Where:
- `totalApproved` = count of LI proposals that have been approved by the user (currently tracked in liOutcomes as a post-run metric, but not persisted to cosMetrics)
- `totalDelivered` = count where approval was followed by successful execution AND the result was merged/committed to the repo
- `currentDeliveryRate` = totalDelivered / totalApproved (null if totalApproved is 0)

Also add a scoped breakdown:
```json{
  "deliveryByScope": {
    "app-data-gap": {"approved": 3, "delivered": 0},
    "app-improvement": {"approved": 2, "delivered": 0},
    "loop-meta": {"approved": 5, "delivered": 0}
  }
}
```

## Implementation Guidance
- Location: Add these fields alongside existing cosMetrics (the LI data model — likely in `server/services/layeredIntelligence.js` or the LI metrics module).
- Update mechanism: Each time a proposal transitions from "approved" to either "delivered" (merged) or "not delivered" (rejected/abandoned), increment the appropriate counter.
- The delivery rate feeds directly back into the `liHardExclusions` health metric — when it drops below 50%, it contributes to the exclusion gate.
- Keep the model backward-compatible: existing fields unchanged; new top-level key.

## Why This Matters
Without this metric, LI's calibration is incomplete. It knows a task *succeeded* (ran correctly) but not whether the work was *delivered* (produced merged output). This distinction is critical for:
1. Accurate LI health reporting in the exclusion gate
2. Understanding agent pipeline reliability vs. task execution quality
3. Calibrating whether to propose work that requires agent delivery or stick to self-contained changes

atomantic:
## Summary

The `cosMetrics` block the Layered Intelligence reasoner reads reported per-task-type agent **run** success (`lifetimeSuccessRate` — "did the task finish?") and nothing about **delivery** ("did the approved work actually land?"). With the loop reporting 0-of-10 approval-to-completion, a 90% run rate sat in that document looking like a healthy pipeline.

The document now carries the second number, in the same JSON:

```json
{
  "delivery": { "totalApproved": 3, "totalDelivered": 1, "currentDeliveryRate": 33 },
  "deliveryByScope": { "app-improvement": { "approved": 2, "delivered": 1 } },
  "user-task": { "lifetimeSuccessRate": 92, "...": "unchanged" }
}
```

Design decisions worth calling out, since two depart from the issue's suggested implementation:

- **Derived, not stored.** The issue proposed incrementing on-disk counters at each approved → delivered transition. `data/cos/li-outcomes/` already persists both sides of the lifecycle, so `computeDeliveryMetrics` projects the existing `computeProposalOutcomeMetrics` roll-up (#3014) instead. No second aggregate free to drift from its source, and no migration to seed on every existing install.
- **Composed in the hook, not `gatherSources`.** `gatherSources` runs *before* outcomes are reconciled against the run's tracker read, so a delivery figure emitted there would be one cycle stale — and `cosMetrics` and `liOutcomes` in the *same prompt* would disagree about how many proposals are approved. `gatherSources` hands its raw per-type map back on `cosMetricsByType`; the hook re-renders the document once it has post-reconciliation records, then drops the hand-off key. This keeps `gatherSources` overlapping the tracker read as before, rather than serializing behind it.
- **Sentinel handling.** `currentDeliveryRate` is `null` — never `0` — until something has been approved: that rate arms the hard exclusion gate (`LI_HARD_GATE_DELIVERY_THRESHOLD`), so collapsing "nothing to deliver on" into "everything was lost" would lock out a cold loop. The block is *omitted entirely* when no outcome records were gathered at all (outcomes source off / tracker can't report outcomes / store unreadable), keeping "no delivery data" distinct from "nothing approved yet". `deliveryByScope` lists only scopes with at least one approval — a rejected-only scope has no delivery signal, and the merge-rate block already covers its filing side.
- Delivery keys are serialized **first** so the 4000-char prompt-budget cap can't truncate them away (the per-task-type tail already has an interpreted sibling in the prompt via `liScopeAwareness`, derived server-side from the untruncated map; the delivery block has none).

The issue's last requirement — that the delivery rate feed the `liHardExclusions` health metric below 50% — already shipped with #3014/#2824 (`computeLiExecutionHealth` → `LI_HARD_GATE_DELIVERY_THRESHOLD`); this PR adds the missing reasoner-facing surface, so the gate and the prompt now read the same signal from the same derivation.

## Test plan

- `cd server && npx vitest run services/layeredIntelligence.test.js services/layeredIntelligenceBarrel.test.js services/autonomousJobs/layeredIntelligenceHooks.test.js` — 429 pass.
- New unit coverage: `computeDeliveryMetrics` totals/rate, per-scope projection, the null-rate sentinel across empty/null/rejected-only histories, pre-computed-roll-up equivalence, and `__proto__`-as-data.
- New `renderCosMetricsSource` coverage: unchanged document with no delivery data, folded-in delivery, delivery surviving a genuine truncation at the cap, and a task type named `delivery` failing to shadow the block.
- New hook wiring coverage: the *reconciled* outcomes are what get projected, the re-rendered document (not the first pass) reaches `buildPrompt`, `cosMetricsByType` never leaks through as a bogus source block, `delivery: null` when outcomes weren't gathered, and no re-render at all when the `cosMetrics` source is off.
- `cd server && npm test` — 53 DB-backed suites fail identically on clean `main` in this environment (local Postgres schema not provisioned); no LI-related failures.

Closes #3085

### DIFF
--- a/.changelog/NEXT.md
+++ b/.changelog/NEXT.md
@@ -54,3 +54,7 @@
 ## Video generation
 
 - **[issue-3100] New Control mode for local video generation** — feed a control clip (a depth pass, a pose track, edges) alongside your prompt and the render follows its structure and motion instead of inventing them. The reference can be a fresh upload or any prior render from your history, and a strength slider dials how tightly the output tracks it. A half-res preview toggle skips the refinement pass so you can check whether a control clip fits before committing to a full render. Requires an LTX-2 model; the ~654 MB Control weight downloads with progress from the mode panel and is covered by the existing model verify/repair buttons.
+
+## Layered Intelligence
+
+- **[issue-3085] The self-improvement loop can now see whether its approved proposals actually get delivered, not just whether its agent runs finish.** The `cosMetrics` block it reasons over reported per-task-type run success — "did the agent's task complete?" — and nothing about delivery, so a 90% run rate sat there looking healthy beside a pipeline that had landed none of ten approved proposals. That same block now carries a `delivery` figure (approved, delivered, and the current delivery rate) plus a per-scope breakdown, so the two numbers are read side by side. The rate is reported as unknown rather than 0% until something has actually been approved — a loop with no track record shouldn't read as a loop that fails everything.
--- a/server/services/autonomousJobs/layeredIntelligenceHooks.js
+++ b/server/services/autonomousJobs/layeredIntelligenceHooks.js
@@ -367,8 +367,10 @@ export async function buildTaskInput({ app } = {}) {
   // never be read as a healthy pipeline. Done here rather than inside gatherSources
   // because only this point has POST-reconciliation outcomes — a pre-reconcile snapshot
   // would report a different approval count than the liOutcomes block in the same
-  // prompt. `cosMetricsByType` is gatherSources' internal hand-off for exactly this
-  // re-render; drop it so it can never leak into buildPrompt as a source block.
+  // prompt. Present only when the `cosMetrics` source is enabled, so an app that turned
+  // that source off gets no block (its delivery signal rides liOutcomes / liSelfEval).
+  // `cosMetricsByType` is gatherSources' internal hand-off for exactly this re-render;
+  // drop it once consumed rather than leaving a spent key on the source map.
   if (sources.cosMetricsByType) {
     const rendered = renderCosMetricsSource({
       metricsByType: sources.cosMetricsByType,
--- a/server/services/autonomousJobs/layeredIntelligenceHooks.js
+++ b/server/services/autonomousJobs/layeredIntelligenceHooks.js
@@ -53,6 +53,8 @@ import {
   resolveJiraBlockKey,
   applyJiraBlockingLabel,
   computeOutcomesReport,
+  computeDeliveryMetrics,
+  renderCosMetricsSource,
   computeSelfEvalSummary,
   computeProposalExecutionAwareness,
   computeCrossReferenceAnalysis,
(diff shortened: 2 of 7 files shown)


atomantic:
Review rounds complete — `codex` then `claude`, per the configured reviewer order.

**codex** (2 findings, both fixed in b2f410c):
- The `cosMetrics` source was skipped outright when `learning.json` had no `byTaskType`, which withheld the delivery half too — and that half comes from the `li-outcomes` records, not `learning.json`. An install with no CoS run history but with approved-but-undelivered proposals got no delivery signal at all. `gatherSources` now hands back an empty per-type map, and `renderCosMetricsSource` returns `''` only when *both* halves are empty, so the source is still omitted rather than rendering a hollow `{}`.
- Missing coverage for the unreadable-outcome-store path (`{ read: false }`) — behavior was already correct, now asserted.

**claude** (5 findings, all doc/test-quality rather than behavior; fixed in e125c24):
- `computeDeliveryMetrics`' `{ metrics }` reuse seam had no production caller. Removed as speculative surface.
- Its doc claimed the display rate arms the hard exclusion gate. It doesn't — the gate recomputes from the records via `computeLiExecutionHealth`, and with zero approvals `LI_DELIVERY_MIN_SAMPLE` already disarms it. The null sentinel stands on prompt-honesty grounds; the doc now says that instead of asserting a coupling a future editor would act on.
- "Both halves independently optional" overstated the scope — that holds *within* the `cosMetrics` source, which still gates the whole document. Corrected in all three places.
- The `__proto__` guard asserted `{}.approved`, which passes on a plain-object implementation too. Now asserts the null prototype directly.
- Replaced a hook test that mocked away everything it claimed to check with a real `buildPrompt` assertion that a blank `cosMetrics` document renders no block.

Issue #3089, opened 2026-07-26
Replace the generic CoS quick-task templates with slashdo command templates that carry implied settings and provider-correct invocation


atomantic:
## Problem / Goal
The eight built-in "Quick Templates" in the CoS agent task form are generic phrase
stubs (`server/services/taskTemplates.js:15-88`) — "Fix the bug where", "Refactor",
"Add tests for", "Improve the user experience of". They seed a sentence fragment the
user has to finish typing anyway, so they save nothing and mostly sit collapsed.

Meanwhile PortOS bundles 19 real, high-leverage workflows as slashdo commands
(`lib/slashdo/commands/do/*.md`) that a CoS agent can execute end to end. Those are
what a quick-launch chip row should offer: one click → "run `/do:plan-task` on this
idea", with the run's settings already correct for that workflow.

## Context

**Current template system**
- `server/services/taskTemplates.js` — `BUILT_IN_TEMPLATES` (8 generic stubs) merged
  with user templates from `data/cos/task-templates.json` (`{ version, userTemplates,
  usage }`). Each template carries `{ id, name, icon, description, context, category,
  isBuiltin }`; user templates add `provider/model/effort/app`.
- `server/routes/cosTemplateRoutes.js` — CRUD + `POST /templates/:id/use`. **These
  routes are not Zod-validated** (hand-rolled `if (!name || !description)`), which
  breaks the repo's "all route inputs validated via `lib/validation.js`" convention;
  extending the payload shape is the moment to fix it.
- `client/src/components/cos/TaskAddForm.jsx:133-143` — `applyTemplate` sets only
  `description/model/provider/effort/app`. The stored `context` field is **never read
  by any consumer**, and none of the run-shape toggles (`useWorktree`, `openPR`,
  `simplify`, `prCompletion`) are touched, so a template can't express "this workflow
  needs no worktree."
- Chips render at `TaskAddForm.jsx:358-400`, fed by `getCosPopularTemplates(8)`.

**Per-provider invocation is not a prefix swap.** slashdo installs itself differently
per host CLI (`lib/slashdo/src/environments.js`, `getTargetFilename`/`getSkillName` in
`lib/slashdo/src/transformer.js:179-206`):

| Env | namespacing | install path | invocation |
|---|---|---|---|
| Claude Code | `subdirectory` | `~/.claude/commands/do/plan-task.md` | `/do:plan-task <args>` |
[...]

Issue #3114, opened 2026-07-26
Converge the two slashdo-capability gates and the two launchable-workflow catalogs


atomantic:
## Problem / Goal

There are now **two** answers to "how does this host run a bundled slashdo
workflow", and **two** catalogs of which workflows are launchable. Both pairs
have already drifted.

## Context

**Two provider→slashdo-shape decisions.** #3089 added
`resolveSlashdoStyle()` (`server/lib/slashdoInvocation.js`), which classifies a
provider into `slash-namespaced` (`/do:x`), `slash-flat` (`/do-x`), or `skill`.
`server/services/agentPromptBuilder.js` still carries two older gates that answer
a closely-related question — whether the completion workflow may *type*
`/do:pr` / `/do:push` / `/simplify`:

- `hasSlashdo` — `!isTui && (providerId === 'claude-code' || 'claude-code-bedrock')`
  (id allowlist; a path-configured or renamed `claude` binary is not recognised,
  unlike every other provider predicate in `providerModels.js`)
- `tuiSlashdoFree` — `isTui && (isOpencodeCommand(providerCommand) || leanMode)`
  (so a **codex TUI** is treated as slashdo-capable and told to run `/do:pr`,
  which it cannot — codex gets Agent Skills, not slash commands)
- a third copy in the full path at the `buildCompletionGuidelineBullet` call site
  omits `leanMode`, so a lean Claude TUI gets `/do:pr` in the guideline bullet
  while its completion section says to use plain git.

These were deliberately NOT unified in #3089: they depend on the opposite
blank-command posture (`inferTuiCommand` resolves a blank command to `claude`, so
an unidentified TUI must count as slashdo-capable there, while
`resolveSlashdoStyle` deliberately refuses to read blank as Claude). Unifying
naively regresses 8 existing light-path TUI tests. The convergence is real work
that needs the blank-command posture made explicit, not a rename.

**Two launchable-workflow catalogs.** `POST /api/cos/tasks/slashdo`
(`server/routes/cosTaskRoutes.js`) has its own `SLASHDO_COMMANDS` registry, fed by
`client/src/components/apps/SlashDoPanel.jsx`. #3089 added a second set as the
CoS quick templates (`BUILT_IN_TEMPLATES` in `server/services/taskTemplates.js`).
They already disagree: `push` / `better-swift` exist only in the route's registry;
`plan-task` / `depfree` / `scan` / `replan` only in the templates. The route also
[...]

PR #3129, opened 2026-07-27
unify the three divergent slashdo command catalogs and stop the route inlining Claude-only invocations

### LINKED ISSUE #3108: Unify the three divergent slashdo command catalogs and route POST /tasks/slashdo through resolveSlashdoInvocation
## Problem

PortOS now has **three independent hardcoded lists** of "which bundled slashdo
commands can a CoS agent run, and with what run-shape":

| Site | Contents |
|---|---|
| `server/routes/cosTaskRoutes.js:37` `SLASHDO_COMMANDS` | allowlist + descriptions for `POST /api/cos/tasks/slashdo` |
| `client/src/components/apps/SlashDoPanel.jsx:8` `SLASHDO_COMMANDS` | labels + descriptions + Swift gating for the app-overview buttons |
| `server/services/taskTemplates.js` `BUILT_IN_TEMPLATES` | the quick-task templates added in #3089 |

They have already diverged:

- **Membership**: `push` and `better-swift` exist only in the first two;
  `plan-task`, `depfree`, `scan` only in the third.
- **Descriptions**: `replan` is "Audit PLAN.md, archive completed items, prune
  stale work" in the route vs "Audit the backlog — prune shipped items, surface
  new work" in the templates. Same for `next`/`review`/`release`/`better`.
- **Run shape**: the route hardcodes `reviewLoop: false` alongside
  `useWorktree/openPR/simplify: false`; `SLASHDO_RUN_SETTINGS` omits
  `reviewLoop`. So the same `/do:next` behaves differently depending on which
  button queued it.

Separately, the **route path still does the exact thing #3089 was built to
stop**: `cosTaskRoutes.js:159` writes a Claude-only literal into the
description (`Run /do:${command} for ${app} — …`) and unconditionally inlines
the raw command body into `task.context` for *every* provider — including
Claude Code, where `resolveSlashdoInvocation` deliberately returns
`inlinesBody: false`. The two paths now disagree about what is correct for the
same provider, and the SlashDoPanel buttons are the higher-traffic path.

## Decisions already made (this is ready to work)

1. **One catalog: `server/lib/slashdoCatalog.js`**, barrel + README row per the
   Module Organization rule. One entry per command:
   `{ command, name, icon, description, argsPlaceholder?, settings, swiftOnly?, hideForSwift?, templateEligible }`.
   `settings` carries the full run shape including `reviewLoop: false` (the
   route's value wins — it is the more complete one).
2. **`templateEligible`** marks the eight quick-template commands
   (`plan-task`, `next`, `replan`, `review`, `release`, `better`, `depfree`,
   `scan`). `taskTemplates.js` derives `BUILT_IN_TEMPLATES` by filtering on it,
   keeping the `slashdo-<command>` ids so existing `usage` counters survive.
3. **The route's allowlist** becomes "any catalog entry", so `plan-task` /
   `depfree` / `scan` become dispatchable from the panel too. `push` and
   `better-swift` join the catalog with `templateEligible: false`.
4. **The client reads the catalog from the server** via a small
   `GET /api/cos/slashdo-commands`, rather than mirroring a fourth copy.
   `SlashDoPanel`'s `classes` (per-command Tailwind colors) stay client-side —
   they are presentation, not catalog data.
5. **Converge the route's non-`next` branch** onto the new mechanism: drop the
   `loadSlashdoCommand(command)` inline and the `/do:` literal; pass
   `slashdoCommand: command` through `addTask` and write a provider-neutral
   description (`Run the do-<cmd> workflow for <app> — <catalog description>`).
   The `next` branch stays as-is — `buildClaimWorkTask` generates a
   work-tracker-aware claim prompt, which is genuinely different work, not a
   slashdo body.
6. **Derive `hasSlashdo` / `tuiSlashdoFree` from `resolveSlashdoStyle`.**
   `agentPromptBuilder.js:1380` and `:1388` are a third, narrower answer to
   "can this host run `/do:*`" — `hasSlashdo` is id-only
   (`claude-code`/`claude-code-bedrock`), so a path-configured `claude`
   provider gets `/do:plan-task` from the new resolver while the completion
   section simultaneously tells it slashdo is unavailable. Map
   `SLASH_NAMESPACED ⇒ hasSlashdo`, `SKILL|INLINE ⇒ slashdoFree`. **This is a
   behavior change** for path-configured/renamed claude providers — it is the
   intended fix, and needs a test asserting the completion workflow and the
   slashdo section agree for such a provider.

## Acceptance criteria

- [ ] `server/lib/slashdoCatalog.js` exists, is the only place a bundled command's
      name/description/run-shape is declared, and is barrel- + README-registered.
- [ ] `cosTaskRoutes.js` and `taskTemplates.js` both read it; neither declares a
      command list of its own. `SlashDoPanel.jsx` fetches it.
- [ ] `POST /api/cos/tasks/slashdo` (non-`next` branch) persists
      `metadata.slashdoCommand` instead of inlining the body, and its description
      contains no `/do:` literal. `cos.test.js`'s `loadSlashdoCommand` assertions
      are updated accordingly.
- [ ] `hasSlashdo` and `tuiSlashdoFree` in `agentPromptBuilder.js` derive from
      `resolveSlashdoStyle`; a test covers a path-configured `claude` provider
      getting a consistent completion workflow + slashdo section.
- [ ] Existing `usage` counters for `slashdo-*` ids survive the refactor.
- [ ] `.changelog/NEXT.md` entry.

## Out of scope

- Changing which commands are offered (membership changes beyond folding in
  `push` / `better-swift` / the three new ones).
- The prompt-payload size problem — tracked separately.

Discovered while working #3089.

atomantic:
## Summary

PortOS had **three independent hardcoded lists** of "which bundled slashdo command can a CoS agent run, and with what run-shape" — the route allowlist in `cosTaskRoutes.js`, the app-overview buttons in `SlashDoPanel.jsx`, and the quick-task templates in `taskTemplates.js`. They had already diverged: `push`/`better-swift` existed only in the first two, `plan-task`/`depfree`/`scan` only in the third, shared commands carried different descriptions, and only the route sent `reviewLoop: false` — so the same `/do:next` was reviewed twice or not, depending on which button queued it.

`server/lib/slashdoCatalog.js` is now the single declaration. The route and the templates read it; the client fetches a narrowed projection over `GET /api/cos/slashdo-commands` instead of mirroring a fourth copy. Per-command button colors stay client-side as presentation.

The route's non-`next` branch also stopped doing the exact thing #3089 was built to prevent: it wrote a Claude-only `Run /do:${command}` literal into the description and inlined the raw command body for **every** provider, including Claude Code, where `resolveSlashdoInvocation` deliberately returns `inlinesBody: false`. It now pins `metadata.slashdoCommand` and lets the prompt builder render the shape the chosen provider can actually run. The `next` branch keeps `buildClaimWorkTask` — a work-tracker-aware claim prompt is genuinely different work, not a slashdo body — now selected by a catalog `claimsWork` flag rather than a `command === 'next'` literal in two places.

Finally, `hasSlashdo`/`tuiSlashdoFree` in `agentPromptBuilder.js` were a third, narrower answer to "can this host run `/do:*`". `hasSlashdo` was an id-only allowlist, so a path-configured `claude` provider got `/do:plan-task` from the resolver while the completion section simultaneously told it slashdo was unavailable — and a codex TUI was told to run `/do:pr`, which it cannot. Both now derive from `hostTypesSlashdoCommands`, which keys on the launch command and resolves a blank one through the same `inferTuiCommand` mapping the spawners use.

**Intended behavior changes** (both asserted by new tests):
- A path-configured / renamed `claude` provider now gets the slashdo completion workflow.
- A codex / antigravity / gemini / kimi TUI no longer gets `/do:pr` commands its CLI doesn't have.

Two more copies of that gate surfaced during the quality pass — the Simplify Step's allowlist and a `/claude/i` id regex in the TUI completion section — so a path-configured claude TUI was still told to hand-review *and* to run `/do:pr` in one numbered list. Both now derive from the one gate.

Closes #3108

## Test plan

- `server/lib/slashdoCatalog.test.js` (new) — the catalog is the single source of truth: every command from all three former lists is present exactly once, no entry hand-declares the derived `/do:x` label, every entry carries the full run shape including `reviewLoop: false`, `templateEligible` yields exactly the prior eight template commands (so `usage` counters keyed `builtin-do-<command>` survive), every command actually ships in `lib/slashdo/commands/do/`, and the client projection keeps the server-only fields off the wire.
- `server/lib/slashdoInvocation.test.js` — `hostTypesSlashdoCommands` across claude-by-id, claude-by-path, blank-command per id shape (`codex-tui`/`antigravity-tui`/`gemini-tui`/`kimi-tui` → false; `claude-ollama`/undefined → true), an explicit command that disagrees with the id, opencode, and lean mode; plus an explicit test that it differs from `resolveSlashdoStyle` *only* on the blank-command reading.
- `server/routes/cos.test.js` — every catalog command dispatches (including the three newly-launchable ones), the bare command name is persisted with no body inlined and no `/do:` literal in the description, an out-of-catalog command (`rpr`) is rejected, and the new catalog endpoint serves what the panel needs.
- `server/services/agentPromptBuilder.test.js` — the three intended gate outcomes above, asserted on the rendered prompt.
- `client/src/components/apps/SlashDoPanel.test.jsx` (new) — the panel renders from the fetched catalog, queues the bare command, opens the run drawer for a `configurable` command and threads `claimsWork` through, honors Swift gating both ways, and degrades to a header with no buttons when the fetch fails.
- Full client suite: 464 files / 5034 tests pass (one known GitTab Escape flake that passes in isolation). Client lint clean.
- Full server suite: the 58 failing files are Postgres-dependent suites plus 3 env-dependent Bedrock model-mapping tests — verified failing identically on unmodified `origin/main` (`4c8786f79`), so unrelated to this change. All 397 tests in the suites this change touches pass.


atomantic:
Superseded by #3114's implementation, which landed on `main` in eb61b7070 while this was in review — a parallel agent worked the near-duplicate issue and shipped a superset of this branch.

Verified against every #3108 acceptance criterion on `main`:

- `server/lib/slashdoCatalog.js` exists as the sole declaration, registered in both `server/lib/index.js` and `server/lib/README.md`.
- `cosTaskRoutes.js` and `taskTemplates.js` both read it (`BUILT_IN_TEMPLATES` now maps `SLASHDO_WORKFLOWS`); `SlashDoPanel.jsx` reads a client mirror at `client/src/lib/slashdoCatalog.js`.
- `POST /tasks/slashdo`'s non-`next` branch persists `slashdoCommand` with no `/do:` literal in the description, and the `next` branch keeps its claim prompt with no `slashdoCommand` attached.
- `hasSlashdo`/`tuiSlashdoFree` derive from `canTypeSlashCommands`, with the blank-command posture made explicit — plus two sites this branch missed: `agentCompletionCleanup`'s `agentOwnsPR` gate (which must agree with the prompt or PortOS opens a second PR) and `brainLinks.js`.
- Existing `slashdo-*` usage counters survive; changelog entry present.

Re-landing the same design here would mean resolving conflicts across 12 files to arrive at work that already exists, so this branch is closed rather than merged. Closing #3108 as a duplicate of #3114.

PR #3137, opened 2026-07-27
feat(creative-commissions): display renders + config side-by-side on desktop with smaller preview cards

atomantic:
## Summary
- Displays a Creative Commission's render history alongside its editable config side-by-side on desktop (stacked on mobile), instead of stacked full-width sections
- Shrinks the render grid to 2 columns max and adds a width-capped preview thumbnail so each card reads as a compact preview rather than a full-bleed hero image

## Test plan
- [x] `npm run lint --prefix client` passes
- [x] `npx vite build --prefix client` succeeds
- [ ] Manually verify in the browser: open a commission with renders in multiple aspect ratios (16:9, 9:16, 1:1) and confirm each card's play button and label are visible and clickable

### DIFF
--- a/client/src/components/creative-commission/RenderHistory.jsx
+++ b/client/src/components/creative-commission/RenderHistory.jsx
@@ -62,7 +62,7 @@ export default function RenderHistory({ runs, feedback, projectsById, projectsLo
   }
 
   return (
-    <div className="grid grid-cols-1 sm:grid-cols-2 xl:grid-cols-3 gap-4">
+    <div className="grid grid-cols-1 sm:grid-cols-2 gap-3">
       {orderedRuns.map((r) => {
         const project = r.projectId ? projectsById.get(r.projectId) : null;
         const focused = r.id === focusRunId;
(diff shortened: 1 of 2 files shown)


atomantic (commented):
Reviewed by /do:review — 1 critical bug found and fixed, 1 lint failure fixed, 1 minor DRY cleanup applied. Verified independently by a second review pass.

### Fixed
- **`client/src/components/creative-commission/RenderHistory.jsx`** — the original `overflow-hidden` + `maxHeight: '160px'` wrapper around `ProjectPreview` only clipped the *visible* area; `ProjectPreview`'s box still derived its actual height from the unclamped grid-column width via its `aspect-video`/`aspect-[9/16]`/`aspect-square` class. For any render narrower-than-16:9 (9:16 portrait, 1:1 square — common for social-style commissions), the box's real height exceeded 160px, which pushed the absolutely-centered play button (`top-1/2 left-1/2`) and the bottom label badge below the visible clipped region — invisible and unclickable. This is the exact anti-pattern `client/src/lib/creativeDirectorPreview.js` already documents and rejects (`previewMaxWidthClass`'s comment: "capping height directly does not work… constraining width instead lets aspect-ratio derive a height that already fits"). Replaced the height clamp with a `max-w-[220px] mx-auto` width cap, matching that documented pattern.
- Same file — `previewAspectClass` was imported but no longer called after the crop was introduced, failing CI lint (`no-unused-vars`, confirmed via `gh pr checks` before the fix). Restored its use in the placeholder branch, which also makes both branches size consistently instead of two different arbitrary pixel heights (160px vs 120px).
- Minor DRY follow-up: hoisted the now-identical `max-w-[220px] mx-auto` wrapper out of both branches of the ternary instead of duplicating it.

### Verified
- `npm run lint` (client) — clean
- `vite build` (client) — succeeds
- Independent second-pass review (fresh sub-agent, full-file read of `RenderHistory.jsx`, `CreativeCommissionDetail.jsx`, `ProjectPreview.jsx`, `creativeDirectorPreview.js`) confirmed the fix and found no further critical or cross-file issues.

### Out-of-diff observation (not fixed — pre-existing design trade-off, not a bug)
The `Configuration` sidebar is now capped at `w-[380px]` (was `max-w-2xl` ≈672px). `CommissionConfigForm`'s field-pair grids use a *viewport* breakpoint (`sm:grid-cols-2`), not a container query, so on ≥640px viewports those pairs still force 2 columns — now visibly tighter (≈170px per field vs ≈330px before). Nothing overflows or breaks (`inputCls` is `w-full` with no min-width), so this is a cosmetic tightening that follows directly from this PR's own width choice, not a defect.

### Manual verification still recommended
Open a commission with renders in 16:9, 9:16, and 1:1 aspect ratios and confirm each card's play button and label are visible/clickable in the browser — the reasoning above is verified against the CSS mechanics but not against a rendered browser.

_Generated by /do:review_

Issue #2191, opened 2026-07-04
TUI agents with zero activity are reaped as 'idle-complete' SUCCESS — task completed, no summary, worktree (and uncommitted work) deleted


atomantic:
Found during a live CoS TUI eval (2026-07-04). Generalizes the tuiHandshake false-success class (#2074 fixed it for merge-queue waits).

Observed twice in one afternoon:
1. `opencode` TUI showed "Model … is not valid", never processed the prompt → 185s idle → `completionReason: "idle-complete", success: true`, task marked **completed**, no summary.
2. `claude` TUI (local qwen3.6:35b) rambled, made one invalid tool call, ended at an interactive "How can I help you today?" prompt with zero file edits → same idle-complete success.

In both cases `cleanupAgentWorktree` then deleted the worktree, so any uncommitted work (or evidence of no work) is destroyed, and the UI shows a completed agent with nothing in it.

**Fix directions:**
- Gate idle-complete success on evidence of work: completion sentinel written, or non-empty `git status` in the worktree, or observed tool-call/output activity after prompt submission. Otherwise mark **failed** with the tail of the TUI buffer as the error analysis.
- Before deleting a worktree on idle-complete, capture `git diff` into the agent archive dir so post-mortems are possible.

PR #3139, opened 2026-07-27
stop failing idle-out TUI runs whose work product isn't files (#3102)

### LINKED ISSUE #3102: idle-no-changes fails reference-watch runs whose work tracker isn't the filesystem
## Problem

A TUI agent that idles out is failed when its worktree is clean:

`server/services/agentTuiSpawning.js` (idle-complete branch, ~L1080–1103)
```js
const hasChanges = await worktreeHasChanges(cwd);
if (!hasChanges) finish({ success: false, exitCode: 1, error: 'TUI agent idled out with zero uncommitted file changes …', reason: 'idle-no-changes' });
```

That gate assumes every agent's work product is a file change. It isn't. The
shipped **v3 `reference-watch` prompt** (`PROMPT_VERSIONS['reference-watch'] = 3`,
`server/services/taskPromptDefaults/versions.js`) routes proposals to the app's
**resolved work tracker** via the `{trackerInstructions}` block
(`server/services/referenceRepos.js#formatTrackerInstructions`). When that tracker
resolves to GitHub/GitLab/JIRA, the agent files issues and — by design, the prompt
says "you do NOT edit application code" — leaves the worktree **clean**.

So a reference-watch run that did its whole job is recorded as
`success: false, exitCode: 1, reason: 'idle-no-changes'` whenever it doesn't write
the `.agent-done` sentinel. Observed on a run that had already filed
`[ref-watch-…]` issues before the 180s idle reaper reaped it.

Only the non-file trackers are affected: a PLAN.md-tracker install still commits
checklist items, so its worktree is dirty and the gate holds.

## Decision (already made — implement as written)

Stop treating "clean worktree" as the only evidence of work; make the gate
conditional instead of adding a second evidence probe (querying GitHub for
recently-filed issues at finalize time is slow, network-dependent, and racy).

1. Add `worktreeChangesExpected` (boolean, default `true`) to agent task metadata.
   Add it to the metadata sanitizer + Zod schema in `server/lib/validation.js`
   (schema parity — see `server/CLAUDE.md`).
2. Set it to `false` when the spawning task's work tracker is non-file-based.
   `triggerReferenceAnalysis` in `server/services/referenceRepos.js` already
   resolves `workTracker.resolved` to pick the `{trackerInstructions}` block —
   set `worktreeChangesExpected: !isFileTracker(workTracker.resolved)` from the
   same resolved value, so the flag can never drift from the prompt the agent got.
3. In the idle-complete branch of `agentTuiSpawning.js`, skip the
   `worktreeHasChanges` gate when `worktreeChangesExpected === false` and fall
   back to the existing `workActivity.active` signal alone (which already
   distinguishes "prompt never submitted" → `idle-no-activity`). Still call
   `captureWorktreeDiff` unconditionally — it is a no-op on a clean tree and the
   diff is useful either way.

Default `true` keeps every existing task type on today's behavior; only a task
that opts out changes.

## Acceptance criteria

- A TUI run with `worktreeChangesExpected: false` that idled out with an active
  work counter and a clean worktree finalizes `success: true`, `reason: 'idle-complete'`.
- The same run with `worktreeChangesExpected` absent/`true` still fails with
  `reason: 'idle-no-changes'` (no behavior change for code-editing task types).
- A run with no work-counter activity still fails `idle-no-activity` regardless
  of the flag.
- `reference-watch` spawned against a GitHub/GitLab/JIRA tracker carries
  `worktreeChangesExpected: false`; against a PLAN.md tracker it carries `true`.
- Unit cover in `server/services/agentTuiSpawning*` tests and
  `server/services/referenceRepos.test.js`.

## Context

Found while investigating an auto-filed agent-failure task. The *classification*
of that failure (an idle-reaper kill mislabelled `context-length` off a keyword in
the agent's own transcript) is fixed separately in `agentErrorAnalysis.js`; this
issue is the underlying "was it actually a failure at all" question.

atomantic:
## Summary

A TUI agent reaped by the idle watchdog was failed whenever its worktree was clean — the `#2191` gate in the idle-complete branch of `server/services/agentTuiSpawning.js` treated "no uncommitted file changes" as proof the model produced nothing.

That assumes every agent's deliverable is a file change. The shipped v3 `reference-watch` prompt breaks it: against a GitHub/GitLab/JIRA work tracker the `{trackerInstructions}` block tells the agent to file issues and explicitly *not* to edit application code, so a run that filed its `[ref-watch-…]` issues and then hit the 180s idle reaper was recorded as `success: false, exitCode: 1, reason: 'idle-no-changes'`. Only the non-file trackers are affected — a PLAN.md-tracker install commits checklist items, so its tree is dirty and the gate holds.

Implements the decision on #3102 (make the gate conditional, rather than adding a second, network-dependent evidence probe at finalize time):

- **`worktreeChangesExpected` task metadata** (boolean, default = absent/`true`) — added to `ALLOWED_TASK_METADATA_KEYS` in `server/lib/cosValidation.js` and to the `createCosJobSchema.taskMetadata` Zod shape (schema parity — Zod strips undeclared keys, so an unlisted flag would silently vanish from a job's `taskMetadata`).
- **`isFileTracker(tracker)`** in `server/lib/workTracker.js` — true for PLAN.md (work lands as a committed file), false for `github`/`gitlab`/`jira`. An unknown/absent tracker is treated as file-based, matching `formatTrackerInstructions`' PLAN.md fallback.
- **`triggerReferenceAnalysis`** stamps `worktreeChangesExpected: isFileTracker(workTracker.resolved)` off the **same** resolved value it already uses to pick the `{trackerInstructions}` block, so the flag can't drift from the prompt the agent actually received.
- **The idle-complete gate** skips `worktreeHasChanges` when the flag is falsy, falling back to the existing `workActivity.active` signal alone — so a prompt that never submitted still fails `idle-no-activity` regardless of the flag. Read through `isFalsyMeta` so the TASKS.md string round-trip (`'false'`) works. `captureWorktreeDiff` still runs unconditionally (no-op on a clean tree, useful either way).

Default `true` means zero behavior change for every code-editing task type.

## Test plan

- `cd server && npx vitest run services/agentTuiSpawning.test.js services/referenceRepos.test.js lib/workTracker.test.js lib/cosValidation.test.js lib/validation.test.js lib/index.test.js` → **6 files / 328 tests passed**. New cover:
  - `agentTuiSpawning.test.js` — `worktreeChangesExpected: false` on a clean tree finalizes `success: true` / `idle-complete` and never calls `git.getStatus` (while both `getDiff` calls still fire); the `'false'` string round-trip behaves identically; `true` still fails `idle-no-changes`; and `false` does **not** rescue a run with zero work-counter activity (`idle-no-activity`).
  - `referenceRepos.test.js` — `plan` → `true`, `github`/`gitlab`/`jira` → `false`, asserted alongside the existing `metadata.workTracker` + injected-instructions assertions.
  - `workTracker.test.js` — `isFileTracker` per tracker plus the unknown/absent fallback.
  - `validation.test.js` / `cosValidation.test.js` — sanitizer keeps an explicit `false` and drops non-booleans; the job schema preserves the flag and rejects a non-boolean.
- `cd server && npm test` → 1030 files / 21135 tests passed; 58 files failing. All 58 are pre-existing environmental failures reproduced on `main` at the same SHA: PostgreSQL-required suites (`Pipeline series require PostgreSQL …` from `lib/pgFileFacade.js`) and Bedrock model-id assertions that read the live environment. None of the modules touched here are in that list.

Closes #3102


atomantic:
Follow-up filed as #3140: the **scheduled** `reference-watch` path (`buildImprovementTaskPrompt` in `cosTaskGenerator.js`) never substitutes `{trackerInstructions}` at all, so it ships the literal placeholder to the agent and needs the same `worktreeChangesExpected` stamp. Out of scope for this PR, which fixes the on-commit trigger path (`triggerReferenceAnalysis`).

PR #3206, opened 2026-07-29
Harden game asset integrity and blocker UX

atomantic:
## Summary

- verify every bound game sprite and music source by SHA-256 before compiling or launching a bundle
- add deterministic manifest staleness checks, imported object/props support, malformed-record blockers, and guarded managed-app launch actions
- make Game detail workspaces full-bleed with pinned Bundle, Assets, and Feedback tabs
- show every blocking asset inline with its full explanation, direct source-record link, and one-click navigation from the bundle warning
- guard integrity responses across workspace navigation so stale verification state cannot enable the wrong launch action

## Test plan

- `cd server && npm test` — passed
- `cd server && npm test -- --run services/games/integrity.test.js services/games/compile.test.js routes/games.test.js` — 14 passed
- `cd client && npm test` — 481 files / 5,358 tests passed
- `cd client && npm test -- --run src/pages/Game.test.jsx src/components/games/GameCompilePanel.test.jsx src/components/games/GameBindings.test.jsx src/components/Layout.test.jsx` — 14 passed
- `cd client && npm run lint` — passed
- `cd client && npm run build` — passed
- `git diff --check` — passed

## Dependency audit

`npm audit --audit-level=high` reports the existing React Router RSC-mode advisory in the unchanged client lockfile. PortOS uses the browser SPA router rather than RSC actions; the automated remediation is a forced breaking downgrade, so dependency changes are intentionally outside this PR.


atomantic:
Closing as superseded by #3207, which implemented the same asset-integrity preflight and merged shortly after this PR was opened. Both branches independently created `server/services/games/integrity.js`, which is why the merge shows `add/add` conflicts on that file and its test.

`main` has since taken two review passes on top of #3207 (`a883f45db`, `4fc336856`) that this branch predates, so merging it now would revert them:

- `hashIfPresent` here swallows the error; `main` logs the errno for non-`ENOENT` so a permissions failure isn't reported to the user as atlas corruption.
- `getGameIntegrity` here calls `inspectCurrentBundle` *after* appending the missing-app blocker; `main` deliberately calls it before, or a byte-perfect bundle falsely reports `stale`.
- The compile handler here collapses "build succeeded but the page could not reload" into "build failed", so a successful build can report `Bundle compilation failed`.

Separately, this branch adds `gameName` to the bundle hash inputs while leaving `BUNDLE_SCHEMA_VERSION` at `2`, which would silently invalidate every already-compiled bundle across installs.

The client-side workspace UX in here is genuinely not on `main` and is worth having — the full-bleed `/game/:id` workspace, the inline blocker explanations with source-record deep links, and the three-way music blocker granularity. That work is re-filed against current `main` as #3228, with the rejected pieces called out so they don't come back.

Leaving the `codex/game-workspace-ux` branch on the remote as a starting point for the client diffs.

PR #3258, opened 2026-07-31
fix([issue-3246]): preserve completed CLI agents on shutdown

### LINKED ISSUE #3246: Direct-CLI shutdown guard drops the .agent-done exception, discarding completed work
Found by the v2.36.0 release-integrity gate (cross-PR interaction).

## Problem

The direct-CLI shutdown guard drops the `.agent-done` exception the TUI guard was built with.

`410be46bb` (#3202) established the abandon-on-shutdown contract with **three** exceptions. `9dc9ac8e9` ("dedupe run teardown… abandons on the **same terms**") replicated it with **two**.

`server/services/agentTuiSpawning.js:548-554`:

```js
if (isHostShuttingDown()
  && !sentinelPresent()
  && !userTerminatedAgents.has(agentId)
  && !pausedAgents.has(agentId)) {
```

`server/services/agentCliSpawning.js:819`:

```js
if (isHostShuttingDown() && !terminatedByUser) {
```

Paused is covered by the earlier return at line 801; only the **sentinel** exception is missing.

## Precondition

Only reachable with the CoS runner unavailable — `spawnDirectly` is otherwise unused (`agentLifecycle.js:568-571`). That narrows the blast radius but does not remove it.

## Reproduction

A programmatic-I/O task whose sole deliverable is the `.agent-done` payload writes the sentinel; the CLI child is still streaming its final message when pm2 SIGTERMs. Line 819 abandons before `finalizeAgent`, so `dispatchTaskOutputHookOnce` (`agentFinalization.js:419`) never reads it.

Next boot requeues the task as interrupted, and `resolveTaskResumePatch` returns `{}` for a `discardWorktree` task (`agentWorktreeCleanup.js:410`) — the replacement agent re-runs the whole reasoning pass whose answer is already on disk.

For an ordinary code task it is worse: `.agent-done` is written *after* `/do:pr`, so the requeue opens a **second PR**.

## Decision (chosen approach — ready to work)

Add the missing `!sentinelPresent()` condition to `agentCliSpawning.js:819`, matching the TUI guard verbatim. A sentinel on disk means the agent already produced its deliverable; abandoning at that point discards completed work regardless of which spawn path produced it.

Since these two guards must agree by construction and have now drifted once, extract the predicate into one shared helper both call, rather than fixing the copy. That is the root-cause action, not a nicety — the drift is the bug.

## Acceptance criteria

- A direct-CLI agent with `.agent-done` present at SIGTERM finalizes and dispatches its output hook rather than being abandoned.
- The TUI and direct-CLI paths call one shared shutdown-abandon predicate; a test asserts both honor the sentinel.

atomantic:
## Summary
- share one host-shutdown abandon predicate across TUI and direct-CLI agents
- let direct-CLI agents with an existing `.agent-done` sentinel continue through finalization
- cover the shared exception policy and direct-CLI sentinel regression

## Test plan
- `cd server && npm test -- lib/hostShutdown.test.js services/agentCliSpawning.test.js services/agentTuiSpawning.test.js`
- `cd server && npm test -- lib/index.test.js`
- `cd server && npm test` *(environment-bound: focused changes pass, but the broad suite has pre-existing PostgreSQL-required failures when no test database is available)*
- `git diff --check`

Closes #3246


atomantic:
The configured Codex review gate reached its 3-round cap and remains fail-closed. Round 2 found and this branch fixed the signal/null-exit success regression. The final round still reported three actionable concerns: stale `.agent-done` files need run scoping, completion needs a durable handoff if shutdown outlives the graceful deadline, and synthetic sentinel-backed success should normalize run-tracking exit metadata. Focused shutdown tests pass, but this PR remains open with review status `dirty` for follow-up; it has not been merged.

PR #3261, opened 2026-07-31
feat([issue-3254]): compose memory into POST sessions

### LINKED ISSUE #3254: Let composed POST sessions include a memory drill (make memory a session topic)
## Problem / Goal

Memory is the one POST practice topic with a real config block that a Full/Quick
composed session still can't draw from. Issue #3252 gave `memory` a full config
block (`enabled`, per-drill-type flags, per-item flags) and a Practice Plan row,
but left it as a `surface: 'standalone'` topic — practised only from its own tab.

The consequence: `PostDrillConfig`'s Session Composition checkboxes are derived
from the topic registry's `surface: 'session'` topics, so `memory` is deliberately
NOT offered there. A user who wants a memory rep interleaved into their daily
POST has to go to the Memory tab and run it separately.

## Context

- Registry: `server/lib/postTopics.js` — `memory` is `{ module: 'memory', surface: 'standalone' }`.
  Flipping it to `'session'` is the one-line half of this change.
- `SESSION_MODULE_OPTIONS` in `client/src/components/meatspace/post/PostDrillConfig.jsx`
  derives from `POST_TOPICS.filter(t => t.surface === 'session')`, so `memory`
  appears as a selectable module automatically once the registry flips.
- The blocker is the launcher: `PostSessionLauncher.jsx` builds `allEnabledDrills`
  from three sources only (`math` / `llm` / `cognitive`). There is no `memory`
  source, so today selecting `memory` in Session Composition would compose an
  EMPTY session — which is exactly why `memory` was excluded from the options list
  in the first place (see the comment on `SESSION_MODULE_OPTIONS`).
- The generator already exists: `POST /api/meatspace/post/drill` handles the three
  `memory-*` types (`server/routes/meatspacePostRoutes.js`, memory branch) and
  `generateMemoryDrill` now scopes its auto-picked item to the user's enabled
  Practice Plan items.
- `usePostSession.finishDrill` already scores memory drills correctly and preserves
  `memoryItemId` so schedule/mastery advancement fires (`MEMORY_DRILL_TYPES` in
  the client's post `constants.js`).

## Proposed approach

1. Flip `memory` to `surface: 'session'` in `server/lib/postTopics.js` AND its
   client mirror in `client/src/components/meatspace/post/constants.js` (the
   `postTopics.mirror.test.js` parity test will fail until both move).
2. Add a fourth drill source in `PostSessionLauncher.jsx`: enumerate
   `config.memory.drillTypes` the same way the cognitive source does, mapping each
   enabled type to `{ type, domain: 'memory', config: { count }, timeLimitSec:
   DOMAINS.memory.timeBudgetSec }`. `DOMAINS.memory.drillTypes` already excludes
   `memory-fill-blank` (it's in `POST_UNSUPPORTED_DRILL_TYPES`), so use that list,
   not the topic's full list.
3. **Empty-pool guard.** `generateMemoryDrill` returns `null` when there are no
   memory items at all, which the route turns into a 500 `MEMORY_DRILL_FAILED`
   mid-session. Gate the memory source on the user actually having at least one
   memory item — the launcher already has no memory-item fetch, so this needs a
   count (cheapest: have `getPostConfig`/the launcher's existing recommendations
   fetch carry a `memoryItemCount`, or fetch `getMemoryItems` in the launcher's
   mount effect alongside the review reps).
4. Decide whether `memory` joins `DEFAULT_CONFIG.sessionModules`. Recommendation:
   **no** — keep the default composition as-is (`['mental-math', 'cognitive']`) so
   existing installs see no change; a user opts in from Session Composition.

## Acceptance criteria

- [ ] Selecting `memory` in Config → Session Composition puts a memory drill into
      Full and Quick sessions, respecting the Practice Plan's per-drill-type and
      per-item switches.
- [ ] A composed memory drill scores and advances the item's schedule/mastery on
      session submit (already wired — assert it end-to-end).
- [ ] An install with zero memory items never composes a memory drill (no
      mid-session `MEMORY_DRILL_FAILED`), and the Session Composition checkbox
      explains why when it's unavailable.
- [ ] `DEFAULT_CONFIG.sessionModules` is unchanged, so existing installs compose
      exactly what they did before.
- [ ] The `postTopics` mirror parity test passes with `memory` as a session topic.

## Out of scope

- Any change to how memory drills are generated or scored.
- Per-item selection *within* a composed session (which memorized text a given
  session's memory drill targets) — the lowest-mastery auto-pick, now scoped to
  enabled items, is the behavior.

Discovered while implementing #3252, which explicitly proposed adding `memory` to
the selectable session modules; it was left out because the composition source
above doesn't exist yet and adding the checkbox alone would let a user build an
empty session.

atomantic:
## Summary

- expose Memory as an opt-in Session Composition module while preserving the existing default mix
- compose enabled memory drill types into Full and Quick sessions
- gate composition on a confirmed enabled memory-item pool and explain unavailable state in Config
- keep Practice Plan memory-rotation counts visible after Memory becomes session-capable

## Test plan

- `cd client && npm test -- --run src/components/meatspace/post/PracticePlan.test.jsx src/components/meatspace/post/PostSessionLauncher.test.jsx src/components/meatspace/post/PostDrillConfig.test.jsx src/components/meatspace/post/constants.test.js`
- `cd server && npm test -- --run lib/postTopics.test.js lib/postTopics.mirror.test.js services/meatspacePost.test.js routes/meatspacePostRoutes.drill.test.js`
- `cd client && npm run lint`
- `cd client && npx vite build`
- broad client suite: one unrelated WordplayTrainer timeout passed on immediate isolated rerun
- broad server suite: environment-bound catalog user-type test requires PostgreSQL; related suites pass

Closes #3254


atomantic:
Review gate summary after the configured 3-round Codex cap:

- Round 1: fixed Element Flash item compatibility and unrunnable Practice Plan previews.
- Round 2: fixed one-line Sequence Recall compatibility and pinned composed memory drills to compatible enabled items.
- Round 3: fixed Config so Memory can only be added to Session Composition when its topic, a supported drill, and a compatible enabled item are all runnable.
- Focused client tests, lint, and production build pass after the final fix.

The final finding was addressed and pushed, but the configured review cap was reached before a clean follow-up verdict; the gate therefore remains inconclusive/fail-closed for merge.

PR #3263, opened 2026-07-31
feat: add progressive Powers technique ladder (#3253)

### LINKED ISSUE #3253: Make the POST Powers drill a technique-anchored ladder — only generate questions with a real fast mental path, and show a visual breakdown on a miss
## Problem / Goal

The POST **Powers** mental-math drill has no difficulty progression and no teaching
layer. It draws a base uniformly from `[2, 3, 5]` and an exponent uniformly from
`2..maxExponent` (default 10), so a first-ever session can hand a user `5^2` (trivial,
25) as question 1 and `5^9` (1,953,125) as question 2. There is no ramp, no pairing of
base difficulty with exponent size, and no cap that keeps a question inside what a
human can actually do in their head.

Two things need to be true, and neither is today:

1. **Every generated question must have a documented fast mental path.** A powers
   question is only a *fast-math* question if there is a real technique that gets you
   to the answer in seconds. `5^9` is reachable — but only by splitting the exponent
   off a memorized anchor (`5^8 = 625² = 390,625`, then `× 5`). It is a late-ladder
   question that presupposes an earlier-taught technique, not a warm-up. Any
   `(base, exponent)` pair with **no** documented path should never be generated at
   all.
2. **A miss must teach the method, visually.** Today a wrong answer shows a
   restatement of the question, not a way to think about it.

Goal: make Powers a **technique-anchored progressive ladder** — the same shape the
multiplication and cognitive drills already use — where each rung introduces one named
mental-math method, the generator only emits pairs that method (or an earlier one)
covers, and a miss renders the worked breakdown for that method.

## Context

**Generator — no ramp, no coupling** (`server/services/meatspacePost.js:1247`):

```js
export function generatePowers(bases, maxExponent = 10, count = 8) {
  bases = Array.isArray(bases) && bases.length > 0 ? bases : [2, 3, 5];
  const questions = [];
  for (let i = 0; i < count; i++) {
    const base = bases[Math.floor(Math.random() * bases.length)];
    const exp = Math.floor(Math.random() * (maxExponent - 1)) + 2; // 2 to maxExponent
    questions.push({ prompt: `${base}^${exp}`, expected: Math.pow(base, exp) });
  }
  return { type: 'powers', config: { bases, maxExponent, count }, questions };
}
```

Base and exponent are drawn **independently**, so difficulty is uncorrelated with
question order and with the base's own difficulty. `2^10` (1024, a recall fact) and
`5^10` (9,765,625) are equally likely draws in the same drill.

**Default config makes the very first session the hardest case**
(`server/services/meatspacePost.js:71`):

```js
'powers': { enabled: true, bases: [2, 3, 5], maxExponent: 10, count: 8, timeLimitSec: 90 },
```

That's 27 equally-likely `(base, exp)` pairs. Roughly 16 of them sit on commonly-held
anchors (`2^2..2^10`, `3^2..3^5`, `5^2..5^4`); the remaining ~11 (≈41%) — `3^6..3^10`,
`5^5..5^10` — have no fast path unless a splitting technique has been taught first.
At `timeLimitSec: 90` across `count: 8`, that's ~11s per question.

**Adaptive difficulty does not rescue it** (`server/lib/postAdaptive.js:21`):

```js
'powers': { field: 'maxExponent', base: 10, min: 4, max: 16, step: 2, harderIsHigher: true },
```

It tunes a single scalar (`maxExponent`) and requires `minSamples: 3` scored sessions
(`ADAPTIVE_DEFAULTS`) before it engages — so a new user gets the full `2..10` range
unconditionally, and even the easiest adaptive setting keeps a uniform base draw.
`maxExponent` alone cannot express "`2^12` is fine but `5^6` is not," which is exactly
the distinction that matters here.

**Prior art to follow — two ladders already exist in this codebase:**

- `server/lib/postMultiplicationLadder.js` — built for the *same* complaint about
  multiplication ("used to start at a fixed 2-digit × 2-digit difficulty … which is far
  too hard as a warm-up"). Hand-authored ordered rungs, mastery-gated advancement,
  `level` stamped into the generated drill config so scored history buckets by level.
- `server/lib/postProgression.js` — `createProgression()` is the shared mastery/floor
  engine (extracted from the multiplication ladder), plus `COGNITIVE_LADDERS` for the
  cognitive drills. Powers should be wired onto `createProgression()`, not reinvented.

Schema already supports laddered drills: `progressive`, `level` (`postValidation.js:107-109`)
survive the drill-request and session-submit round trip.

**The miss surface is a restatement, not a lesson** (`client/src/components/meatspace/post/PostDrillRunner.jsx:242`):

```js
function formatHint(prompt, expected) {
  const match = prompt.match(/^(-?\d+)\s*([+\-x^])\s*(-?\d+)$/);
  if (!match) return `${prompt} = ${expected}`;
  const [, a, op, b] = match;
  if (op === 'x') return `${a} × ${b} = ${expected}`;
  if (op === '^') return `${a}^${b} = ${expected}`;   // "5^9 = 1953125"
  return `${prompt} = ${expected}`;
}
```

Rendered in the training-mode feedback overlay (`PostDrillRunner.jsx:120-127`). The
end-of-session review (`DrillQuestionReview.jsx`, mounted from
`PostSessionSummary.jsx:157`) shows prompt / answered / expected / time and no
breakdown at all.

## Proposed approach

**1. New module `server/lib/postPowersLadder.js`** (pure, side-effect-free; mirrors
`postMultiplicationLadder.js`), built on `createProgression()` from `postProgression.js`.

Define a `POWERS_TECHNIQUES` table. Each entry is one named method plus the exact
`(base, exponent)` pairs it makes tractable:

| id | Technique | Covers | Worked example |
|----|-----------|--------|----------------|
| `recall-2` | Powers of 2 — recall table | `2^2..2^10` | `2^8 = 256` (known) |
| `recall-small` | Small squares & cubes | `3^2..3^5`, `5^2..5^4`, `7^2`, `9^2` | `3^4 = 81` |
| `double-chain` | Double up from `2^10 = 1024` | `2^11..2^16` | `2^13 = 1024 → 2048 → 4096 → 8192` |
| `halve-shift` | `5^n = 10^n / 2^n` — write `10^n`, divide by the known `2^n` | `5^5..5^7` | `5^6 = 1,000,000 / 64 = 15,625` |
| `split-exponent` | `b^(m+k) = b^m · b^k` off memorized anchors | `3^6..3^9`, `5^8..5^10` | `5^9 = 5^8 · 5 = 390,625 · 5 = 1,953,125` |

(The table above is the starting proposal, not frozen — the implementer may adjust
coverage boundaries, but the invariant below is not negotiable.)

**Hard invariant: a pair with no technique entry is never generated.** This is the
"no reliable fast path ⇒ not a reasonable question" rule, enforced in code — the
generator's candidate pool is derived *from* the technique table, never from a
`bases × 2..maxExponent` cross-product.

Ladder rungs are ordered by technique, each rung adding its technique's pairs to the
pool of everything below it (so mastered material keeps recurring as review):

```
L0  recall-2         2^2..2^10
L1  + recall-small   + 3^2..3^5, 5^2..5^4, 7^2, 9^2
L2  + double-chain   + 2^11..2^16
L3  + halve-shift    + 5^5..5^7
L4  + split-exponent + 3^6..3^9, 5^8..5^10
```

Advancement is mastery-gated via `createProgression()` (accuracy + speed over a rolling
window), exactly as multiplication does — earned rungs never age out.

**2. Rewire `generatePowers`** (`meatspacePost.js:1247`) to take the resolved level,
draw only from that rung's pool, and stamp `level` + `technique` into the returned
drill config. Within a drill, order questions so the earliest rung's pairs come first
(warm-up → hardest), rather than shuffling uniformly. Keep the legacy
`bases`/`maxExponent` path intact for manual (non-progressive) config, matching how
`generateMultiplication` keeps its `maxDigits` legacy shape.

**3. Wire the level resolution** in `resolveDrillConfig` alongside the existing
multiplication/cognitive progressive paths, so the Powers drill picks up progression the
same way. `postAdaptive.js`'s `maxExponent` knob stays for manual mode only.

**4. Visual breakdown on a miss** — new `client/src/lib/powersBreakdown.js`, a pure
function `powersBreakdown(base, exponent) → { technique, label, steps: string[] }`
returning the worked lines for that pair's technique:

```
5^9  —  split the exponent
  5^4 = 625                      (anchor)
  5^8 = 625 × 625 = 390,625
  5^9 = 390,625 × 5 = 1,953,125
```

```
2^13  —  double up from 2^10
  2^10 = 1,024
  ×2 → 2,048   ×2 → 4,096   ×2 → 8,192
```

Render it in **both** miss surfaces: the training-mode overlay (replacing the `^` branch
of `formatHint` in `PostDrillRunner.jsx:242`) and per missed question in
`DrillQuestionReview.jsx`. Because it's derived from `(base, exponent)` at render time —
not stamped into the persisted question — historical sessions get breakdowns too, and no
`questionResultSchema` change is needed.

**5. Bind the two halves with a test** that imports the server ladder's full generatable
pair set and asserts `powersBreakdown()` returns a non-fallback, multi-step explanation
for **every** pair. That test is what makes "no reliable fast path ⇒ never generated"
enforceable rather than aspirational. It imports a client lib from a node test, so
`client/src/lib/powersBreakdown.js` must stay node-safe (no DOM/JSX imports).

**6. Migration** — existing installs carry `powers: { bases, maxExponent, … }` in their
stored POST config. Add `progressive: true` as the default for new installs and treat a
stored config without it as progressive-on (opting into the ladder), with manual mode
still reachable from `PostDrillConfig.jsx`. No on-disk shape change beyond the additive
`progressive`/`level` fields the schema already accepts.

## Acceptance criteria

- [ ] `server/lib/postPowersLadder.js` exists, is pure/side-effect-free, and builds its
      ladder on `createProgression()` from `postProgression.js`.
- [ ] A `POWERS_TECHNIQUES` table maps every generatable `(base, exponent)` pair to a
      named technique; the generator's candidate pool is derived from that table.
- [ ] **No `(base, exponent)` pair outside the technique table can ever be generated** —
      asserted by a test that enumerates the full pool.
- [ ] A user with no scored history starts at rung 0 (powers of 2 recall). A brand-new
      user can no longer be served `5^9`, `5^10`, or `3^10`.
- [ ] Within one drill, questions are ordered easiest-rung-first rather than uniformly
      shuffled.
- [ ] Advancement is mastery-gated (accuracy + speed over the rolling window) and earned
      rungs never age out, matching `postMultiplicationLadder.js` semantics.
- [ ] `level` (and the technique id) are stamped into the generated drill config and
      survive the session-submit round trip through `postValidation.js`.
- [ ] `client/src/lib/powersBreakdown.js` returns a multi-line worked breakdown for every
      pair the ladder can generate; a cross-importing test proves the coverage is total.
- [ ] A missed powers question shows the visual breakdown in the training-mode overlay
      (`PostDrillRunner.jsx`) **and** in the end-of-session review
      (`DrillQuestionReview.jsx`) — not the current `5^9 = 1953125` restatement.
- [ ] Manual (non-progressive) config still works via the existing
      `bases`/`maxExponent` path in `PostDrillConfig.jsx`.
- [ ] `client/src/lib/powersBreakdown.js` is node-safe and registered in the
      `client/src/lib/` barrel + `README.md` per the module-organization rule.
- [ ] Tests: `server/lib/postPowersLadder.test.js` (rung ordering, pool derivation,
      mastery gating, no-uncovered-pairs) and a client test for the breakdown renderer.

## Out of scope

- Reworking the other math drills (`doubling-chain`, `serial-subtraction`,
  `estimation`) — multiplication and the cognitive drills already have ladders; those
  three are separate follow-ups if they show the same complaint.
- Changing `postAdaptive.js`'s generic knob mechanism. It stays for manual mode; the
  ladder supersedes it only for progressive Powers.
- Adding new bases beyond `{2, 3, 5, 7, 9}` or a general n-th-root/logarithm curriculum.
- Any LLM-generated hint text — the breakdowns are deterministic, hand-authored per
  technique (no provider call on a drill miss).

atomantic:
## Summary
- replace random progressive Powers questions with cumulative, named mental-math technique rungs
- teach deterministic worked steps after misses in training and session review
- preserve manual base/exponent mode and expose Powers ladder mastery in Drill Config

## Test plan
- npm test -- lib/postPowersLadder.test.js lib/postValidation.test.js lib/index.test.js services/meatspacePost.test.js (server)
- npm run lint (client)
- npm test -- --run src/lib/powersBreakdown.test.js src/lib/index.test.js src/components/meatspace/post/PostDrillRunner.test.jsx src/components/meatspace/post/DrillQuestionReview.test.jsx src/components/meatspace/post/PostDrillConfig.test.jsx (client)

Closes #3253


atomantic:
Review handoff after the configured 3-round Codex cap:

- Round 1: fixed mastery attribution so cumulative lower-rung questions cannot unlock a higher technique.
- Round 2: added Powers maintenance-review scheduling and stalled-practice recommendations.
- Round 3: fixed the final finding so due Powers maintenance reps honor topic, session-composition, module, and per-drill opt-outs.
- Final focused validation passes: server 261 tests; client 59 tests; client lint; syntax and diff checks.

The last finding is fixed in dd45a2b19, but the reviewer cap is exhausted, so this PR remains open for a fresh human/reviewer confirmation rather than merging without a clean final verdict.

PR #3272, opened 2026-08-01
fix(cos): align task estimates with learning buckets

atomantic:
Automated PR created by PortOS Chief of Staff.

## Summary

Aligned CoS task ETA displays with the authoritative task-learning buckets so scheduled and archived self-improvement work uses its own history.
## Changes
- Added a documented, barrel-exported client task-type resolver that mirrors task-learning's metadata-first classification.
- Updated pending task and running agent ETA surfaces to use the shared resolver instead of divergent text-only heuristics.
- Added unit and rendered-UI regressions for scheduled and archived-agent estimate selection.
- Documented the learning-aligned ETA behavior in the Chief of Staff feature guide.
- Validated focused client tests (25 passing), lint, production build, and diff checks.


atomantic:
Review loop update: Codex completed two rounds (one ETA bucket-source correction was applied and re-review was clean). The required pinned Ollama reviewer was attempted twice, but its local review endpoint returned an empty HTTP reply both times, so no valid findings verdict was available. Per the review contract, this is inconclusive rather than clean; PR remains open and unmerged.

Issue #3288, opened 2026-08-01
Extract videoGen POST param preparation into a service, mirroring imageGen


atomantic:
`POST /` in `server/routes/videoGen.js` (lines ~745-1461) is a single ~717-line handler that does request validation, backend-pin resolution (grok vs local), model-existence checks, multipart upload staging for four distinct upload kinds (source / last-frame / audio / IC-reference) with rollback bookkeeping, IC-LoRA mode and reference-count validation, keyframe validation, history lookup for extend + IC reference, LoRA-array normalization, chunk-count resolution, and finally job enqueueing — all inline in a route file.

Per the module contract in `CLAUDE.md`, multi-step domain orchestration over models + services belongs in `server/services/`, not `server/routes/`.

**The sibling route already shows the shape.** `server/routes/imageGen.js` `POST /generate` is ~134 lines because it extracted this exact concern into `server/services/imageGen/prepareParams.js` + `server/services/imageGen/variants.js`. `videoGen.js`'s own header comment says it "mirrors the imageGen route surface", but it never got the same treatment.

### Why it matters
Twenty-plus distinct `ServerError` throw sites share mutable state via `stagedDurablePaths` / `cleanupAllStaged`. Every new video-gen parameter (the file already carries inline references to #3231, #2859, #1760) means editing this monolith, and any fix risks touching an unrelated validation branch or a cleanup path.

### Proposed shape
Add `server/services/videoGen/prepareParams.js` mirroring `imageGen/prepareParams.js`. It returns
`{ backend, effectiveModel, sourceImagePath, lastImagePath, audioFilePath, icReferencePaths, loras, effectiveChunks, ... }`
or throws `ServerError`. The route becomes: parse -> call service -> `enqueueJob` -> respond.

### Constraints
- **Validation order is load-bearing** — several throw sites depend on earlier ones having run. Preserve the exact sequence.
- **Cleanup-on-throw must be preserved exactly.** Every early return currently unwinds staged uploads; a missed path leaks files.
- Land the cleanup test coverage first (see the companion issue on `videoGen.test.js` never asserting `unlink`/`copyFile`) — without it this refactor has no safety net.

Found by a `/do:better` DevSecOps audit (2026-08-01). Deferred from automated remediation as too high-risk to do unattended.

PR #3324, opened 2026-08-02
test: assert videoGen POST / rolls back staged uploads on every failure path

### LINKED ISSUE #3289: Cover videoGen upload-cleanup and rollback paths in videoGen.test.js
`server/routes/videoGen.js` `POST /` has 20+ throw sites that call `cleanupAllStaged()` / `cleanupTempUploads()` before throwing, plus a dedicated failed-durable-copy branch at ~`videoGen.js:867` that does `copyFile(...).catch(async (err) => { unlink(durablePath); cleanupAllStaged(); throw ... })`.

`server/routes/videoGen.test.js` mocks `fs/promises` (`unlink: vi.fn(async () => {})`, `copyFile: vi.fn(async () => {})`, ~lines 130-135) but **never asserts on either mock** — `grep -n "unlink\b" server/routes/videoGen.test.js` returns only the mock declaration. Across all 97 tests in the 1512-line suite, every error-path case asserts only `r.status` / `r.body` and `enqueueJob).not.toHaveBeenCalled()`.

### Consequence
A regression that stopped calling `cleanupAllStaged()` entirely, or unlinked the wrong path, would not fail a single test — staged uploads would silently accumulate on disk on every rejected request. The `copyFile` rejection branch is never exercised at all, since the mock always resolves.

### Proposed coverage
1. Mock `copyFile` to reject once; assert the route responds with an error **and** that `unlink` was called with the staged `durablePath` and with every path in `stagedDurablePaths`.
2. For at least one representative late-stage throw (e.g. the music-video scene lookup at ~`videoGen.js:1118`), assert `unlink` was called with the pre-staged audio and IC-reference upload paths.
3. Assert the happy path does *not* unlink the durable destination.

Each new test must be verified falsifiable: break the cleanup call, confirm the test fails, restore.

This is a prerequisite for #3288 (extracting the handler into a service) — that refactor needs this safety net first.

Found by a `/do:better` DevSecOps audit (2026-08-01).

atomantic:
## Summary

`server/routes/videoGen.js` `POST /` has 20+ throw sites that call `cleanupAllStaged()` / `cleanupTempUploads()` before rethrowing, plus a `copyFile` rejection branch in `stageUploadDurable` that unlinks its own half-written destination and rolls back every earlier staged copy. `videoGen.test.js` mocked `unlink` and `copyFile` but never asserted on either mock — across all 97 tests, error paths asserted only `r.status` / `r.body` / `enqueueJob` not-called. A regression that dropped the cleanup entirely, or unlinked the wrong path, stayed green while leaking a durable copy under `data/uploads` on every rejected request. The `copyFile` rejection branch was never exercised at all, since the mock always resolved.

Five new cases in a `POST / — upload staging rollback (#3289)` block:

1. **`copyFile` rejects mid-staging** — `sourceImage` stages, `audioFile`'s copy fails. Asserts 500 `VIDEO_GEN_UPLOAD_STAGE_FAILED` **and** that `unlink` got the half-written destination, the already-staged `sourceImage` durable copy (`stagedDurablePaths`), and both multipart temp files.
2. **Late-stage throw — music-video scene 404** — the scene lookup runs after the reference frame is already copied into `data/uploads`; asserts that copy is unlinked.
3. **Late-stage throw — keyframes/mode mismatch** — fires with both a staged source frame and a staged IC-reference clip in flight; asserts both durable copies plus both temp files are unlinked.
4. **Late-stage throw — legacy inputs vs keyframes** — the only path that stages an uploaded `lastImage` before a rejection, so a durable copy missing from `stagedDurablePaths` would leak here and nowhere else.
5. **Happy path** — `copyFile` wrote the durable destination, the job was handed that exact path, the OS temp file is dropped, and nothing under `data/uploads` is — the queue worker owns the durable copy's lifetime.

The only production-code-adjacent change is to the test harness: `setPendingUpload` now accepts several files so one request can carry more than one durable copy in flight. That is what makes the rollback observable at all. `server/routes/videoGen.js` is untouched.

This unblocks #3288 (extracting the `POST /` handler into a service) — that refactor needs this safety net in place first.

Review also surfaced a genuine production leak that is out of scope for a test-only change and is filed separately as #3326: a *rejected await* after staging (e.g. `getMusicVideoProject`, `getTrack`, `loadHistory`, `enqueueJob`) bubbles past `cleanupAllStaged()` entirely, since only explicit `throw` sites call it.

## Test plan

- `cd server && npx vitest run routes/videoGen.test.js` → 110 passed (was 105).
- Full `cd server && npm test`: the 54 failing files are all Postgres-backed suites failing identically on unmodified `main` in this environment (no local Postgres); none is `videoGen.test.js`.
- **Falsifiability** — each cleanup call was broken in turn in `videoGen.js`, the suite re-run, and the source restored. Every break failed exactly one test, the intended one:
  - drop `unlink(durablePath)` in `stageUploadDurable`'s catch → case 1 fails
  - drop `cleanupAllStaged()` in `stageUploadDurable`'s catch → case 1 fails
  - drop `cleanupAllStaged()` before the music-video 404 → case 2 fails
  - drop `cleanupAllStaged()` before the keyframes mismatch throw → case 3 fails
  - drop `cleanupAllStaged()` before the legacy-inputs conflict throw → case 4 fails
  - pop `lastImage` off `stagedDurablePaths` → case 4 fails
  - unlink `durablePath` instead of `file.path` on the success path → case 5 fails
  - skip staging entirely (hand the raw temp path through) → case 5 fails

Closes #3289

### DIFF
--- a/server/routes/videoGen.test.js
+++ b/server/routes/videoGen.test.js
@@ -80,9 +80,11 @@ vi.mock('../services/mediaJobQueue/index.js', () => ({
 // this via `setPendingUpload({ fieldname, ... })` before issuing the request;
 // the mocked uploadFields middleware reads it off the holder, attaches it as
 // req.files keyed by fieldname, and clears it. Mutable wrapper avoids
-// reaching into vi mock internals.
+// reaching into vi mock internals. Several files may be staged in one request
+// (`setPendingUpload(a, b)`) — the rollback cases need more than one durable
+// copy in flight to prove a later failure unwinds the earlier ones.
 const pendingUpload = { current: null };
-const setPendingUpload = (file) => { pendingUpload.current = file; };
+const setPendingUpload = (...files) => { pendingUpload.current = files.flat(); };
 
 vi.mock('../lib/multipart.js', () => ({
   // Bypass the streaming parser. If a test set a pending upload via
@@ -90,8 +92,7 @@ vi.mock('../lib/multipart.js', () => ({
   // route exercises the upload-staging path; otherwise pass through.
   uploadFields: () => (req, _res, next) => {
     if (pendingUpload.current) {
-      const f = pendingUpload.current;
-      req.files = { [f.fieldname]: f };
+      req.files = Object.fromEntries(pendingUpload.current.map((f) => [f.fieldname, f]));
       pendingUpload.current = null;
     }
     next();
@@ -134,6 +135,7 @@ vi.mock('fs/promises', () => ({
   copyFile: vi.fn(async () => {}),
 }));
 
+import { copyFile, unlink } from 'fs/promises';
 import * as videoGenService from '../services/videoGen/local.js';
 import * as mediaJobQueue from '../services/mediaJobQueue/index.js';
 import { getProject as getMusicVideoProject } from '../services/musicVideo/projects.js';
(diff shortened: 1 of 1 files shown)


atomantic:
Heads-up: #3288 just merged (PR #3349) and moved the `POST /` staging + validation region out of `server/routes/videoGen.js` into `server/services/videoGen/prepareParams.js`.

Nothing in this PR's target file changed — `server/routes/videoGen.test.js` was deliberately left untouched to avoid a conflict — and the route's HTTP behavior (including every `copyFile`/`unlink` call these tests assert) is preserved verbatim, so a rebase onto `main` should be clean and the assertions should still hold.

One overlap worth knowing about: #3349 also closed #3326, so the rejected-await rollback (`getMusicVideoProject` / `getTrack` / `loadHistory` / `enqueueJob`) is now guarded by `withStagedRollback` and covered at the service level in `server/services/videoGen/prepareParams.test.js`. This PR's explicit-throw coverage is still the route-level half and doesn't duplicate it.

PR #3325, opened 2026-08-02
Classify every data/ directory so the Data Manager's biggest rows are actionable

### LINKED ISSUE #3285: [ux-data-manager-unknown-category-rows] Biggest data rows are unactionable
**Screen:** `/data` (Data Manager) — audited 2026-08-01 at 1440x900 and 375x812
**Checklist failed:** 4 (copy states an outcome, not a mechanism) and 5 (no dead ends)

**What the user is trying to do:** work out what is eating 51.9 GB and reclaim some of it — the page's own subtitle is "Storage visibility, archiving, and cleanup".

**Why the current design impedes it:** The single largest category — 15.9 GB across 1,303 files, called out by name in the "Largest" stat tile — is described as **"Unknown category"**. So are `python` (1.4 GB), `image-to-3d` (1.3 GB), `sprites` (921.9 MB), `lora-datasets` (493.7 MB), `image-refs` (44.5 MB) and others: 7 of the top 13 rows by size tell the user nothing about what the data is or whether losing it costs anything. Worse, this is a dead end rather than just missing copy — `server/services/dataManager.js:66` falls back to `{ label: d.name, description: 'Unknown category', archivable: false, deletable: false }`, so an unrecognised directory is *also* rendered with no archive and no delete affordance. The biggest thing on a cleanup page is the one thing the cleanup page refuses to act on, and the UI never says why.

**Proposed change:** Two parts. (1) Add the missing entries to the `CATEGORIES` map in `server/services/dataManager.js` — at minimum `training-runs`, `sprites`, `lora-datasets`, `image-to-3d`, `image-refs`, `python` — each with a real description and honest `archivable` / `deletable` flags, and add a test asserting every directory the scanner can emit has a `CATEGORIES` entry so new data dirs can't silently regress to "Unknown". (2) In `client/src/pages/DataManager.jsx`, when a row does fall through to the unknown fallback, say so in outcome terms — "Not classified — PortOS doesn't know if this is safe to remove" — instead of the bare label, so the absent Archive/Delete buttons read as a deliberate safety stance rather than a broken row.

Scope: medium

atomantic:
## Summary

On `/data`, seven of the thirteen largest rows — including the single biggest directory on disk — rendered as "Unknown category". The fallback in `dataManager.js` also returned `archivable: false, deletable: false`, so an unrecognised directory lost its Archive/Purge affordances entirely: the biggest thing on the cleanup page was the one thing the cleanup page refused to act on, and the UI never said why.

- **Classified every top-level `data/` directory** (~65 entries, up from 23) with a real description and conservative permission flags. `deletable` is only true where the bytes are genuinely reproducible or already duplicated elsewhere, and the description says what survives — e.g. `training-runs` is purgeable because the finished adapters live in `data/loras/`, `python` because the virtualenv rebuilds itself. Generated art that exists nowhere else (`sprites`, `image-to-3d`, `lora-datasets`) stays protected, as do `certs` (private key), `sharing` (purging it would force a full resync and can resurrect deleted records) and the legacy `pipeline-issues` / `pipeline-series` recovery dirs.
- **Added a coverage guard** (`server/services/dataManager.categories.test.js`). The directory set is *derived*, never hand-maintained: `PATHS` in `lib/fileUtils.js`, backup's `DEFAULT_EXCLUDES`, a source scan of `server/` + `scripts/` for the `join(PATHS.data, 'x')` / `` `${PATHS.data}/x` `` / `join(root, 'data', 'x')` spellings, and `data.reference/`. It caught one directory (`pipeline-comparative-rank`) that manual review had missed.
- **Outcome-phrased fallback copy.** `getDataOverview` / `getCategoryDetail` now emit `classified: false` for an unrecognised directory with the description "Not classified — PortOS doesn't know if this is safe to remove". `DataManager.jsx` styles that row as a warning and, on expand, replaces the generic "protected" line with an explanation that Archive and Purge are being withheld deliberately plus what the user can do instead. The client treats only an explicit `classified === false` as unknown, so an older server response degrades to the existing behaviour.

## Test plan

- `cd server && npx vitest run services/dataManager.categories.test.js services/dataManager.test.js services/backup.test.js lib/index.test.js` — 59 passed.
- `cd client && npx vitest run src/pages/DataManager.test.jsx` — 2 passed (new suite: the unclassified row renders the outcome copy, expands to the withheld-actions explanation with no Archive/Purge buttons, and a known protected category keeps its original wording).
- Verified the coverage test fails loudly when a directory is unclassified (it did, on `pipeline-comparative-rank`) before the entry was added.

Closes #3285


atomantic:
Review loop: `codex[gpt-5.6-terra]`, 3 rounds, all findings applied.

- **Round 1** — coverage guard scanned only three spellings of a data path and missed `dataPath('x')`, leaving `data/spotify` and `data/youtube` unclassified while the test stayed green. Added the pattern, classified both (`fbdb558c6`).
- **Round 2** — the guard matched only inline string literals, so the collection-store idiom `join(PATHS.data, TYPE)` was invisible and `data/commission-feedback` slipped through. The scan now resolves the captured identifier against its `const` declaration in the same file (`84a72621a`).
- **Round 3** — clean: no remaining unclassified directories, no new correctness bugs.

Rounds 1 and 2 also flagged the pre-existing `deletable: true` on `images`, `videos`, `loras`, `messages`, `screenshots`. Those predate this branch and flipping them would silently remove a capability, so they're unchanged here — the genuinely unrecoverable cases are filed as #3327 (`loras` is excluded from backups *and* purgeable, which is the sharpest edge).

Issue #3284, opened 2026-08-01
[ux-prompt-manager-no-stage-search] 40+ prompt stages with no search


atomantic:
**Screen:** `/prompts` (Settings → Prompt Manager, Stages tab) — audited 2026-08-01 at 1440x900 and 375x812
**Checklist failed:** 7 (information hierarchy matches the user's task)

**What the user is trying to do:** find one specific prompt stage and edit it.

**Why the current design impedes it:** The Stages pane is a flat, unfiltered, unsorted-beyond-alphabetical list of 40+ entries in a narrow column, and the page has **no search or filter input at all** — `client/src/pages/PromptManager.jsx` uses `useSearchParams` only to hold the selected stage in the URL, never to filter the list. Finding "Creative Director — Treatment" among rows that also include several other `Creative Director — …`, `Brain …`, `CoS …`, `Catalog Ingest — …` and `Importer — …` entries means scrolling a 7,410px page and reading two-line descriptions. Compounding it, every row carries a bare trash icon at rest, immediately to the right of the row's own click target, so a mis-tap while hunting through 40 rows lands on Delete rather than select.

**Proposed change:** In `client/src/pages/PromptManager.jsx`, add a search input pinned to the top of the Prompt Stages pane that filters rows on title + description as you type, and group the filtered rows by their name prefix (Brain, CoS, Creative Director, Importer, Catalog Ingest, …) with a `SYSTEM`-only toggle, so the taxonomy already encoded in the names becomes navigable structure. Also demote the per-row delete: reveal it on row hover/focus, or move it into the detail pane for the selected stage, so the destructive control is not a permanent neighbour of the row's select target.

Scope: small

PR #3333, opened 2026-08-02
Serve the prompt system-stage list from the API instead of mirroring it in three places

### LINKED ISSUE #3314: Serve the system-stage list from the API instead of mirroring it in three places
Discovered while working issue #3284 (Prompt Manager stage search + grouping).

## Problem

The set of "system" prompt stages — the ones PortOS features call by name and that `DELETE /api/prompts/:stage` refuses to remove — is hardcoded in three places:

1. `server/routes/prompts.js`, the `GET /:stage/usage` handler (a map of `key -> usedBy[]`).
2. `server/routes/prompts.js`, the `DELETE /:stage` handler (a bare array of the same keys).
3. `client/src/lib/promptStageGroups.js`, `SYSTEM_STAGE_KEYS` (relocated there from an inline array in `client/src/pages/PromptManager.jsx`).

`GET /api/prompts` returns the raw stage-config map and says nothing about which stages are system stages, so the client has no choice but to guess.

This was survivable while the client copy only drove a cosmetic `SYSTEM` badge. #3284 turned it into a **navigation affordance**: the Stages pane now has a "System only" filter over a 127-item list. A key missing from the client copy means that stage is unreachable through the filter even though the server would refuse to delete it.

The lists have already drifted in a visible way: `cos-task-enhance` appears in all three lists but has no entry in `data.reference/prompts/stage-config.json` at all (it is referenced from `server/services/taskEnhancer.js`), so it is a key that can never match a rendered row.

## Proposed change

1. Extract the `key -> usedBy[]` map into one module (`server/lib/promptSystemStages.js`), exporting the map plus a derived key list and an `isSystemStage(key)` predicate. Register it in the `server/lib/` barrel + README per the module-organization rule.
2. Point both `server/routes/prompts.js` handlers at it — the DELETE guard derives its array from `Object.keys(...)` rather than restating it.
3. Have `GET /api/prompts` return the key list alongside the stages (e.g. `{ stages, systemStages }`) — the route already has the data.
4. Client: `PromptManager` reads the list off the load response and threads it into `buildStageGroups({ systemStageKeys })`; delete `SYSTEM_STAGE_KEYS` / `isSystemStage` from `client/src/lib/promptStageGroups.js` and its README row.
5. Decide `cos-task-enhance` while you are in there: either ship a stage-config entry for it (with a `data.reference` seed + migration) or drop it from the system list.

## Scope

Small. Net code reduction on the client; roughly 15 new server lines. No schema change, no migration beyond whatever step 5 concludes.

## Not in scope

A declared `family` / `category` field on the stage schema (which would retire the name-prefix grouping heuristic in `promptStageGroups.js`) — that needs a schema change plus a backfill across 127 records and belongs in its own issue.

atomantic:
## Summary

The set of prompt stages PortOS features call by name was hardcoded in three places — the `key -> usedBy[]` table in `GET /api/prompts/:stage/usage`, a bare key array in the `DELETE /api/prompts/:stage` guard, and `SYSTEM_STAGE_KEYS` in `client/src/lib/promptStageGroups.js`. Since #3284 the client copy drives a navigation affordance (the Stages pane's "System only" filter over a 127-row list), so a key missing from it makes that stage unreachable through the filter even though the server still refuses to delete it.

- New `server/lib/promptSystemStages.js` holds the one table: `SYSTEM_STAGE_USAGE` (`key -> usedBy[]`), the derived `SYSTEM_STAGE_KEYS`, `isSystemStage(key)`, `systemStageUsedBy(key)`. Registered in the `server/lib/` barrel + README per the module-organization rule.
- Both `server/routes/prompts.js` handlers read from it; the DELETE guard derives its keys instead of restating them.
- `GET /api/prompts` now returns `{ stages, systemStages }`.
- `PromptManager` reads the list off the load response and threads it into `buildStageGroups({ systemStageKeys })`. `SYSTEM_STAGE_KEYS` / `isSystemStage` are gone from `client/src/lib/promptStageGroups.js` and its README row.

### `cos-task-enhance` (issue step 5) — shipped a real stage rather than dropped from the list

`server/services/taskEnhancer.js` genuinely resolves through the prompt-stage config: `getStage('cos-task-enhance')` for provider/model and `buildPrompt('cos-task-enhance', …)` for the template — the hardcoded `FALLBACK_PROMPT` is only the miss path. v0.10.29 shipped it as "configure via Prompt Manager instead of hardcoded defaults", but no `stage-config.json` entry ever existed, so the stage rendered no row, there was nothing to configure, and every run silently took the fallback with the active provider's default model. Dropping the key from the system list would have made that permanent; the smaller change would also have been the wrong one.

So it now ships as a real stage: `data.reference/prompts/stages/cos-task-enhance.md` (the previous hardcoded fallback, mustache-ified) + a `stage-config.json` entry (`model: "default"` — the tier that resolves to exactly the model the service already used, so no behavior change on upgrade) + `scripts/migrations/218-cos-task-enhance-stage.js` via the standard `makeSeedMigration` helper, since boot runs migrations but not `setup-data.js`.

One coupled fix that fell out of it: with a stage-config entry present, `stage.model` is non-null in `taskEnhancer.js` for the first time, and the service passed it to the provider verbatim — so the literal string `default` would have gone out as a model name. It now resolves tiers through `resolveModel()` from `stageRunner.js` like every other stage caller.

### Wire-shape change

`GET /api/prompts` previously returned `{ stages }`. The only consumer anywhere in the repo is `getPrompts()` in `client/src/services/apiPrompts.js`, called once from `PromptManager.jsx`; the endpoint is not part of any peer-sync or federation payload, so no older install reads it. The change is additive either way — an older client ignores the new key, and the new client uses `Array.isArray(res.systemStages)` so an older server (which sends nothing) degrades to "no system stages known" rather than crashing. Both directions are covered by tests.

## Test plan

- `server/lib/promptSystemStages.test.js` (new): key list is derived not restated, predicate coverage, prototype keys (`toString`/`constructor`) don't leak through as usage, and a drift guard that every system-stage key has a shipped `stage-config.json` entry — the exact failure `cos-task-enhance` was in.
- `server/routes/prompts.test.js` (new): pins the `{ stages, systemStages }` wire body, the usage report for system vs user stages, the DELETE guard + `?force=true`, and that the guard and the usage endpoint agree on the same list. The client suite mocks `apiPrompts`, so this is the only place the response shape is actually asserted.
- `scripts/migrations/218-cos-task-enhance-stage.test.js` (new): standard `runSeedStageMigrationTests` suite — seeds when absent, no-op re-run, no-clobber of a customized template, and a `data.reference` drift catch.
- `client/src/lib/promptStageGroups.test.js`: `systemOnly` now drives off the passed `systemStageKeys`; added cases for an empty/absent list and for `systemStageKeys` being ignored when `systemOnly` is off.
- `client/src/pages/PromptManager.test.jsx`: added a case proving the badge and filter follow whatever the server names (a stage no hardcoded client array ever mentioned), and one for an older server that omits the key.
- `cd client && npm test` → 515 files / 5892 tests pass.
- `cd server && npm test` → the 54 failing files are all pre-existing environment failures in this worktree (Postgres not reachable: `Pipeline series require PostgreSQL`, `pgFileFacade`, …) plus `services/taskPromptDefaults.test.js`, which fails identically on a clean `main`. Every suite touched by this change passes.

Closes #3314


atomantic:
Review-gate note (codex `gpt-5.6-terra`, 2 rounds):

Round 1 flagged that the extracted registry omits the ~100 other stages PortOS resolves by literal key (`pipeline-series-concept-judge`, `catalog-ideas-scenes-concepts`, `soul-contradiction-detector`, …), which are therefore still deletable without `?force=true`. That is pre-existing behavior, not introduced here — the same ten keys were guarded before the extraction — and adding them to this table would also badge ~100 of 127 rows `SYSTEM`, destroying the "System only" filter #3284 added. Fixed in-branch by documenting the registry's scope honestly (curated protected set, not an index of every call site); the real fix is filed as #3335 with a decided design (split the badge set from a generated deletion-protection manifest).

Round 2: no branch-introduced correctness findings.

Confirmed by review: `PromptManager` is the only production consumer of `GET /api/prompts`, and it handles a missing `systemStages` key safely.

PR #3337, opened 2026-08-02
Paginate tracks/collections lists and fold three duplications into existing helpers (#3294)

### LINKED ISSUE #3294: Small cleanups: reuse formatDurationSec, paginate tracks/mediaCollections, hoist getActiveApps
Three small independent cleanups from the 2026-08-01 audit, grouped because none justifies its own issue.

### 1. `WorkEditor` hand-rolls an existing duration formatter
`client/src/components/writers-room/WorkEditor.jsx:1235-1237` — `AnalysisRunBanner`'s elapsed readout computes
`const mm = Math.floor(elapsed / 60); const ss = elapsed % 60; const timeStr = \`${mm}:${ss.toString().padStart(2, '0')}\``
which is byte-for-byte the output of `formatDurationSec(seconds)` in `client/src/utils/formatters.js:217-222`.
Single site (not a systemic pattern), but the project convention is to never redefine a formatter inside a component.
Fix: `import { formatDurationSec } from '../../utils/formatters.js'` and use it.

### 2. Two list endpoints return unbounded results while their siblings paginate
`server/routes/tracks.js:180` (`res.json(await tracks.listTracks())`) and `server/routes/mediaCollections.js:54` (`res.json(await svc.listCollections())`) return the full list with no limit/offset, while structurally identical list endpoints — `server/routes/albums.js`, `artists.js`, `loras.js:35-40` — already support `?limit`/`?offset` via `isPaginationRequested` / `paginateArray` from `server/lib/validation.js`.
Tracks accumulate with every upload and generation, so these two are unbounded and growing, and a shared list-consuming client component has to special-case them.
Fix: apply the same `isPaginationRequested` / `paginateArray` pattern `loras.js` uses. That pattern stays backward-compatible — it returns the full array when neither param is present and only switches to the bounded envelope once a caller opts in.
(Note: `parsePagination`'s boundary contract is now pinned by tests in `server/lib/validation.test.js`.)

### 3. `getActiveApps()` re-fetched inside a loop
`server/services/cosTaskGenerator.js:705-733` — `spawnPriority0OnDemand` calls `await getActiveApps().catch(() => [])` *inside* `for (const request of onDemandRequests)` rather than once above it. `getActiveApps()` is backed by a 2s in-memory cache so this is usually a cache hit, but a burst of more than one on-demand request within a cycle (or a miss at the cycle boundary) means a real disk read per request instead of per cycle. The app registry cannot change mid-loop.
Fix: hoist the call above the `for`.

Found by a `/do:better` DevSecOps audit (2026-08-01).

atomantic:
## Summary

Ships the three cleanups in #3294 plus the two follow-ups added in its comment thread.

**Pagination (the substantive one).** `GET /api/tracks` and `GET /api/media/collections` returned the full list with no ceiling, while the structurally identical `albums`, `artists` and `loras` list endpoints already accept `?limit`/`?offset`. Tracks grow with every upload and generation, so these two were unbounded and growing. Both now use the same `isPaginationRequested` / `paginateArray` pattern (`defaultLimit: 50`, `maxLimit: 500`), which stays backward-compatible: with neither param present the response is still the bare array, and only a caller that opts in gets the `{ items, total, limit, offset }` envelope. Every current client caller (`apiTracks.listTracks`, `apiImageVideo.listMediaCollections`) passes no query params, so nothing changes for them — and each route gained a regression test pinning that bare-array default, mirroring the guard `albums.test.js` already carries.

**`formatDurationSec` reuse.** `AnalysisRunBanner` in `WorkEditor.jsx` hand-rolled `mm:ss`. Output is identical for the values it produces (`elapsed` is always a non-negative integer); the only divergence is negative input, where the helper returns `—` instead of the old `"-1:-5"` garbage.

**`getActiveApps()` hoist.** `spawnPriority0OnDemand` re-fetched the app registry inside its `for (const request of onDemandRequests)` loop. The registry cannot change mid-loop, and the 2s cache can miss at a cycle boundary, so it now reads once above the loop.

**Feature Agents staleness (from the issue comment).** `OutputTab`'s refresh button called `setOutput` outside the fetch effect, so it was not covered by that effect's `cancelled` guard — a refresh landing after an agent switch could paint the wrong agent's output. Both paths now share one loader guarded against a ref holding the current selection. Neither `OutputTab` nor `RunsTab` cleared prior state on `agent.id` change, so the previous agent's output/runs stayed on screen until the new fetch landed; both now clear on switch. The output reset is keyed on `agent.id` alone so a run starting (`currentAgentId` change) does not wipe streamed socket output.

Closes #3294

## Test plan

- `cd server && npm test` — 1104 files pass. 54 files fail, of which **52 are the pre-existing baseline on `main` in this environment** (Postgres-backed suites that cannot reach a test DB here — verified by running the same suite on `main`, which fails 52 files / 1311 tests). The 2 extra are `lib/dataRoot.test.js` and `services/installState.test.js`, which assert install-root path resolution against the code checkout and fail purely because this branch was built in a git worktree nested under `data/`; they touch nothing in this diff.
- `cd client && npm test` — 516 files / 5912 tests, all pass.
- New coverage: bare-array-default and bounded-envelope tests for both `GET /api/tracks` and `GET /api/media/collections`; two `OutputTab` tests covering the refresh-after-switch response and the clear-on-switch behaviour.
- `npx eslint` clean on all changed client files.


atomantic:
Review notes (codex, gpt-5.6-terra, 3 rounds — all findings applied):

- **Round 1** flagged that guarding `OutputTab` by agent id alone still let two fetches for the *same* agent resolve out of order, and that `RunsTab`'s index-keyed `expanded` map survived an agent switch. Fixed in 6596c662e with a monotonic request-sequence guard (covers mount, run change, and the refresh button) plus an `expanded` reset. A mutation check confirms the guard is load-bearing: removing it fails all three staleness tests.
- **Round 2** flagged that hoisting `getActiveApps()` widened the blast radius of a failed read — one failure would clear *every* app-targeted request in the cycle as "unknown app" rather than just one. Fixed in 89324a3c0 using `null` as a failure sentinel distinct from a legitimately empty list, per the sentinel-vs-empty rule in CLAUDE.md; a failed read now leaves the requests queued for the next cycle.
- The same finding's other half — that the hoisted read could go stale if the registry mutates mid-loop — was **not** actioned. The issue explicitly determined the registry cannot change mid-loop, and `getActiveApps()` is backed by a 2s cache, so the previous per-request call was already serving a snapshot up to 2s old; the hoist doesn't change the freshness contract.
- **Round 3** found only a stale comment in the pre-existing test, corrected in 745ee463d.

Full server suite fails 54 files here, but 52 of those are the pre-existing baseline on `main` in this environment (Postgres-backed suites with no reachable test DB — verified by running the same suite on `main`). The other 2, `lib/dataRoot.test.js` and `services/installState.test.js`, assert install-root path resolution and fail only because this branch was built in a worktree nested under `data/`; neither is touched by this diff.

Issue #3372, opened 2026-08-02
server suite fails two tests whenever it runs from inside a CoS/claim worktree (dataRoot, installState)


atomantic:
## Problem

A full `NODE_ENV=test npx vitest run` from inside a git worktree under `data/cos/worktrees/…` — i.e. every CoS agent and every `/do:next` claim — fails two tests that pass in the primary checkout:

- `server/lib/dataRoot.test.js` → "resolves a relative PORTOS_DATA_ROOT to an absolute path"
- `server/services/installState.test.js` → "reads the ledger from the data install root by default, not the code checkout"

Verified pre-existing on clean `origin/main` (observed while working #3359: the failures persist with the branch diff stashed, and both pass when the same suite runs in the primary tree).

## Cause

`resolveInstallRoot()` deliberately refuses a pinned data root whose path looks like a worktree checkout (leak safety, #1947). The dataRoot test pins a **relative** value:

```js
process.env[DATA_ROOT_ENV] = 'some/rel';
expect(resolveInstallRoot(REAL_ROOT)).toBe(join(process.cwd(), 'some', 'rel'));
```

Inside a worktree, `process.cwd()` is `<repo>/data/cos/worktrees/<name>/server`, so the resolved absolute pin contains `data/cos/worktrees/` and trips the worktree detector. The resolver correctly refuses and returns the fallback, and the assertion fails. `installState.test.js` fails through the same path for its default `rootDir`.

The production behavior is right — the *test* smuggles the ambient cwd into its expectation.

## Decision

Fix the tests, not the resolver. Make the relative-pin case independent of where the suite runs: stub `process.cwd()` to a fixed non-worktree path (e.g. `/Users/me/PortOS`, matching the file's existing `REAL_ROOT` fixture style) for the duration of that assertion, and do the same for the `installState` default-root case. Do not relax `isWorktreeRoot()` — refusing a worktree-looking pin is the #1947 protection.

## Why it matters

Agents run the suite from a worktree by design, so every agent's local full-suite run currently reports two failures it must triage and dismiss by hand. That erodes trust in "the suite is green" exactly where an agent is deciding whether its own change broke something. CI never sees it (CI runs from a normal checkout), so it will not self-correct.

## Acceptance criteria

- `NODE_ENV=test npx vitest run lib/dataRoot.test.js services/installState.test.js` passes from a `data/cos/worktrees/<name>` checkout **and** from the primary checkout.
[...]

Issue #3368, opened 2026-08-02
Direct-CLI and TUI agent failures never record a task resume pointer, so every retry restarts from scratch


atomantic:
## Problem

When a CoS agent fails, `cleanupAgentWorktree` deliberately PRESERVES a branch that has commits (`preserveBranchWithCommits`, `server/services/agentWorktreeCleanup.js`) so the task's retry can resume from it rather than redo the work. Turning that preserved branch into an actual resume requires a *pointer* on the task record — written by `recordTaskResumePointer`.

Only ONE of the three spawn modes writes it. `runAgentCompletionCleanup` (`server/services/agentCompletionCleanup.js`) records the pointer for runner-mode completions; the direct-CLI (`agentCliSpawning.js`) and TUI (`agentTuiSpawning.js`) close/finish handlers call `cleanupAgentWorktree` and stop there. So a failed direct-CLI or TUI run leaves a branch full of committed work on disk that nothing will ever point a retry at — the retry starts clean and redoes it.

## How this surfaces

Most visibly with the PR-claim verification added in #3358: a run that committed, pushed, and simply failed to open its pull request is now recorded `pr-missing` and retried. On runner mode the retry adopts the branch and just opens the PR. On direct-CLI/TUI the retry starts from scratch and redoes the whole task — with the previous branch still sitting there.

It is NOT specific to that failure mode, though: any failed direct-CLI/TUI run with commits has always had this gap. #3358 only made it easier to hit.

## Work

Mirror the runner-mode block in both spawn paths' cleanup `finally`:

1. After `cleanupWorktreeFn(...)`, when the run did NOT succeed and the task has an id, re-read the persisted task and record the pointer only if its status is `pending` (a `blocked` task that exhausted its retry budget must not get dead metadata — `updateTask` would strip it on the next terminal write).
2. `recordTaskResumePointer({ task, agentId, agentMetadata })` needs the agent's persisted metadata for the worktree fields; both spawn paths already have `agentId` and can read the record.
3. The three copies of this block are then identical — extract one helper (e.g. `recordResumePointerIfRetrying({ agentId, task, success })` beside `recordTaskResumePointer`) and call it from all three rather than triplicating it.

## Acceptance criteria

- A failed direct-CLI run whose branch survived cleanup leaves a resume pointer on its task, and the retry adopts the branch instead of starting clean.
- Same for a failed TUI run.
- A run whose task is already `blocked` (retry budget exhausted) gets NO pointer.
[...]

PR #3375, opened 2026-08-02
fix: hold a failed task non-spawnable until its resume pointer is written

### LINKED ISSUE #3373: A failed task becomes spawnable before its resume pointer is written, so a fast retry still starts from scratch
## Problem

A failed task is flipped back to `pending` *before* the resume pointer that lets its retry adopt the previous run's branch is written, so a fast dequeue can claim the retry with no pointer and start from scratch — the exact failure #3368 exists to close.

The ordering, for every spawn mode:

1. `finalizeAgent` (`server/services/agentFinalization.js:535`) applies `resolveFailedTaskUpdate` — a non-actionable failure under the retry budget writes `status: 'pending'`.
2. `completeAgent` emits `agent:completed`, whose listener (`server/services/cos.js`, `init()`) does `setImmediate(() => refillPerpetualForCompletedAgent(...).then(() => dequeueNextTask({ ignoreTaskId })))`.
3. Only *after* `finalizeAgent` returns does the caller run `cleanupWorktreeFn(...)` (git merge/push/PR probes — hundreds of ms to seconds) and then `recordResumePointerIfRetrying`.

`ignoreTaskId` does not close this: it is threaded onto the dequeue `ctx` purely for generator/in-flight tallies (`quotaBurn`, `perpetualWork`, `cosTaskGenerator`, `cosTaskStore`) — the priority tiers themselves never consult it, so the just-failed task is a normal `pending` candidate the moment step 1 lands.

Two ways it goes wrong:

- The retry is spawned during step 3 and reads task metadata with no `existingBranch`, so `agentWorkspacePrep` builds a clean worktree and redoes work that is sitting on the preserved branch.
- Worse, the retry then holds the task at `in_progress`, so `recordResumePointerIfRetrying`'s own status gate (correctly) declines to write — the pointer is silently dropped and never retried.

This is not new to #3368: runner mode has had this shape since the pointer was introduced. #3368 only extended the same ordering to direct-CLI and TUI, so all three modes now share it.

## Why it is not just "move the write earlier"

The pointer *cannot* be resolved before cleanup: `resolveResumePointer` asks whether the branch survived and still holds commits, and that is exactly what `cleanupAgentWorktree` decides (`preserveBranchWithCommits`, or an aborted removal on a dirty tree). Writing it before cleanup would point retries at branches that were about to be merged and deleted.

## Work

Keep the task non-spawnable until cleanup and the pointer write have both settled. The suggested shape, mirroring what `handleOrphanedTask` (`agentManagement.js`) already does by folding the patch into its own retry write:

1. Have `resolveFailedTaskUpdate`'s retry branch write an intermediate non-spawnable status (or a `retryPendingCleanup` metadata flag the dequeue tiers skip on) instead of going straight to `pending`.
2. After `cleanupWorktreeFn` + `recordResumePointerIfRetrying`, do the flip to `pending` in one write that carries the resume metadata — so the task becomes spawnable and pointed in the same update.
3. Make that final flip unconditional-on-crash-safe: if the process dies between steps 1 and 2, the boot-time orphan sweep (`handleOrphanedTask`) must recognize the intermediate state and complete the transition, or the task strands forever. That sweep already resolves resume pointers, so it is the natural place.

The `blocked` case needs no change — a task that exhausted its retry budget is not spawnable and gets no pointer either way.

## Acceptance criteria

- A failed run's retry is never dequeued between the failure verdict and the resume-pointer write.
- A retry that does spawn carries `existingBranch` / `resumeWorktreePath` when the previous run left commits behind.
- A process killed mid-transition leaves a task the orphan sweep can finish, not a permanently non-spawnable one.
- A `blocked` task still gets no pointer and no intermediate state.
- Tests cover: a dequeue attempted during the cleanup window does not claim the task; the pointer and the `pending` flip land in one write; the orphan sweep recovers the intermediate state.

Found by the codex review pass on #3370 (the PR for #3368).

atomantic:
## Summary

A retryable agent failure flipped its task straight back to `pending` inside `finalizeAgent`, while the resume pointer that lets the retry adopt the branch cleanup preserved can only be resolved *after* `cleanupAgentWorktree` decides what survived — hundreds of milliseconds to seconds later. The dequeue that `agent:completed` schedules could claim the retry in that window and start it on a clean worktree, redoing work sitting on the preserved branch; worse, the retry then held the task `in_progress`, so the pointer write's own status gate (correctly) declined and the pointer was dropped for good. `ignoreTaskId` never closed this — it is only threaded onto generator/in-flight tallies, never consulted by the priority tiers.

**The hold.** A retryable failure now leaves the task `in_progress` carrying a `retryPendingCleanup` marker (new `server/lib/taskRetryHold.js`). No dequeue tier looks at `in_progress` (every tier selects through `grouped.pending` / `getAutoApprovedTasks`), and `spawnAgentForTask` rejects a non-`pending` task — so the hold is enforced by the existing selection surface rather than by a new flag each tier has to remember. It also keeps the federation claim, since `updateTask` only releases the lease when a status other than `in_progress` is written.

**The release.** `recordResumePointerIfRetrying` is renamed `releaseRetryHold` and now flips the task to `pending` **with** the resume metadata and the cleared marker in ONE `updateTask` — never spawnable-then-pointed. It still runs after cleanup in all three spawn modes; the runner path's call moved into a `finally` because a JIRA-branch task skips the whole worktree block and would never have been released (and a throw from the JIRA/pipeline/CD steps would have stranded it). The no-hold fallback keeps the pre-#3368 behavior (pointer-only write) for a task already read as plain `pending`.

**Crash safety.** The marker is persisted metadata, so a process killed mid-transition leaves evidence: `handleOrphanedTask` recognizes a held task and finishes the transition (resolve pointer → `pending` → drop marker) without charging orphan-retry budget for a decision `finalizeAgent` already made, and before the commit check — a failed run that committed is exactly what the pointer is for. `resetOrphanedTasks` recovers a hold immediately on the boot pass (nothing can be mid-cleanup there) and only after a 10-minute liveness grace on the 15-minute periodic pass, so a slow merge/push/PR probe is never mistaken for a dead process.

**Wake-up.** An `in_progress → pending` flip now emits `tasks:changed` action `requeued`, which `cos.init` turns into a dequeue — the release lands long after the completion dequeue ran, so without it the retry would idle until an unrelated timer. The orphan sweep's requeue gets the same wake for free.

`blocked` is untouched: a task that exhausted its retry budget is never held and never pointed.

**Drive-by (same mechanism).** `resumePointerMetadata`'s *clear* returned `null`s, which `updateTask` preserves and `TASKS.md` serializes as the literal string `"null"` — read back as a truthy `existingBranch` naming a branch that does not exist, sending the next attempt down the resume path for nothing. It now clears with `undefined`, which `updateTask` deletes from the merged metadata.

No on-disk migration: the marker is additive task metadata that only ever exists between a failure verdict and its release, and the sweep clears any left by an older/crashed process.

## Test plan

`cd server && NODE_ENV=test npx vitest run` → **24170 passed, 2 failed, 211 skipped (1195 files)**. The two failures are the known worktree-only ones being fixed in #3372 (`lib/dataRoot.test.js` relative-`PORTOS_DATA_ROOT`, `services/installState.test.js` ledger root); both are unrelated to this diff and fail identically on `origin/main` in a worktree.

New/updated coverage:
- `lib/taskRetryHold.test.js` — marker arm/clear (including the `undefined`-not-`null` contract), the markdown `'true'` round-trip, and the stale/grace boundary incl. an undateable hold.
- `services/agentErrorAnalysis.test.js` — the retry verdict is `in_progress` + marker; a held task is excluded by the REAL selection helpers (`groupTasksByStatus().pending`, `getAutoApprovedTasks`, `getAwaitingApprovalTasks`) so no tier can see it; a blocked verdict carries no marker.
- `services/cleanupAgentWorktree.test.js` — the flip + pointer + marker clear land in exactly one `updateTask`; the hold is released even with nothing to resume; a hold that vanished (sweep won the race / retry already spawned) writes nothing; a blocked task is never flipped; a failed release leaves the hold on disk.
- `services/agentManagement.test.js` — the orphan sweep finishes an interrupted transition, charges no orphan budget, does not let the commit check complete a held task, and still skips a user-terminated task.
- `services/agentCompletionCleanup.test.js` — the release fires for a JIRA-branch task (which skips worktree cleanup) and when a cleanup step throws.
- `services/cosTaskStore.test.js` — `requeued` action on `in_progress → pending`; `undefined` metadata keys are dropped from disk rather than persisted.
- `services/cos.test.js` — the sweep consults the hold before `handleOrphanedTask`, and the `tasks:changed` listener dequeues on `requeued`.

Closes #3373


atomantic:
Codex review round 1 raised four findings; three are addressed in 14d991f59 (owner-scoped hold marker + status-gated release + terminal-write strip; spawn-path refusal of a held task; boot recovery bypassing the 60s recently-completed grace).

The fourth — `cosTaskMerge` ranking `in_progress` above `pending` regardless of recency, so a peer's stale snapshot can revert a requeue and drop its resume pointer — is pre-existing and out of scope here: it applies identically to the orphan sweep's requeue and to the pre-#3373 failure requeue on `main` today, and fixing it means changing federation merge semantics. Filed as #3376 with the decision (extend the existing non-monotonic `challenged` exemption to the `in_progress`/`pending` pair) ready to work.

Issue #2441, opened 2026-07-12
CoS: let a sub-agent challenge a wrong reviewer rejection (bounded, escalates to user) instead of silently fixing or blocking


atomantic:
## Summary

When a CoS review-loop reviewer **rejects** a sub-agent's work, the agent has no way to *dispute* the verdict — it can only silently "fix" whatever was flagged or the task gets `blocked`. There is no `challenged` status and no two-way worker↔reviewer channel. A wrong rejection (common with noisy local reviewers) just burns budget on a "fix" that wasn't needed, or blocks good work.

Idea surfaced by the sibling **Conductor** comparison ([source](https://github.com/silasfelinus/conductor/blob/main/docs/2026-07-11-portos-cos-learnings.md)): Conductor's *challenge protocol* lets the Worker set `status: challenged`, append its full case to a `TALKBACK` log, and the Reviewer responds — if the challenge holds, work returns to `ready`; if not, it escalates to human arbitration. Critically, **challenges share the same retry budget** so an agent can't loop forever disputing.

## Why this matters for PortOS

Local/CLI reviewers are known to emit wrong or empty verdicts (see e.g. the recurring "ollama reviewer yields no verdict / slow coder GGUF times out" pattern). Today the only outcomes are *fix* or *block* — there's no path for the worker to say "this rejection is wrong, here's why" and get a human tiebreak. This turns a noisy reviewer into wasted spend or a false block.

Single-user trust model note: "escalate to human" = surface to the one PortOS user for a decision, not multi-actor arbitration.

## Current state

- `server/services/cosTaskMerge.js:54` `STATUS_RANK = { completed, blocked, in_progress, pending }` — full status vocabulary; no `challenged`.
- `server/lib/cosValidation.js` — reviewers are a CSV flag threaded into the claim prompt (`normalizeReviewers`, `buildReviewWithArgs`); review is an **external one-way handoff** to the slashdo/`do:better` loop.
- `server/services/cosAgents.js:316` `completeAgent` records a one-way `validationPassed` verdict — no rebuttal path.
- `server/routes/cosTaskRoutes.js:44` — `reviewLoop` flag + reviewers wiring.

## Proposed work

1. Add a `challenged` task status (with `STATUS_RANK` + merge/federation coverage — see the "new schema version trips 3 coverage guards" checklist).
2. When a reviewer rejects, allow the worker one bounded challenge: record the worker's case (reason + evidence) on the task, **consuming a retry-budget slot** so it can't loop.
[...]

Issue #1563, opened 2026-06-23
Federate CoS agent history + tasks with instance ownership/claim metadata


atomantic:
Part of #1561 (full-sync federated node pairs). This is the most complex sub-issue and the one explicitly called out: mirror CoS agent history **and** add metadata so federated peers never duplicate agent work.

## Current state (audit)

- CoS state lives in `data/cos/state.json` (`agents` map, config, stats) — `server/services/cosState.js`.
- Completed agents archive to `data/cos/agents/<YYYY-MM-DD>/<agentId>/` with `metadata.json`, `output.txt`, `prompt.txt` — `server/services/cosAgents.js:66`.
- Tasks live in markdown (`data/COS-TASKS.md` / `data/TASKS.md`), parsed by `server/lib/taskParser.js`.
- **None of `data/cos/` is in any sync category** — `dataSync.js` has no `cos` path/category; CoS is fully local today.
- Worktrees: `data/cos/worktrees/<agentId>/`, branch `cos/<taskId>/<agentId>` — `server/services/worktreeManager.js:163`. No instance recorded.

## Ownership gap

Instance identity already exists: `getInstanceId()` → `data/instances.json` `self.instanceId` (`server/services/instances.js:163`). But:
- Agent metadata (`server/services/agentLifecycle.js:261` `registerAgent`) records ~37 fields and **none is `instanceId`/`claimedBy`**.
- Task metadata (`taskParser.js`) has no `claimedBy`/`claimedAt`/`instanceId`.
- The duplicate-spawn guard `spawningTasks` Set is **in-memory, single-instance** (`agentLifecycle.js:116`).
- Orphan reset (`server/services/cos.js:481` `resetOrphanedTasks`) only checks local agent state.

## Why this is dangerous to naively sync

If task state were synced without ownership, two peers would each parse the same `pending` task, both `spawnAgentForTask()`, create conflicting worktrees/branches on the same repo, and race the orphan-reset → duplicated work, force-push wars, incoherent retry counts. CoS must NOT be folded into full-sync until claim/lease metadata exists.

## Proposal

**1. Stamp instance provenance**
- Add `metadata.instanceId = await getInstanceId()` at agent registration (`agentLifecycle.js:261`).
- Add `instanceId` to worktree creation metadata (`worktreeManager.js`).

**2. Task claim + lease**
[...]

PR #3377, opened 2026-08-02
fix: resolve an in_progress/pending merge conflict by recency so a stale peer can't revert a requeue

### LINKED ISSUE #3376: A peer's stale in_progress snapshot can revert a newer requeue, discarding its resume pointer
## Problem

`pickContentBase` (`server/services/cosTaskMerge.js`) resolves a same-task conflict by `statusRank` first and only falls back to `updatedAt` when the ranks are equal. `in_progress` outranks `pending`, so a **legitimate, newer `in_progress → pending` transition can be reverted by a peer's stale pre-transition snapshot** — the merge adopts the remote `in_progress` record and discards the local `pending` one along with the metadata that write carried.

Two live transitions have this shape:

- The orphan sweep's requeue (`handleOrphanedTask`, `server/services/agentManagement.js`) — reverting it also discards the `existingBranch` / `resumeWorktreePath` resume pointer it just resolved, so the eventual retry restarts from scratch.
- The retry-hold release added in #3373 (`releaseRetryHold`, `server/services/agentWorktreeCleanup.js`) — same loss, plus the task goes back to reading `in_progress` with a claim whose agent is gone.

The `challenged` case already has an exemption for exactly this reason ("Challenge lifecycle is NOT monotonic (#2441) … a pure status-rank comparison would let a stale `challenged` snapshot permanently revert a newer resolution"). `in_progress → pending` is the same non-monotonic shape and has no exemption.

It self-heals within one sweep interval (the reverted task is `in_progress` with no running agent, so the next `resetOrphanedTasks` requeues it), so the symptom is a retry delayed by up to 15 minutes and possibly restarted clean rather than resumed — not a permanent stall. It only bites a multi-machine (federated) install.

## Decision

Extend the non-monotonic exemption rather than reorder the ranks. When **exactly one** side is `in_progress` and the other is `pending`, decide by `updatedAt` (newest edit wins), falling back to the existing rank comparison when the stamps are equal or absent. This is symmetric — it depends only on the pair — so both peers converge on the same winner, matching the `challenged` exemption's contract. `completed` must still win outright (rule 2, monotonic completion), so the exemption is scoped to the `in_progress`/`pending` pair only.

Claim resolution is separate (`resolveClaim` by lease) and is unaffected.

## Work

- `server/services/cosTaskMerge.js`: add the `in_progress` vs `pending` recency exemption to `pickContentBase`, next to the `challenged` one, with a comment naming both requeue paths.
- `server/services/cosTaskMerge.test.js`: a stale remote `in_progress` snapshot must NOT beat a newer local `pending` requeue (and the reverse direction must pick the same record); an in-flight spawn (`pending` local, newer remote `in_progress`) must still adopt `in_progress`; equal/absent stamps keep today's rank behavior.
- Confirm no `SCHEMA_VERSIONS` bump is needed: the wire shape is unchanged, only the local resolution of a conflict — an older peer resolving the same pair the old way converges on the next sweep exactly as it does today.

## Acceptance criteria

- A newer `in_progress → pending` requeue survives a merge against a peer's stale `in_progress` snapshot, keeping its resume-pointer metadata.
- A genuinely newer `in_progress` (a peer that just spawned) still wins over an older `pending`.
- `completed` still wins outright against either.
- Merging is symmetric: both peers pick the same record for the same pair.

Found by the codex review pass on #3375 (the PR for #3373), where it was reported as "federation can undo a completed hold release". Pre-existing — it applies identically to the orphan sweep's requeue on `main` today.

atomantic:
## Summary

`pickContentBase` (`server/services/cosTaskMerge.js`) resolves a same-task federation conflict by lifecycle rank, which assumes a task only ever moves forward. Two live paths move it backward:

- the orphan sweep's requeue (`handleOrphanedTask`, `agentManagement.js`)
- the retry-hold release (`releaseRetryHold`, `agentWorktreeCleanup.js`, #3373)

Both transition `in_progress → pending`, and the same write records the resume pointer (`existingBranch` / `resumeWorktreePath`). Because `in_progress` outranks `pending`, a peer holding a stale pre-requeue snapshot won on rank alone and discarded the requeue along with that metadata — the eventual retry then restarted from an empty tree instead of resuming the preserved branch. It self-heals within a sweep interval, so the symptom is a delayed, clean-restarted retry on federated installs, not a stall.

This extends the non-monotonic exemption `challenged` already carries (#2441) to the `in_progress`/`pending` pair, but on a **causal** test rather than the plain recency the issue proposed. Deciding the pair by `updatedAt` alone would also hand the win to an ordinary content edit landing on a peer's stale `pending` copy — reverting a genuinely running task and inviting a second agent onto in-flight work, which is worse than the delayed retry being fixed. So:

- `cosTaskStore.updateTask` stamps `lastRequeuedAt` on the `in_progress → pending` transition and clears it on the next spawn.
- `server/lib/taskRequeue.js#isPostSpawnRequeue` answers whether that stamp postdates the `lastSpawnedAt` the other side is reporting.
- The exemption fires only then. Every other pairing — an edit, a stale stamp from a previous cycle, a peer too old to stamp — falls through to the lifecycle rank, exactly as today.

Two supporting changes, rolled in because the fix depends on them:

- **The retired run's lease is no longer re-imposed.** `cosTaskStore` releases the claim in the very write that leaves `in_progress` (#1563), so a stale lease can't block the retry for a lease window — but rule 3 resolves the claim independently of content and re-attached it once the requeue won the content. The counterpart's claim is now excluded when the requeue wins; a claim held by the winning `pending` side is still honored, because dispatch takes the lease *before* flipping the status (`agentLifecycle.js`), so `pending` + live claim is a legitimate in-flight shape a peer must keep seeing.
- **`resolveClaim`'s both-live tiebreak is now total.** Same owner and same `leaseExpiresAt` with a differing `claimedAt` previously left each peer holding its own triple forever.

`completed` still wins outright (rule 2, monotonic completion) — the exemption requires that exact status pair. The rule depends only on the pair, so it is symmetric and both peers pick the same record and the same claim.

No `SCHEMA_VERSIONS` bump: the wire shape is unchanged (`lastRequeuedAt` is additive free-form metadata, like the resume pointer beside it) and only the local resolution of a conflict changes. An older peer neither stamps nor reads it, so it resolves the pair the old way and converges on the next sweep exactly as it does today.

## Test plan

- `server/services/cosTaskMerge.test.js` — new `in_progress → pending requeue merge (#3376)` block:
  - a requeue postdating the peer's spawn survives its stale `in_progress` snapshot, keeping `existingBranch` / `resumeWorktreePath` (both merge directions)
  - a **newer edit** on a stale `pending` copy does NOT revert a running task or strip its lease
  - a requeue stamp from a **previous cycle** does not beat a later spawn
  - a genuinely newer `in_progress` still wins over an older `pending`
  - `completed` wins outright over a newer `pending` *and* a newer `in_progress`
  - the retired run's lease is not re-imposed; a live claim on the winning `pending` side is kept
  - an un-stamped (older) peer keeps today's rank behavior
  - two live claims with the same owner and expiry converge on one `claimedAt`
- `server/services/cosTaskStore.test.js` — `lastRequeuedAt` is stamped only on `in_progress → pending`, carried forward by a later edit, and cleared by the next spawn.
- Verified the merge guard is load-bearing: disabling it fails the requeue test and only that test.
- `cd server && NODE_ENV=test npx vitest run` → **1171 files, 24192 passed**, 211 skipped (DB-backed suites, no `portos_test` here).

Review loop: codex `gpt-5.6-terra`, 3 rounds — dispositions in a PR comment below.

Closes #3376


atomantic:
Review loop (codex, 3 rounds) — dispositions:

**Round 1 — fixed.** Rule 3 re-attached the losing `in_progress` snapshot's live lease onto the adopted `pending` record, re-imposing the very lease the requeue write releases (#1563) and blocking the retry for a full lease window. The counterpart's claim is now excluded when the requeue wins; a claim held by the winning `pending` side is still honored (dispatch takes the lease before flipping the status, so `pending` + live claim is a legitimate in-flight shape).

**Round 2 — fixed, and it changed the design.** The rule as specified in the issue (decide the pair by `updatedAt`) also handed the win to an ordinary content edit landing on a peer's stale `pending` copy — reverting a genuinely running task and, with the round-1 carve-out, stripping its live lease. That invites a second agent onto in-flight work, which is worse than the delayed retry this fixes. The exemption is now causal rather than recent: `cosTaskStore` stamps `lastRequeuedAt` on the `in_progress → pending` transition and clears it on the next spawn, and the exemption fires only when that stamp postdates the `lastSpawnedAt` the other side reports (`server/lib/taskRequeue.js#isPostSpawnRequeue`). Everything else falls through to the lifecycle rank, unchanged — including for a peer too old to stamp. Rolled in alongside: `resolveClaim`'s both-live tiebreak is now total (same owner + same expiry with differing `claimedAt` previously left each peer holding its own triple forever).

**Round 3 — won't fix, reasoning recorded in `server/lib/taskRequeue.js`.** The remaining finding is that `lastRequeuedAt > lastSpawnedAt` compares wall-clock stamps and so is skew-sensitive across peers. The comparison that matters is not cross-peer: a machine only requeues a run of its own (`releaseRetryHold` is scoped to this instance's agent; the orphan sweep leaves a task alone while a peer holds a live lease, via `isHeldByOther`), so both stamps are written by the same clock and the peer's copy is a replica. Replacing wall-clock with a monotonic run generation would mean redesigning the module's LWW foundation (`updatedAt`, `leaseExpiresAt`), which is well outside this issue. The strict `>` is also deliberate — `>=` would let a pre-spawn `pending` copy with an equal stamp win — and the fallback on any indeterminate pairing is today's behavior plus the next sweep's requeue.

PR #3411, opened 2026-08-03
fix: merge storyboard scene writes against fresh state so sibling renders don't clobber job ids (#3400)

### LINKED ISSUE #3400: Guard storyboards.js scene enqueues against read-modify-write clobbering
## Problem
`server/services/pipeline/storyboards.js` (~33-117, 126-191): enqueueStoryboardSceneVideo / enqueueStoryboardShotStartFrame read `scenes` at entry, enqueue a job, then write back the entire pre-read array via updateStage. Two near-simultaneous enqueues for different scenes on the same issue can clobber each other's freshly-written sceneVideoJobId/startFrameJobId (jobs keep running untracked). Marked UNCERTAIN in audit — verify updateStage's internal serialization first.

## Decision
If updateStage doesn't already merge per-issue serially against fresh state, change these two paths to pass a targeted per-index patch or mutator function so the write merges against the freshest persisted scenes array.

## Acceptance
- Concurrent enqueues for different scenes both retain their job ids (add a unit test simulating interleaved calls); no change to single-call behavior

atomantic:
## Summary

The audit finding was marked **UNCERTAIN** pending "verify `updateStage`'s internal serialization first." **The race is real** — `updateStage` serializes the writes but does not merge them, so serialization never helped here.

`updateStage(issueId, stageId, patch)` is just `updateStageWithLatest(issueId, stageId, () => patch)`, and that does queue on the per-series write tail. But all three storyboard write paths read `stages.storyboards.scenes` at entry, mutated a copy, and passed the **whole pre-read array** as the patch — a value computed *outside* the lock. The tail then shallow-merges that stale array over the freshest stage, so the second of two concurrent enqueues reverts the first scene's freshly-stamped job id. The media job keeps running with nothing left pointing at it, and the filename hook (which only attaches a completed render when the target slot already carries the job id) silently drops the result.

Fixed by routing all three through a shared `patchStoryboardScene(issueId, index, mutate)` that hands `updateStageWithLatest` a **mutator**, keeping the read-modify-write of `scenes` inside the write region so only the targeted index is replaced. This mirrors the existing `persistComicPageSlot` treatment in `comicPages.js` and `renderComicCoverLike` in `covers.js` — the same fix, already applied to the comic-page surface.

Paths changed:
- `enqueueStoryboardSceneVideo` — `scenes[idx].sceneVideoJobId`
- `enqueueStoryboardShotStartFrame` — `scenes[sIdx].shots[tIdx].startFrameJobId`
- `refineStoryboardScenePrompt` — `scenes[idx].description`. **Beyond the issue's stated scope**, included deliberately: it is the identical bug in the same file, and with an LLM round-trip sitting between the read and the write it has by far the widest clobber window of the three. Leaving it would have left the fix half-applied.

Single-call behavior is unchanged. One deliberate behavior change: a scene or shot that disappeared between the caller's read and the write now 404s (`PIPELINE_SCENE_NOT_FOUND` / `PIPELINE_SHOT_NOT_FOUND`) instead of resurrecting itself into the persisted array — validating against the stale snapshot was the only reason that ever "worked."

Per the trust model this is a same-process re-entrancy race from one user's pipeline run (two render buttons, or a fan-out over scenes), not a cross-human race — squarely the "serializing two write paths that mutate the same record" case CLAUDE.md calls expected.

## Test plan

- `cd server && NODE_ENV=test npm test` — **24297 passed, 211 skipped, 0 failed.** (A first run also flagged `dataManager.categories.test.js` with `ENOENT … server/test-data`; it passes in isolation and on re-run — a parallel-run race over a gitignored runtime dir in a fresh worktree, unrelated to this change.)
- New `describe('concurrent storyboard scene enqueues (#3400)')` in `server/services/pipeline/visualStages.test.js` installs a genuinely stateful serialized write tail (one shared persisted stage, one mutex, `computeFn` evaluated inside the lock) and drives enqueues concurrently via `Promise.all`:
  - two scene-video enqueues for different scenes → both `sceneVideoJobId`s survive
  - two shot start-frame enqueues on different scenes → both `startFrameJobId`s survive
  - a scene-video enqueue concurrent with a sibling-scene prompt refine → neither reverts the other
  - a scene that vanished before the write landed → 404 instead of resurrection
- **Falsification check:** reverted `storyboards.js` to the pre-fix version and re-ran — all 4 new tests fail with exactly the clobber symptom (`expected undefined to be 'job-scene-0'`), plus the 3 existing happy-path assertions that now pin the mutator path. Restored, all 126 pass.
- Test-harness note: the shared `updateStageWithLatest` mock now reads from a per-test `persistedStages` (reset in `beforeEach`) so the outer `getIssue` read and the write tail agree on state, and `beforeEach` restores mock *implementations* rather than only clearing call logs — otherwise the stateful tail installed by these tests would leak into the rest of this 1500-line file.

Closes #3400


atomantic:
### Review gate — codex `gpt-5.6-terra`, 1 round

Two findings, both legitimate, both **deferred to #3413** rather than applied here. Reasoning:

**1. Index-based targeting can retarget onto the wrong scene under a concurrent reorder.** Correct, and the deeper issue — but storyboard scenes carry no durable id at all (`issuesShared.js:395` is a bare `scenes.slice(0, 200)`), and the whole surface is index-addressed: the route param, the job owner string `…:storyboards:scene<idx>`, and `storyboardsFilenameHook.js`'s completion routing. Fixing it means stamping ids, a migration for every existing install's persisted scenes, and a legacy-owner fallback for in-flight jobs. Out of scope for a targeted race fix.

   Importantly this is **not a regression** from this PR: pre-fix, the same reorder was clobbered wholesale by the stale array being written back. Index-targeting within the *fresh* array is strictly better than what it replaces.

**2. A media job is orphaned when the locked write rejects.** Also correct — `enqueueJob` runs before `patchStoryboardScene`, so a vanished scene now 404s with the job already queued. This PR does newly expose that path (pre-fix the write couldn't 404; it silently resurrected the deleted scene into the array instead — worse). `comicPages.js#persistComicPageSlot` and `covers.js#renderComicCoverLike` have the identical enqueue-then-persist shape, so compensating only here would leave the three render surfaces inconsistent. #3413 covers all three together, using the exported `cancelJob` (`mediaJobQueue/index.js:965`).

No findings applied; no code change since the review. Worktree verified clean afterward (codex made no edits). Full server suite green at 24297 passed.

Issue #3517, opened 2026-08-05
Scheduling controls latch in busy loading state on API failure in useGoalDetail


atomantic:
### Problem
In [`useGoalDetail.js`](file:///Users/adameivy/github.com/atomantic/PortOS/client/src/hooks/useGoalDetail.js#L148-L167) (lines 148–167), `handleSchedule`, `handleRemoveSchedule`, and `handleReschedule` set `setSchedulingBusy(true)` before calling `api.scheduleGoalTimeBlocks(goal.id)`, `api.removeGoalSchedule(goal.id)`, or `api.rescheduleGoalTimeBlocks(goal.id)`. These functions do not wrap the async calls in `try ... finally` blocks. If an API request fails, `request()` throws an error, causing the unhandled rejection to abort execution before `setSchedulingBusy(false)` is reached.

### Impact
When time-block scheduling operations fail (e.g. server error 500 or network glitch), the scheduling controls in `GoalPlanSection.jsx` remain permanently disabled in the loading state ("Scheduling...") with no way for the user to re-arm or retry without reloading the entire page.

### Fix
Wrap the API calls inside `try ... finally` blocks in `handleSchedule`, `handleRemoveSchedule`, and `handleReschedule` in `client/src/hooks/useGoalDetail.js` to ensure `setSchedulingBusy(false)` is always executed even on error.

### Acceptance criteria
- [ ] `handleSchedule` uses `try ... finally` to call `setSchedulingBusy(false)`.
- [ ] `handleRemoveSchedule` uses `try ... finally` to call `setSchedulingBusy(false)`.
- [ ] `handleReschedule` uses `try ... finally` to call `setSchedulingBusy(false)`.
- [ ] If a scheduling API call fails, the schedule buttons re-enable so the user can retry without page reload.

PR #3551, opened 2026-08-05
fix: don't open the goal check-ins list or refresh when the check-in request fails

### LINKED ISSUE #3518: Goal Check-In failure expands empty list and triggers refresh without error feedback
### Problem
In [`useGoalDetail.js`](file:///Users/adameivy/github.com/atomantic/PortOS/client/src/hooks/useGoalDetail.js#L140-L146) (lines 140–146), `handleCheckIn` calls `await api.checkInGoal(goal.id).catch(() => null)`. When the API call fails, `.catch(() => null)` returns `null`, but `handleCheckIn` proceeds unconditionally to call `setCheckInsOpen(true)` and `onRefresh()`.

### Impact
When triggering a Goal Check-In fails, the check-ins collapsible accordion expands open showing no new check-in entry, while `onRefresh()` is called pointlessly. The user sees an empty panel open with no indication that the check-in request failed.

### Fix
In `client/src/hooks/useGoalDetail.js`, store the result of `api.checkInGoal(goal.id)` and only execute `setCheckInsOpen(true)` and `onRefresh()` if the response is non-null.

### Acceptance criteria
- [ ] `handleCheckIn` verifies `res != null` before setting `checkInsOpen` to `true` and calling `onRefresh()`.
- [ ] Failed check-in attempts do not force open the check-ins drawer or trigger a list refresh.

atomantic:
## Summary

`handleCheckIn` in `client/src/hooks/useGoalDetail.js` fired `api.checkInGoal(goal.id).catch(() => null)` and then called `setCheckInsOpen(true)` + `onRefresh()` unconditionally. A failed check-in therefore expanded the check-ins accordion onto the unchanged list and triggered a pointless refresh — the user saw a panel open with no new entry and no indication the request had failed.

The response is now the signal: the accordion opens and the panel refreshes only when the server actually returns the created check-in (the route responds `201` with the check-in object). A failure leaves the accordion closed and skips the refresh.

Two supporting details:

- The busy-flag reset moved into a `try/finally`, mirroring the `runSchedulingAction` fix from #3517 in the same file. A bare `.catch()` only covers a rejected promise, so a call that threw before returning one would have skipped the reset and latched the button on "Checking in..." until a page reload.
- No custom toast was added. `checkInGoal` does not pass `{ silent: true }`, so `request()` in `apiCore.js` already toasts the failure — per the repo's "custom catch ⇒ `silent: true`" rule this stays a single-layer error UI rather than double-toasting.

## Test plan

- Extended `client/src/hooks/useGoalDetail.test.jsx` with a `handleCheckIn` suite covering the success path (opens the list, refreshes once), the in-flight busy flag, a rejected request, a `null` response body, a synchronous throw from the api call, no unhandled rejection reaching the caller, and retry-after-failure.
- Bypass probe: reverting the hook change makes 4 of the new tests fail, so they are testing the fix rather than restating it.
- `cd client && npx vitest run src/hooks/useGoalDetail.test.jsx` — 23 passed.
- `cd client && npx vitest run src/hooks src/components/goals` — 83 files, 764 tests passed.
- `cd client && npx biome lint --error-on-warnings src/hooks/useGoalDetail.js src/hooks/useGoalDetail.test.jsx` — clean.

Closes #3518


atomantic (commented):
### Code Review Summary

**PR #3551: fix: don't open the goal check-ins list or refresh when the check-in request fails**

#### Overview
This PR addresses issue #3518, preventing the goal check-ins accordion from expanding and preventing an unnecessary panel refresh when a check-in request fails or returns no data.

#### Key Improvements Verified
1. **Conditional Accordion Expansion & Refresh**:
   - `handleCheckIn` now checks `if (checkIn == null) return;` before calling `setCheckInsOpen(true)` and `onRefresh()`.
2. **Robust Error Handling (`try/finally`)**:
   - `setCheckingIn(false)` is guaranteed to run inside `finally`, ensuring the loading/busy state is reliably reset even if `api.checkInGoal` throws synchronously or rejects asynchronously.
   - Avoids double-toasting since API failure notifications are already handled centrally by `request()`.
3. **Comprehensive Unit Test Coverage**:
   - Added 7 targeted unit tests in `useGoalDetail.test.jsx` testing:
     - Normal success (opens list and refreshes).
     - Busy state toggle while in flight.
     - Rejected promise handling (leaves list closed, skips refresh, resets busy flag).
     - Null response handling.
     - Synchronous error throwing before promise creation.
     - No unhandled promise rejections.
     - Retry capability after failure.

#### Verdict
LGTM! 🚀 The code is clean, robust, and exceptionally well-tested.

PR #3584, opened 2026-08-05
fix: split the WorkEditor header into two compact mobile rows so the prose textarea clears the fold (#3568)

### LINKED ISSUE #3568: Primary prose writing area pushed below fold by multi-row toolbar wrapping in WorkEditor
### Problem
`client/src/components/writers-room/WorkEditor.jsx:723-806` renders the top header/toolbar with `flex flex-wrap items-center gap-2`. The toolbar contains the title text input (`min-w-[180px]`), status selector, view mode toggle group (Edit, Read, Review), Save button, Snapshot button, and Work menu button.

### Impact
On ~375px wide mobile viewports, the flex-wrapping header controls stack into 4-5 vertical rows consuming ~250px-300px of top viewport space. On standard mobile screens (e.g. 667px height), this pushes the primary prose editor text area below the fold before the mobile keyboard even appears.

### Fix
Refactor the header layout in `client/src/components/writers-room/WorkEditor.jsx` for mobile viewports (`<sm`): collapse secondary header items (status select, snapshot, view mode toggles) into a compact sub-bar or the overflow work menu so the header fits within 1-2 compact rows on mobile.

### Acceptance criteria
- [ ] On ~375px mobile viewports, the `WorkEditor.jsx` header takes no more than 2 compact rows.
- [ ] The primary prose editor textarea is immediately visible above the fold on mobile screens.

atomantic:
## Summary

The Writers Room work header laid every control out in one `flex-wrap` row — title input (`min-w-[180px]`), status select, Edit/Read/Review toggle group, Save, Snapshot, Work menu. At ~375px that wrapped into 4-5 rows consuming ~250-300px, so on a 667px-tall phone the prose textarea started below the fold before the keyboard opened.

The header is now at most two compact rows on a phone:

- **Row 1** — title, Save, Snapshot, Work menu.
- **Row 2** — a full-width sub-bar (`w-full order-last`) holding the status select and the view-mode toggle.

Snapshot and the three view-mode buttons drop their text labels under `sm` (`hidden sm:inline`) so both rows stay single-line at 375px; each carries an explicit `aria-label` so the icon-only form keeps its accessible name.

Nothing moved into the overflow menu and nothing is duplicated per breakpoint: the sub-bar is `sm:contents`, so at `sm+` it generates no box and its two children are direct flex items of the header row again. The markup order *is* the desktop order (title · status · view mode · Save · Snapshot · menu), so no control carries an `order-*` class and desktop tab order still follows what is on screen — an earlier pass restored the desktop sequence with `sm:order-*` instead, which decoupled focus order from the layout; the second commit removes that.

Every control in the two phone rows also clears the 44px touch floor (`min-h-[44px] sm:min-h-0`), and the three hand-copied view-mode buttons collapse to a map over a new `VIEW_MODES` constant. Only configured breakpoints are used (`sm`); no `xs` variant is involved.

## Test plan

- New `client/src/components/writers-room/WorkEditor.test.jsx` (7 cases) asserts the layout contract jsdom can see: the sub-bar is `w-full order-last sm:contents` and holds status + view mode; the first row is title, Save, Snapshot and the Work menu, none of which carry an `order-*` class; DOM order equals the desktop left-to-right order (via `compareDocumentPosition`); each control renders exactly once (no per-breakpoint duplicates); the icon-only Snapshot and view toggles keep their accessible names with their labels `hidden sm:inline`; the mobile-visible controls carry `min-h-[44px] sm:min-h-0`; and switching to Read view from the regrouped toggle still swaps the surface.
- Bypass-probed the guards (dropping `order-last`, and mutating an order class in the earlier revision) so they fail when the contract is broken rather than passing vacuously.
- `cd client && npx vitest run src/components/writers-room` → 7 files, 42 tests pass (includes the existing `WorkEditor.perf.test.jsx` memo/prop-stability guard).
- Full client suite: `cd client && npx vitest run` → 554 files, 6536 tests pass.
- `npx biome check` clean on both changed files.

Closes #3568


atomantic (commented):
### Review Summary

| Agent | Files Checked | Issues Found | Fixed |
|-------|--------------|-------------|-------|
| Surface Scan (Runtime) | 3 | 0 | 0 |
| Surface Quality | 3 | 0 | 0 |
| Security Audit | 3 | 0 | 0 |
| Cross-File Tracing (State) | 3 | 0 | 0 |
| Cross-File Contract | 3 | 0 | 0 |
| **Total** | **3** | **0** | **0** |

### Highlights
- **Clean Mobile Layout Strategy**: Re-grouping secondary header controls using `w-full order-last sm:contents` cleanly splits the WorkEditor header into two compact rows at `< sm` while dissolving back into a single flex row at `sm+`. This preserves exact DOM order for desktop tab navigation without duplicate control instances.
- **Accessibility & Touch Floor**: Mobile controls meet the 44px touch floor (`min-h-[44px]`), and view-mode toggle buttons retain explicit `aria-label` attributes when text labels collapse under `sm` (`hidden sm:inline`).
- **Comprehensive Test Suite**: `WorkEditor.test.jsx` verifies the layout contract, touch targets, DOM order, non-duplication, and interactive view-mode switching.

_Generated by /do:review_

PR #3589, opened 2026-08-05
refactor: derive perpetual park rollups in taskSchedule from one shared aggregate (#3574)

### LINKED ISSUE #3574: Extract shared aggregatePerpetualParks(execution, now) primitive to dedup per-app park rollups
## Context

The fix that makes a perpetual scheduled task (e.g. `claim-issue` set to perpetual with a `recheckCron: '0 9 * * *'`) wake the daemon at its recheck time added `perpetualUpcomingEligibility(execution, now)` in `server/services/taskSchedule.js`. A `/simplify` reuse review noted this is now the **second** hand-rolled "walk `perApp` values + the global record, read `parkedUntil`, find the soonest recheck / is-any-app-due" traversal in that file, and the single-record `parkedUntil`-vs-now comparison is inlined a **third** time.

The three copies:
- `perpetualUpcomingEligibility` (`server/services/taskSchedule.js`, ~L1861-1873) — short-circuits to `ready` if any scope is unparked/elapsed, else `scheduled` at the soonest park; returns `{ status, eligibleAt }`.
- `getScheduleStatus` perpetual block (`~L1780-1797`) — computes `nextRecheckAt` as the soonest `parkedUntil` among **parked** records only, plus UI metadata (`globalParked`, `parkedAppCount`, `trackedAppCount`, `parkReason`).
- `shouldRunTask` PERPETUAL branch (`~L1421-1445`) — single-record `parkedUntil ? new Date(...).getTime() : 0` then compare to now.

They already diverge subtly: `getScheduleStatus` always evaluates the global `execution` record via `isParked(execution)`, whereas `perpetualUpcomingEligibility` only folds the global record in when it carries its own `parkedUntil`. Same intent, two rules — exactly the drift risk duplication creates.

## Decision (chosen approach)

Extract ONE shared primitive that both aggregators *project* from, rather than merging the two callers (they answer different questions — parked-only `nextRecheckAt` vs `anyDueNow`-short-circuit — so they can't share a drop-in):

```js
// returns numeric ms of a record's perpetual park (0 when unparked)
function parkedUntilMs(record) { return record?.parkedUntil ? new Date(record.parkedUntil).getTime() : 0; }

// aggregate per-app (+ own-parked global) records into one rollup
function aggregatePerpetualParks(execution, now) {
  // → { trackedCount, parkedCount, globalParked, soonestParkAt (ms|null), anyDueNow, parkReason }
}
```

Then derive:
- `perpetualUpcomingEligibility` → `anyDueNow ? { ready, now } : { scheduled, soonestParkAt }` (null when `trackedCount === 0`, preserving the global `ready` default).
- `getScheduleStatus.taskStatus.perpetual` → project `globalParked` / `parkedCount` / `trackedCount` / `soonestParkAt` (ISO) / `parkReason`.
- `shouldRunTask` PERPETUAL branch and the `updateTaskInterval` park-recompute block (`~L801-807`) can reuse `parkedUntilMs`.

## Why this was deferred

Out of scope for the bug-fix PR that introduced `perpetualUpcomingEligibility`: unifying touches `getScheduleStatus`, a UI-critical path with its own test expectations, and the semantic gap (`anyDueNow` vs parked-only `nextRecheckAt`) makes it non-mechanical. Both the reuse and altitude `/simplify` reviewers agreed it should land as its own PR with `getScheduleStatus` regression coverage.

## Acceptance criteria

- [ ] `aggregatePerpetualParks(execution, now)` + `parkedUntilMs(record)` added to `server/services/taskSchedule.js`.
- [ ] `perpetualUpcomingEligibility` and the `getScheduleStatus` perpetual block both derive from the shared aggregator; the pre-fix behavior of each is unchanged.
- [ ] The global-record inclusion rule is unified (documented once) so the two callers can no longer diverge.
- [ ] Existing `taskSchedule.test.js` perpetual/getUpcomingTasks/getScheduleStatus tests still pass; add coverage asserting both callers agree on the same fixture.

atomantic:
## Summary

`server/services/taskSchedule.js` walked a perpetual task's park records in three
places: the `getScheduleStatus` UI block (parked/draining badge + soonest
recheck), `perpetualUpcomingEligibility` (what `getUpcomingTasks` reports, which
is what decides when the daemon wakes), and an inline `parkedUntil` read in the
`shouldRunTask` PERPETUAL branch.

The first two had already drifted on the global-record inclusion rule:
`getScheduleStatus` only folded the top-level execution record in when its park
had **not** elapsed, while `perpetualUpcomingEligibility` folded it in whenever
it carried a park at all. Same intent, two rules.

This extracts the shared primitives from #3574's chosen approach:

- `parkedUntilMs(record)` — the single reader of `parkedUntil` (ms, or `0` when
  absent/unparseable).
- `aggregatePerpetualParks(execution, now)` — one rollup returning
  `{ trackedAppCount, trackedCount, parkedAppCount, globalParked, soonestParkAt, anyDueNow, parkReason }`,
  with the global-record inclusion rule documented once on the aggregator: every
  `perApp` record is a tracked scope; the top-level record is a tracked scope
  only when it carries its **own** `parkedUntil`.

All four sites now derive from those: `perpetualUpcomingEligibility` projects
`anyDueNow`/`soonestParkAt`, the `getScheduleStatus` perpetual block projects the
counts + `soonestParkAt` (ISO), and both the `shouldRunTask` PERPETUAL branch and
the `updateTaskInterval` park-recompute loop use `parkedUntilMs`. Each caller's
existing behavior is preserved — the aggregator's rules are the union the two
already implemented, not a new policy.

The rollup deliberately exposes no field no caller reads.

Closes #3574

## Test plan

- `cd server && NODE_ENV=test npx vitest run services/taskSchedule.test.js` — 151 passed
  (was 145; 6 new).
- New `perpetual park aggregate — getScheduleStatus and getUpcomingTasks agree`
  describe block pins both callers to the same fixtures: every app parked
  (status `nextRecheckAt` === upcoming `eligibleAt`), an elapsed app park, an
  own-parked global record with no per-app records, an **elapsed** global park
  (the case the two used to disagree on), a global park that is the soonest
  among per-app parks, and no tracked scope at all.
- Full sweep of every suite that imports `taskSchedule` (14 files,
  `cosScheduleRoutes`, `apps/taskTypes`, `taskPromptService`, `apps`, `cos`,
  `workflow`, `cosTaskGenerator*`, `meatspacePostReminder`,
  `cleanupAgentWorktree`, `aiAssignments`, `apps.layeredIntelligence`,
  `autonomousJobs/layeredIntelligenceHooks`) — 661 passed.
- No DB-backed suites run.


atomantic:
Heads-up: #3595 (issue #3590) landed on main and changes the same `updateTaskInterval` park-recompute hunk this PR refactors — the guard is now an elapsed check (`parkedUntilMs > nowMs`) rather than a bare truthiness test. This branch is currently CONFLICTING; the rebase should keep the elapsed check while extracting it into the shared aggregate. `.changelog/NEXT.md` will also conflict additively (keep both entries).

atomantic:
Rebased onto `main` after #3590 landed (`f3c0dfcaa`), which fixed the elapsed-park bug an antigravity review of this PR surfaced. The two overlapped in exactly one hunk — the `updateTaskInterval` park-recompute guard:

- main's fix inlined its own `const parkedUntilMs = rec?.parkedUntil ? ... : 0` local (shadowing the module-level helper this PR adds) and gated on `> nowMs`.
- Resolved to the shared helper plus that guard: `if (parkedUntilMs(rec) > nowMs)`. Both intents kept — #3590's "never restamp an already-due park" behavior and its comment, and this PR's "one reader of `parkedUntil`". #3590's regression test is untouched and passes against the refactored code.

One fidelity fix while resolving: the aggregator picked `parkReason` from the first parked app that *had* a reason, where `getScheduleStatus` used `parkedApps[0]?.parkReason` — so a first parked app with no reason used to fall through to the global reason (or null) and would have started borrowing the next app's. Now tracks the first parked record itself, matching the pre-refactor read exactly.

## Test plan

- `cd server && NODE_ENV=test npx vitest run services/taskSchedule.test.js` — 152 passed (151 from this PR + #3590's new elapsed-park test).
- All 14 suites importing `taskSchedule` — 662 passed.

Issue #3675, opened 2026-08-09
Extract usePersistedOption hook for AutopilotPanel persisted-option scaffold


atomantic:
## Problem

`client/src/components/pipeline/AutopilotPanel.jsx` persists each autopilot option through the same seven-touch-point scaffold, hand-copied per option: a `DEFAULT_*` const, a `useState`, a `*EditedRef`, an `edit*` callback, a hydrate-if-untouched line in the settings-load effect, an override-collect line in the run payload, and a `start()` dep-array entry. The observer toggle (added in the observing-orchestrator feature) is the Nth verbatim copy of this scaffold — identical in behavior to the adjacent `selfImprove` toggle.

Missing any one of the seven edits produces the silent "I toggled it and ran, but it didn't apply" stale-closure bug the file's own comments warn about, and each future option costs seven edits in a 1000+ line component.

## Proposed change

Extract a `usePersistedOption(key, defaultValue)` hook (in `client/src/hooks/`, barrel + README per the module-organization rules) that owns the state/editedRef/hydrate/override-collect lifecycle, then migrate the existing AutopilotPanel options onto it. The hook should return `{ value, edit, hydrate, override }` (or an option-descriptor table equivalent) so the settings-load effect and the run-payload builder iterate the registered options instead of enumerating them by hand.

## Acceptance

- New hook registered in `client/src/hooks/` barrel + README.
- AutopilotPanel options (selfImprove, observer, and siblings using the same scaffold) migrated; behavior unchanged.
- `client` suite green, including `AutopilotPanel.test.jsx` (note: this suite has known CI flakes — confirm locally).

PR #3677, opened 2026-08-09
feat: observing orchestrator mode for series autopilot

atomantic:
## Summary
- Adds an opt-in **Observing orchestrator** mode to Series Pipeline Autonomous mode — the continuous, self-dispatching sibling of the existing one-shot "Improve the pipeline itself" post-mortem.
- With the toggle on, each autopilot step's fresh telemetry (a retried/escalated child runner, a skipped step, a filed gap, an editorial check that threw) triggers a bounded diagnosis call asking what in PortOS should change (missing/misordered step, ineffective verify→resolve loop, broken check or stage prompt, misconfigured gate, missing UI option). A dispatchable diagnosis is filed immediately as an **auto-approved** CoS task that works in a worktree, opens a PR, and merges through the normal review loop with no human gate — so a defect noticed at step 4 is being fixed while later steps still run.
- Guards: execute-mode only, every pass billed against and gated on the daily cos budget, capped at 3 mid-run passes + 1 terminal pass per run, a higher confidence bar than the human-gated post-mortem it supersedes when both are enabled, and repeat diagnoses collapse onto one open task via the existing dedup key.
- Shared diagnosis vocabulary/shaping between the post-mortem and the observer is extracted into `diagnosisCore.js` so the two passes stay consistent instead of drifting.
- New `pipeline-observer` stage prompt + stage-config entry, seeded onto existing installs via migration 237 (mirrors the 235 self-improve seed migration).
- Live progress frames surface each observation/dispatch in the Autopilot panel; the run's status banner lists everything filed.

## Test plan
- [x] `NODE_ENV=test npx vitest run` for `server/services/pipeline/seriesAutopilot/**`, `server/lib/pipelineValidation.test.js`, and the new migration test — 282 tests pass
- [x] `npx vitest run src/components/pipeline/AutopilotPanel.test.jsx` in `client/` — 45 tests pass
- [x] `biome lint` on the changed client file — clean


atomantic:
Rebased onto current `main` — this branch conflicted with the `usePersistedOptions` registry refactor (#3675), which landed after it was cut.

The branch added `foundationGate`/`foundationThreshold`/`foundationRounds`/`selfImprove`/`observer` in the old per-option scaffold idiom (a `useState` + `editedRef` + edit callback + hydrate line + override-collect line each). `main` already carries all of those in `OPTION_SPECS` except `observer`, so the resolution keeps the registry and re-expresses `observer` as a single spec row (`readBoolean`, `persistOnEdit: true`, mirroring `selfImprove`), with the control reading `opt.observer` / `options.edit('observer', …)`.

Verified locally: `client` biome lint clean, `AutopilotPanel.test.jsx` + `usePersistedOptions.test.jsx` 57 passed, server observer/seriesAutopilot + migration 237 suites 78 passed. Migration `237-pipeline-observer-stage` does not collide — `main` is at 236.

Issue #3658, opened 2026-08-09
Privacy Center: household subjects (multi-person vault, consent-gated opt-outs)


atomantic:
Split out of #2148, whose federation half is closed by ADR `docs/decisions/2026-08-08-privacy-records-machine-local.md`. This half is unaffected by that decision — household subjects is a **same-machine** feature with no federation dependency.

## Summary

Let the Privacy Center manage PII, orgs, and broker opt-outs for more than one consenting person (partner, child, parent) — the multi-subject model PortOS adapted from unbroker but left at `self` only in v1. The schema already anticipates it (`privacy_consents.subject`).

## Decisions (made — build to these, don't re-litigate)

- **Subjects get their own table, not a free-text string.** Add `privacy_subjects` (`id` UUID PK, `display_name`, `relationship` TEXT, `created_at`, `updated_at`). A migration seeds one row for `self` and backfills it everywhere. A free-text `subject` column would make renames lossy and scoping queries stringly-typed.
- **Scope by FK, defaulted.** Add `subject_id` (FK → `privacy_subjects`, NOT NULL, defaulted to the seeded `self` row) to `privacy_vault_records`, `privacy_orgs`, `privacy_org_holdings`, `privacy_change_events`, `privacy_broker_cases`. Existing rows migrate to `self`, so the change is invisible to a single-subject install.
- **Consent is engine-enforced, not UI-enforced.** Carried verbatim from unbroker's no-consent-no-action rule: `privacy_consents` gains `subject_id` + `method` (how consent was captured). `privacyOptOut.js` and `privacyScan.js` **refuse** to scan or submit for a subject with no active consent row — a hard guard in the service, mirroring the existing disclosure-allowlist guard. The UI hiding the button is not sufficient.
- **Machine-local, like everything else in the Privacy Center.** No federation, no tombstones, hard deletes — ADR `docs/decisions/2026-08-08-privacy-records-machine-local.md` applies unchanged. Deleting a subject cascades their vault/org/holding/case rows.
- **Vault encryption is unchanged.** One `PRIVACY_VAULT_KEY` per install covers all subjects; per-subject keys buy nothing when a single user operates the install.

## Affected files

- `server/lib/db/schema/privacy.js` + `server/scripts/init-db.sql` (schema parity — both, per `server/CLAUDE.md`)
- `scripts/migrations/` — new slice: create `privacy_subjects`, seed `self`, add + backfill `subject_id`
[...]

PR #3679, opened 2026-08-09
Observing orchestrator mode for series autopilot — auto-dispatch pipeline fixes mid-run

atomantic:
## Summary

Adds an opt-in **observer** mode to Series Pipeline autonomous runs: the continuous, self-dispatching sibling of the existing `selfImprove` post-mortem. When a step's fresh telemetry says the *automation* misbehaved (an editorial check that threw, a delegated runner retried, a step skipped), the run spends a bounded diagnosis call (new `pipeline-observer` stage, migration 237) and files an auto-approved CoS task — worktree, PR, review-loop-then-merge — so the pipeline hardens itself while the run continues.

Structural guards: execute-mode only, budget-gated, 3 mid-run passes + 1 terminal, a 0.7 confidence bar, and first-line task dedup. Supersedes the `selfImprove` terminal diagnosis when both are enabled.

The two follow-up commits are a refactor pass: the vocabulary, diagnosis shape, terminal ladder, prompt vars, and filed-task frame shared by `selfImprove` and the observer move into `diagnosisCore.js` (a deliberate leaf outside the `session.js` import cycle, so `OBSERVER_AREAS` derives from `SELF_IMPROVE_AREAS` rather than hand-mirroring it); mid-run observer passes ride a per-run promise chain so a heavy-model call no longer blocks the next pipeline step; and the "opted in, and in execute mode" predicate — previously spelled three different ways — is defined once.

### Why this is opened by hand

The agent that wrote this branch committed and pushed it, then died three consecutive times inside `/do:pr`'s self-review gate before it could open the PR, so the work sat pushed-but-invisible. The reaper bug that killed it is fixed separately; this PR just recovers the stranded branch. It has **not** been through the review gate its author never reached — please treat it as unreviewed.

## Test plan

- `server/services/pipeline/seriesAutopilot/observer.test.js` — 333 new lines covering the opt-in predicate, budget/pass caps, the confidence bar, dedup, and a drift guard asserting every trigger type is also a retained type (otherwise the trigger silently never fires).
- `client/src/components/pipeline/AutopilotPanel.test.jsx` — the new option's wiring and persistence.
- `scripts/migrations/237-pipeline-observer-stage.test.js` — seeds the `pipeline-observer` stage config + template onto existing installs.
- CI on this PR is the first full run this branch has had.


atomantic:
Closing as a duplicate of #3677.

The recovery agent that opened this PR reported the observer-mode work as "stranded, pushed but never PR'd" — that was incorrect. #3677 already existed for the same three commits on `cos/task-mslfojnv/agent-f0410288`, opened ~6 minutes earlier.

The two branches carry an identical autopilot diff, but this one is based on an older `main` (`6f30f9222`) and so is missing the Privacy Center subject-scoping work (#3658). Merging it would have reverted those files. #3677 is based on current `main` and has now been rebased onto the `usePersistedOptions` registry refactor (#3675), with `observer` re-expressed as an `OPTION_SPECS` row.

No work is lost — everything here ships via #3677.

PR #3686, opened 2026-08-09
fix: isolate unlockPass tests from the checkout's real data/ directory

### LINKED ISSUE #3683: unlockPass.test.js reads the developer's real data/ directory — 2 failures on any checkout with real series
## Problem

`server/services/pipeline/seriesAutopilot/unlockPass.test.js` fails 2 of 25 tests when run from a checkout that has a populated `data/` directory — i.e. the primary dev checkout on a live install:

```
× unlocks only the canon this series owns and keeps another series' canon frozen
    unlockPass.test.js:172  expect(counts.canon).toBe(0)          → got 3
× leaves the universe world-field locks alone when a sibling series shares the universe
    unlockPass.test.js:295  expect(counts.worldFields).toBe(0)    → got 1
```

CI is green and always will be: a fresh CI checkout has no real `data/`, so the leak is invisible there. It only bites a developer running the suite locally, where it reads as a spurious regression on whatever branch happens to be checked out.

## Root cause

The suite partial-mocks `lib/fileUtils.js`, replacing `tryReadFile` / `ensureDir` / `atomicWrite` / `readJSONFile` with an in-memory `fileStore` — but it never overrides `PATHS`. So `PATHS.data` still resolves to the checkout's real `data/`.

`unlockPass.js` calls `listSeries()` (`server/services/pipeline/series.js:457`), which goes through `store().loadAll()`. The collection store enumerates `PATHS.data/series/` with a real `readdir`, so `loadAll()` returns the developer's actual series alongside the two the test created. The sole-universe-ownership check then finds real sibling series sharing the universe and correctly keeps the locks the test expected to be released — the assertions are right, the fixture is polluted.

The telltale is the warning the suite already prints on an isolated checkout, which disappears on a populated one:

```
⚠️ autopilot unlock: could not list series to prove sole-universe ownership (db down) — keeping universe-side locks
```

## Verification that this is pre-existing

Bisected against the observing-orchestrator merge (#3677/#3681), which touched `series.js` and was the obvious suspect:

| code | `data/` | result |
|---|---|---|
| post-merge `080d5fad5` | worktree (empty) | 25 passed |
| post-merge `080d5fad5` | primary (real) | 2 failed |
| pre-merge `8908ed2eb` | worktree (empty) | 25 passed |
| pre-merge `8908ed2eb` | primary (real, symlinked in) | **2 failed** |

Pre-merge code fails identically once it can see real data, so the merge is not implicated — `unlockPass.js` and `unlockPass.test.js` are both untouched by it. Copying the primary's `.env` into the worktree did **not** reproduce it; only the real `data/` does.

## Fix

Use the existing shared helper — `makePathsProxy` from `server/lib/mockPathsDataRoot.js`, already used correctly by 44 other suites — to point `PATHS.data` at a `createTempDataRoot()` temp dir, merging the existing `fileStore` overrides in via `extraOverrides`. That is exactly the migration path the helper's docblock describes.

## Same leak in three sibling suites

These also partial-mock `fileUtils.js` with no `PATHS` override and should be checked and converted in the same pass:

- `server/services/pipeline/continuityBible.test.js`
- `server/services/pipeline/reverseOutline.test.js`
- `server/services/pipeline/manuscriptReview.test.js`

## Acceptance criteria

- [ ] `unlockPass.test.js` passes 25/25 from the primary checkout with a populated `data/`.
- [ ] The suite's `PATHS.data` resolves to a temp dir, via `makePathsProxy` / `createTempDataRoot`.
- [ ] The three sibling suites above are audited; any that read `PATHS.data` are converted the same way.
- [ ] A regression probe proves the isolation actually holds — plant a series record in the real-data location and confirm the suite still sees only its own fixtures (per the repo's guard-tests-need-a-bypass-probe convention).
- [ ] No production code changes: this is a test-isolation fix only.

atomantic:
## Summary

`unlockPass.test.js` failed 2 of 25 tests on any checkout with a populated `data/` directory — i.e. the primary dev checkout on a live install. CI was permanently green because a fresh CI checkout has no `data/`, so the failures read as a spurious regression on whatever branch happened to be checked out.

The suite partial-mocked `lib/fileUtils.js` with an in-memory `fileStore` but never overrode `PATHS`, so `PATHS.data` still resolved to the repo's own `data/`. `listSeries()` goes through the series collection store, whose `listIds()` enumerates `PATHS.data/pipeline-series/` with a **real `readdir`** that the fileUtils overrides never intercept. On a populated checkout that readdir succeeds and returns the developer's actual series — which also suppresses the store's in-process-id fallback, replacing the suite's own fixtures. The sole-universe-ownership check then correctly found sibling series and kept the locks the test expected to be released: the assertions were right, the fixture was polluted.

**Fix (test-only, no production code touched):** redirect `PATHS.data` at a per-run temp dir using the shared `makePathsProxy` helper from `server/lib/mockPathsDataRoot.js` — the migration path its docblock describes, already used by 44 other suites — merging the existing fileUtils overrides in through its `overrides` option.

**Sibling suites audited.** The three named as suspects (`continuityBible.test.js`, `reverseOutline.test.js`, `manuscriptReview.test.js`) are already isolated and needed no change: each replaces `series.js` wholesale with a `/mock/series/<id>` `recordDir` and fully replaces `fileUtils.js` (exporting no `PATHS` at all), so none of them can reach the real collection store. Verified by running all three against a checkout with a planted record.

## Test plan

Bypass probe, per the repo's guard-tests-need-a-bypass-probe convention — a probe series record was planted at `data/pipeline-series/<id>/index.json` in the worktree to simulate a populated checkout:

| code | planted record | result |
|---|---|---|
| before fix | none | 25 passed |
| before fix | **planted** | **2 failed** (reproduces the issue exactly) |
| after fix | **planted** | 28 passed |

- `unlockPass.test.js` — 28/28 with the polluted `data/` in place (25 original + 3 new isolation assertions).
- `continuityBible` / `reverseOutline` / `manuscriptReview` — 82/82 with the polluted `data/` in place.
- Full `services/pipeline` suite — 1707 passed, 13 skipped.
- Probe data removed afterward; the temp root is cleaned up in `afterAll`.

Three added assertions pin the isolation so it cannot quietly regress: `PATHS.data` resolves to the temp root, the store's record directory is anchored under it, and `listSeries()` returns only the suite's own fixtures. The last one fails in exactly the original way if the redirect is ever dropped.

One note on what the fix does *not* change: the `⚠️ autopilot unlock: could not list series … (db down)` warning still prints once, from the test that deliberately makes the lookup throw. `listSeries()` genuinely succeeds in the other cases, so the passing assertions are not vacuous.

Closes #3683


atomantic:
Reviewer pass (antigravity) — two findings raised, both verified against the running suite and neither holds. No code change.

**1. "`vi.mock` hoisting makes `tempRoot` throw `ReferenceError`; wrap it in `vi.hoisted()`."** Not reproducible, and the suggested fix is what actually breaks the file. Vitest hoists the `vi.mock` *registration*, but the factory is lazy — it runs when the mocked module is first imported, which here is the `await import('../series.js')` well below the `const tempRoot`. The suite runs 28/28 as written. Applying the suggestion verbatim fails at load:

```
ReferenceError: Cannot access '__vi_import_3__' before initialization
Test Files  1 failed (1)   Tests  no tests
```

`vi.hoisted` runs before ESM import bindings initialize, so `createTempDataRoot` — an imported binding — is in the TDZ inside it. The plain top-level const is also the established pattern for this helper (`socialAccounts.tombstone.test.js`, `digital-twin-documents.tombstone.test.js` pass `dataRoot: tempRoot` the same way), and it is unchanged from the pre-existing `fileStore` reference in the same factory.

**2. "`lists only the series this suite created` is order-dependent — earlier tests' series leak in."** They do not: `beforeEach` clears `fileStore`, so `readJSONFile` returns null for every earlier id and `loadAllResult` drops them as unreadable. Exact equality is the stronger assertion and is the one that reproduces the original failure, so `toContain` would weaken the probe. Confirmed order-independent with `--sequence.shuffle`: 28/28.

Suite is green on a checkout with a populated `data/` — the condition that produced the original 2 failures.

Issue #3736, opened 2026-08-10
Collapse the per-runtime video mode gates into one contract in modeContract.js


atomantic:
## Problem

`prepareParams.js` now carries one near-identical mode gate per video runtime. Each copy is the same five steps:

1. resolve `requestedMode` from `mode || (hasSource ? 'image' : 'text')`
2. check it against the model's `supportedModes`
3. text-forbids-a-conditioning-image
4. image-requires-a-source
5. (until #TBD, chunks>1 requires i2v — now hoisted to one shared `videoChainUnsupportedError`)

Today that shape exists for `wan22` (`prepareParams.js`, the `runtime === 'wan22'` block) and `minimax_h3` (the block above it), and again at the render boundary in `local.js` (`buildWan22Args` / `assertMiniMaxH3SupportedInputs`). The copies have already drifted — wan22 has no FFLF/last-image rules at all, and the two error-code families (`WAN22_*` vs `MINIMAX_H3_*`) diverge for identical request shapes.

Adding MiniMax H3 image-to-video paid this tax a second time: ~40 lines of near-copy plus a new `modeContract.js` leaf holding just the H3 table.

## Why it matters

Every new runtime costs ~40 lines of copy-paste across two files, and each copy is an independent drift site. The failure mode is silent: a runtime whose gate forgets one rule renders a materially different clip than requested rather than erroring.

## Proposed shape

One `assertVideoModeContract({ model, mode, hasFirstImage, hasLastImage, label, codePrefix })` in `server/services/videoGen/modeContract.js` (the leaf added for H3 — extend it rather than adding another file), driven entirely by `model.supportedModes`. Runtime-specific extras stay where they are: H3's fps / negative-prompt / tiling / frame-grid rules, wan22's `frameStride`.

Per-runtime error-code prefixes survive as a parameter — verified no client code reads these codes (`grep -rn "WAN22_\|MINIMAX_H3_" client/src` is empty), so they are server/API-contract only and can stay distinct without duplicating the logic.

The existing `miniMaxH3InputError` / `videoChainUnsupportedError` in `modeContract.js` are the seed of this — they already prove the two-boundary pattern works with a returned-error rule table.

## Acceptance

- [ ] `wan22` and `minimax_h3` mode gates in `prepareParams.js` and `local.js` all resolve through one shared contract function
[...]

PR #3760, opened 2026-08-10
Give every video model a resolved supportedModes so capability stops being a hardcoded runtime list

### LINKED ISSUE #3737: Give every video model a resolved supportedModes so capability stops being a hardcoded runtime list
## Problem

"Which modes does this video model actually support?" has no single answer today:

- `wan22` and `minimax_h3` entries in `server/lib/mediaModels.js` declare `supportedModes`.
- `mlx_video` and `ltx2` entries declare nothing — so `isModelAllowedForMode()` (`client/src/lib/videoGenParams.js`) returns `true` for every mode, including ones the runtime only *fakes* (mlx_video's FFLF conditions on a single frame and drops the other).
- `hunyuan` still carries a legacy `mode: 't2v'` field that nothing reads.

Because the registry can't answer, consumers re-derive capability from `runtime` string comparisons instead — there are 25+ such sites (`grep -rn "'ltx2'" server client`). Enabling MiniMax H3 image-to-video (#TBD) hit this twice and had to add `LAST_FRAME_ANCHORED_RUNTIMES` in `server/services/videoGen/runtimes.js` plus a `lastFrameAnchored` decoration in `listVideoModels()` just so the client's advisory-note predicate and the server's last-frame resize couldn't drift.

## Why it matters

Every capability question that the registry can't answer becomes a hardcoded runtime list, usually duplicated once per side of the wire. The failure mode is a UI that claims a capability the render path doesn't honor (or vice versa) — wrong output, not an error.

## Proposed shape

Backfill an explicit `supportedModes` onto **every** shipped video entry at load, keyed by runtime, the same way `applyVideoDisclosures` / `applyVideoFinishProfiles` already backfill shipped facts in `server/lib/mediaModels.js`. Derived at load, not persisted — so no migration and no user-registry churn.

Then:
- `isModelAllowedForMode()` stops defaulting to "true for everything".
- `mlx_video` declares `['text', 'image', 'fflf']` honestly, with `lastFrameAnchored: false` carrying the caveat (already in place).
- Retire `hunyuan`'s legacy `mode: 't2v'` in favor of `supportedModes: ['text']`.

## Acceptance

- [ ] Every shipped entry in `DEFAULT_REGISTRY` resolves a `supportedModes` after load (test asserts none is undefined)
- [ ] Backfill is derived, not written to `data/media-models.json` — a user entry that declares its own list still wins
- [ ] `isModelAllowedForMode` no longer has a "no supportedModes ⇒ allow everything" branch
- [ ] `mode: 't2v'` removed from the hunyuan entry and from any reader

atomantic:
## Summary

"Which modes does this video model support?" had no single answer: `wan22` and `minimax_h3` entries declared `supportedModes`, `mlx_video` and `ltx2` declared nothing, and `hunyuan` carried a legacy `mode: 't2v'` field no reader consulted. Consumers therefore read "declares nothing" as "supports everything" — so the model picker offered FFLF on runtimes whose helper takes a single `--image` and silently drops the other keyframe. That's a UI claiming a capability the render path doesn't honor: wrong output, not an error.

- **New `server/lib/videoModeProfiles.js`** states each runtime's modes once, as facts about its helper's argv: `mlx_video`/`ltx2` text+image+fflf+extend, `wan22` text+image, `minimax_h3` text+image+fflf (also the mode ceiling `modeContract.js` enforces, so the two can't drift), `hunyuan` text only.
- **Resolved on read in `getVideoModels()`**, the way `listVideoModels()` already decorates `lastFrameAnchored`. That covers the load path, the user-model mutators (which bypass `normalizeRegistry`) and peer-synced entries in one place — and keeps the derived list out of `data/media-models.json`, where a persisted copy would read back as a *declared* list that no later correction to the table could reach. No migration; an entry that declares its own list always wins.
- **`isModelAllowedForMode` loses its "no `supportedModes` ⇒ allow everything" branch.** The two hand-rolled copies of the same predicate (`useVideoGenForm`'s `modelSupportsChaining`, `AdvancedParamsPanel`'s `showChunks`) now call it instead of re-reading the field, and `videoChainUnsupportedError` / `videoModeContractError` read the same resolver — so the server no longer waves an unresolved entry through as unconstrained while the picker hides it.
- **`hunyuan`'s `mode: 't2v'` is retired** for `supportedModes: ['text']`. Nothing read it; the seed drops it in one line and existing installs simply resolve `['text']` on their next boot.

mlx_video's FFLF is declared **honestly rather than absent** — the mode does run, and the caveat rides on the existing `lastFrameAnchored: false` the client renders as "last frame is advisory". An unknown runtime (a user-added model, or one synced from a newer peer) keeps the full base mode set rather than vanishing from every picker.

### Acceptance

- [x] Every shipped entry resolves a `supportedModes` (test asserts none is undefined, plus a table row exists for every runtime in the seed, both platforms)
- [x] Derived, not written to `data/media-models.json` — pinned by a test that reads the on-disk file back; a user entry that declares its own list still wins
- [x] `isModelAllowedForMode` no longer has a "no supportedModes ⇒ allow everything" branch
- [x] `mode: 't2v'` removed from the hunyuan entry (there were no readers)

**One deliberate behavior change beyond the issue:** a `wan22` entry that declares no `supportedModes` used to be rejected by the server in *every* mode ("declares nothing ⇒ renders nothing") while the picker offered it in all of them. It now resolves the wan22 row (text + image) at both ends. `services/videoGen/modeContract.test.js` records this.

## Test plan

- `cd server && NODE_ENV=test npx vitest run` — 1301 files / 26.9k tests pass (two unrelated timeout flakes under load, each green in isolation).
- `cd client && npx vitest run` — 576 files / 6838 tests pass (one unrelated `pitchDetect` timing flake, green in isolation).
- `cd client && npm run lint` — clean.
- New: `server/lib/videoModeProfiles.test.js`; new `supportedModes resolution (#3737)` block in `server/lib/mediaModels.test.js` (all-entries-resolved, runtime-table coverage, hunyuan contract, not-persisted, declared-wins-vs-resolved).

Closes #3737


atomantic:
Reviewer pass (antigravity/agy, round 1): one Medium finding, closed as a false positive.

It flagged that `videoModeContractError` calls in `prepareParams.js` omit `audioFile` / `audioStartSec` / `icReferences`, so H3 requests carrying audio or IC conditioning would stage uploads and enqueue a doomed job. They don't: those inputs are rejected earlier in the same function by `IC_LORA_REQUIRES_LTX2` (prepareParams.js:369), `A2V_REQUIRES_LTX2` (:398) and `KEYFRAMES_REQUIRE_LTX2` (:640), all before staging — which is precisely why the wan22 row sets `extraConditioningUnsupported: false` ("folding them in here would double-report"). That code is also #3736's, untouched by this diff.

Its answers to the three questions I posed describe pre-change behavior (it says an absent `supportedModes` still falls back to `true`, which is the branch this PR removes), so they carry no weight either way. No code changes from this round.

PR #3766, opened 2026-08-10
Apply chained-video chunk trims in the concat filter graph instead of pre-encoding each chunk

### LINKED ISSUE #3761: Chained video stitch: apply chunk trims in the concat filter graph instead of pre-encoding each chunk
## Context

Windowed chunk continuation (added alongside the `contextFrames` control) cuts the echoed context prefix off each chained chunk with an in-place re-encode in `generateChainedVideo` (`server/services/videoGen/local.js`), and then — because that trim breaks codec-parameter lockstep with the never-trimmed chunk 0 — `stitchVideos` re-encodes the whole timeline via its `reencode` option.

So an N-chunk chain pays `(N-1)` full x264 crf-18 encode passes whose only product is a frame offset, plus one more encode over the entire timeline. Trimmed chunks also take two lossy passes where chunk 0 takes one, so they are graded slightly differently from it.

## Proposal

Don't rewrite the chunk files. Record the prefix offset per chunk and hand those offsets to `stitchVideos`, which — since it is already re-encoding in this case — can apply them in a concat **filter graph** rather than the demuxer:

```
-i chunk0 -i chunk1 … -filter_complex
  "[1:v]trim=start_frame=P1,setpts=PTS-STARTPTS[v1]; [1:a]atrim=start=P1/fps,asetpts=PTS-STARTPTS[a1]; …
   [0:v][0:a][v1][a1]…concat=n=N:v=1:a=1[v][a]"
```

`server/services/videoTimeline/local.js` already builds exactly this shape and is the reference implementation.

## Payoff

- Total cost drops from `(N-1)` chunk encodes + 1 timeline encode to exactly 1 timeline encode.
- Removes the generation-loss asymmetry between trimmed and untrimmed chunks.
- The archived per-chunk gallery entries stay byte-identical to what the model rendered.
- `stitchVideos`'s `reencode` flag collapses into the trim path (a trimmed concat necessarily re-encodes), and the `frameCounts` option goes away with it.

## Watch out for

- `videoTimeline/local.js`'s own comments document the pitfalls: filter inputs need `fps=` resampling and `aformat` matching or ffmpeg fails with "Input link parameters do not match". Chain chunks are same-model/same-resolution by construction, which helps, but the audio leg still needs care.
- Silent clips: chain chunks share one model and one audio setting so they are never mixed, but the graph must still build without an `[k:a]` leg when there is no audio stream (`hasAudioStream` already exists for this).
- The hand-stitch path from Media History shares `stitchVideos` and must keep its stream-copy fast path — only the trim-bearing caller should reach the filter graph.
- The per-chunk entries are `hidden: true`, so leaving the echo in them is invisible by default; that is the trade this change accepts.

## Scope

`server/services/videoGen/local.js` (`generateChainedVideo`, `stitchVideos`) and its suite. No client, route, or schema changes.

atomantic:
## Summary

Windowed chunk continuation conditions chunk N+1 on the tail of chunk N, and LTX-2's extend pipeline returns `source + extension` — so every chunk after the first opens with an echo of that window, which the stitched timeline already holds.

Cutting that echo used to mean rewriting each chunk file in place. A frame-index trim can't be stream-copied, so an N-chunk chain spent `(N-1)` full x264 crf-18 passes whose only product was a frame offset — and the trimmed chunks then fell out of codec lockstep with the never-trimmed chunk 0, forcing `stitchVideos` to re-encode the whole timeline on top. Trimmed chunks also took two lossy passes where chunk 0 took one, so they were graded slightly differently from it.

The offsets now ride into the stitch instead of into the files:

- **`buildTrimConcatArgs`** (`server/lib/ffmpeg.js`) — pure builder for a `filter_complex` concat that drops leading frames from selected inputs. Mirrors the normalization `services/videoTimeline/local.js` already proved out: `scale`/`pad`/`setsar`/`fps` on the video leg and `aresample`/`aformat` on the audio leg, which is what keeps ffmpeg from aborting with "Input link parameters do not match". Each is applied only when it resolves to a usable number, so a caller with no canonical value gets no normalization rather than `scale=undefined:undefined`.
- **`stitchVideos`** takes a `trims: Map<videoId, {startFrame, frames}>`. The old `reencode` and `frameCounts` options collapse into it. Anything with no offsets — including the hand-stitch path from Media History — keeps the demuxer stream-copy fast path untouched.
- **`generateChainedVideo`** records the measured prefix per chunk instead of calling `trimVideoFromFrame` on it.

Payoff: total cost is exactly one timeline encode; the generation-loss asymmetry between trimmed and untrimmed chunks is gone; and the archived per-chunk gallery entries stay byte-identical to what the model rendered (carrying their echo, which is invisible — those entries are `hidden: true`).

### Details worth a look

- **Audio legs are emitted only when *every* input has an audio stream.** `concat=a=1` needs a leg from each input and referencing `[k:a]` on a silent clip aborts the run; a chain of silent renders takes the video-only graph. When a trim is present but no fps is known, audio is dropped rather than left desynced from a video leg that was cut.
- **The conditioning-window cut now floors at the echo.** The chunk file still holds its prefix, so `tailWindowStartFrame` clamping to 0 (window larger than the chunk's own new footage) would hand the next hop a replay of the chunk *before* it. `Math.max(contextPrefix, …)` keeps the window inside the chunk's own footage — the same frames the pre-trim path selected.
- **Graceful degradation is preserved.** The old code degraded to a repeated seam when a chunk trim failed rather than discarding every chunk already rendered. A filter graph that fails now falls back to the stream-copy demuxer over the untrimmed chunks (never re-encoded, so still in lockstep), and `numFrames` reports the timeline that actually exists.

## Test plan

- `server/lib/ffmpeg.test.js` — 7 new cases over the pure builder: trims only the offset-bearing inputs, omits geometry/rate filters with no usable value (no `undefined`/`NaN` in the graph), emits matching audio legs when all inputs have audio, drops audio rather than desyncing it when a trim has no rate, keeps audio on a rate-less untrimmed concat, clamps negative/non-numeric offsets.
- `server/services/videoGen/local.test.js` — the chain suite now asserts that no chunk file is rewritten, that the offsets appear as `trim=start_frame=` on inputs 1+ (and never on input 0) in the graph, that exactly one `libx264` invocation runs for a 3-chunk chain, that the graph stays video-only for silent chunks, that the window start floors at the prefix, that a failing graph falls back to a stream copy with `numFrames` corrected, and that `contextFrames: 0` still keeps the demuxer fast path.
- Both were bypass-probed: reverting `trimsApplied` to `false` fails 4 of them; dropping the window floor fails 1.
- Validated end-to-end against real ffmpeg (not just argv assertions): two 1s 24fps clips, second offset by 9 frames, built via `buildTrimConcatArgs` — exits 0 and produces exactly 39 frames (24 + 15), both silent and with an AAC track.
- Full server suite: 27026 passed, 251 skipped (DB suites gated off as expected).

Closes #3761


atomantic:
## Review record

**Configured reviewers: `codex, claude`.**

`codex` could not run — the account is at its usage limit until Aug 15 (`ERROR: You've hit your usage limit`), so it produced no findings at any point. Per the established fallback for a reviewer that dies on a spend limit, the loop converged on the surviving reviewer rather than stalling; noting it here so the audit trail is honest.

`claude` ran three rounds plus a convergence check:

**Round 1** — one defect, two notes, all addressed in `06eb5fbb2`:
- The filter-graph fallback summed each input's history `numFrames`, which is the count it was *requested* at. A windowed chunk's file is `context + extension`, so a 3-chunk chain that fell back recorded ~75 frames for a ~123-frame clip and a Remix off it re-rendered short. `trims` entries now carry a measured length even with nothing to cut (`startFrame: 0`), which also covers an extend chain's never-trimmed first chunk.
- `trim` now leads the video leg instead of following `fps=`, so `start_frame` indexes the source frames the caller probed — and the discarded frames aren't scaled and padded first.
- The survivable graph run captures stderr.

Round 1 also found the test probes were mocking every chunk at its rendered frame count, which is what hid the first defect; they now use the lengths a real chain produces (25 for a plain render, 49 for a hop).

**Round 2** — two defects, both addressed in `89b8c74f2`: the fallback log counted measurement-only entries as repeated seams, and `listFileWritten` was set after the write resolved so a partial write leaked the concat manifest. The fatal demuxer run also picked up stderr capture. The latent `chunkExtendLatents` / `numFrames`-default mismatch it flagged as out of scope was rolled in rather than deferred — see the PR body.

**Round 3** — one defect, addressed in `f8a722243`: the round-2 `defaultFrames` lookup dereferenced `chainModel`, which `videoChainUnsupportedError` waves through as `null` when a model is removed from the registry between enqueue and dispatch. That turned a clean 400 into a `TypeError` out of the dispatcher.

**Convergence check on the final delta: CLEAN.**

Every fix is pinned by a test that was bypass-probed — reverting it fails the test. Full server suite: 27028 passed, 251 skipped.

PR #3820, opened 2026-08-11
Classify LoRA safetensors key layouts and refuse un-fusable ones before a video render

### LINKED ISSUE #3799: [ref-watch-phosphene-lora-layout-classification] Classify LoRA safetensors key layouts before use
**Provenance:** From `reference-watch` review of Phosphene (commit `d4c6ac55`, 2026-08-11).

Downloaded video LoRAs arrive in several incompatible safetensors key layouts (bare, ComfyUI-prefixed, diffusers/PEFT, kohya), and PortOS fuses user LoRAs into the ltx2 transformer assuming one layout — a wrong-layout file silently renders noise or dies as an opaque error deep in the Python render. The reference classifies the layout from the safetensors header at install time, normalizes ComfyUI-prefixed keys, and refuses diffusers/PEFT layouts with a clear reason (their alpha/rank scale is not stored in the file, so applying them is guesswork).

**Fix:** `server/lib/safetensors.js` — extend the existing header reader with a layout classifier returning `bare | comfyui | diffusers | kohya | not_a_lora`. `server/services/loras.js#listLoras()` / `#installFromCivitai()` — detect and persist the layout in the sidecar on install (and backfill on list, mirroring the existing base-model header fallback). `server/services/videoGen/local.js#resolveVideoLoras()` — reject un-fusable layouts with an actionable 400 naming the layout and why, instead of letting the render fail opaquely.

**Estimated scope:** medium.

atomantic:
## Summary

Video LoRAs arrive in several mutually incompatible safetensors key layouts, and the LTX-2 MLX loader fuses exactly one shape: `<module>.lora_A.weight` / `<module>.lora_B.weight` pairs, optionally under a `diffusion_model.` prefix that it strips itself (`ltx_core_mlx/loader/fuse_loras.py`, `sd_ops.py`). A kohya/LyCORIS file (`lora_down`/`lora_up` + `alpha`) or a diffusers/PEFT-prefixed file matches nothing in the transformer, so the render burns its full GPU time and comes back as if no LoRA had been selected — no error anywhere. Kohya files are commonly published *with* the ComfyUI `diffusion_model.` prefix, so the prefix alone was never a reliable signal.

- `server/lib/safetensors.js` — extends the existing header reader with `classifyLoraKeyLayoutFromHeader` / `classifyLoraKeyLayout` returning `LORA_KEY_LAYOUTS` (`bare` | `comfyui` | `diffusers` | `kohya` | `not_a_lora`), plus `videoLoraLayoutIssue(layout)` for the user-facing reason a layout can't fuse. Each tensor votes and the dominant layout wins (mixed files exist); ties resolve toward the least-fusable layout. `null` is the explicit "couldn't read the header" sentinel and is never conflated with `not_a_lora`.
- `server/services/loras.js` — both installers stamp `keyLayout` into the sidecar; `listLoras()` backfills it for earlier installs and surfaces it on each entry. The flux-variant and key-layout derivations now share one lazy header read and one coalesced sidecar patch (two fire-and-forget patches would read-modify-write the same sidecar and drop a field). New `getLoraKeyLayout(filename)` resolves sidecar → header.
- `server/services/videoGen/local.js` — `resolveVideoLoras` is now async and rejects an un-fusable layout with a 400 `LORA_LAYOUT_UNSUPPORTED` naming the layout and what to use instead, before any GPU work. An undetermined layout is logged and allowed through: unreadable is not evidence of unusable.

Classification rules were validated against real installed LoRA headers, including a kohya-under-`diffusion_model.` file that today silently no-ops.

## Test plan

- `cd server && NODE_ENV=test npx vitest run lib/safetensors.test.js services/loras.test.js services/videoGen/local.test.js` — 203 passing, including new coverage for each layout, the kohya-under-ComfyUI-prefix case, mixed-layout dominance, `not_a_lora` vs `null`, sidecar backfill/persistence, and the render gate (accepts bare/comfyui, rejects kohya/diffusers/not_a_lora, rejects the whole render when any one selected LoRA is un-fusable, passes undetermined through).
- Full server suite run locally.

Closes #3799


atomantic:
CI note: the one `test (24.x)` failure is `scripts/setup-data-drift.test.js > reproduces the full stage drift table` — a prompt-hash baseline drift on `pipeline-character-foundation.md` / `pipeline-judge-foundation.md`. It fails identically on `main` at this branch's base commit (run 31543531827), so it is pre-existing and unrelated to this change. Every other job (lint, client, DB/smoke) is green.

PR #3882, opened 2026-08-12
fix: validate memory list/timeline query strings with their Zod schemas

### LINKED ISSUE #3867: Wire memoryListSchema and memoryTimelineSchema into the memory GET routes
Two Zod schemas are imported into `server/routes/memory.js` but never wired into the routes they were written for; both handlers hand-parse `req.query` instead.

**Where**
- `server/routes/memory.js:13-22` — `memoryListSchema` and `memoryTimelineSchema` are imported.
- `server/routes/memory.js:27-43` — `GET /` destructures `req.query` manually (`req.query.types ? req.query.types.split(',') : undefined`, `req.query.status || 'active'`, …) with no `validateRequest` call.
- `server/routes/memory.js:64-75` — `GET /timeline` does the same with `memoryTimelineSchema` unused.

Every other route in the same file (`/sync`, `/search`, `/`, `/consolidate`, `/link`) correctly calls `validateRequest(schema, req.body)`, so this is an inconsistency rather than a missing capability.

**Why it matters**

`memoryListSchema` declares `limit: z.number().int().min(1).max(100)`, but the route only bounds `limit` through `parsePagination`'s separate `maxLimit: 500`. The declared cap is silently unenforced — the exported-but-unwired sub-schema creates false confidence, which is the specific failure mode `server/CLAUDE.md` warns about.

**Why this was deferred**

Found during a refactor-only audit (`/do:simplify`), which is contractually limited to observably behavior-preserving changes. Wiring the schemas in **changes validation**: requests with `limit` between 101 and 500 succeed today and would start returning a 400, and the error response shape changes from `parsePagination`'s to `validateRequest`'s. That needs a run allowed to change behavior.

**Suggested fix**

Replace the manual `req.query` destructuring in the `/` and `/timeline` handlers with `validateRequest(memoryListSchema, req.query)` and `validateRequest(memoryTimelineSchema, req.query)`, matching the rest of the file. Decide deliberately whether the real cap is 100 (the schema) or 500 (`parsePagination`) and make the two agree, rather than letting the stricter one win by accident.

atomantic:
## Summary

`GET /api/memory` and `GET /api/memory/timeline` hand-parsed `req.query` while `memoryListSchema` and `memoryTimelineSchema` were imported into `server/routes/memory.js` and never used — the exported-but-unwired sub-schema case `server/CLAUDE.md` calls out. Both handlers now call `validateRequest(schema, req.query)` like every other route in the file.

To make that possible the two schemas learned the query-string form of their own inputs, accepting both it and the already-parsed form:

- comma-separated lists (`?types=fact,learning`) become arrays — the shape `client/src/services/apiMemory.js` already sends
- numeric strings (`?limit=25`) become numbers; a non-numeric `limit` is now a 400 instead of silently falling back to the default
- an empty param (`?types=`, which the CoS Memory tab sends when no type filter is selected) means "unset", not an empty list or `NaN`
- timeline `startDate`/`endDate` accept an ISO datetime **or** a bare `YYYY-MM-DD` — both reach Postgres as bind params, so either is usable, and anything else is now rejected instead of passed through

The declared-cap conflict is resolved in favour of 500: `memoryListSchema` said `max(100)` but the route's real ceiling has always been `parsePagination`'s `maxLimit: 500`, which is also what `memoryTimelineSchema` already declared. Raising the schema to 500 makes the two agree without shrinking a limit that older clients and federated peers may already be passing.

Behaviour changes, deliberately: an unknown memory type, an invalid `sortBy`/`sortOrder`/`status`, a non-numeric or out-of-range `limit`, or a malformed date boundary now return a 400 `VALIDATION_ERROR` rather than being accepted. `parsePagination` is still used by the two routes that keep it (`GET /sync`, `GET /:id/related`).

## Test plan

- `cd server && NODE_ENV=test npm test` — full server suite passes (the one failure, `services/updateExecutor.test.js`, is a known pre-existing local-env flake unrelated to this change and also fails on `main`)
- New coverage in `server/routes/memory.test.js`: both GET routes assert the exact options object handed to `getMemories` / `getTimeline` (defaults case and fully-populated case), plus 400s for an over-ceiling limit, an unknown type, a non-numeric limit, and a malformed date
- New coverage in `server/lib/memoryValidation.test.js`: csv/numeric coercion, empty params falling back to defaults, the 500 ceiling and the boundary value, and the date-boundary union
- Manually traced both client callers (`MemoryTab`, `Tribe`) — `limit: 100`/`25`, `types: []` → `?types=` — against the new schemas; all still validate

Closes #3867


atomantic:
Review loop (reviewer: antigravity / `agy`, headless):

- **Round 1** — 5 findings, all applied: the timeline `appId` filter (supported by both backends, never reachable through the old hand-built options object), a `YYYY-MM-DD` that matches the shape but is not a real day (`2026-02-30` would have reached Postgres as a 500), ISO datetimes carrying a `+HH:MM` offset instead of `Z`, element-length bounds on the csv `categories`/`tags`, and rejection tests for `sortBy`/`sortOrder`/`offset`/`status`.
- **Round 2** — 3 findings, 2 applied: the file backend compared `createdAt` to a date bound lexicographically, which is wrong once a bound can carry an offset, and `Date.UTC` maps a 0-99 year onto 1900-1999 so the calendar check rejected years `0000`-`0099`. Declined: normalizing a bare `endDate` to end-of-day — that would change the midnight semantics these bounds have always had, and a caller wanting part of a day can pass a full timestamp (now noted in the schema).
- **Round 3** — could not run; the reviewer returned `RESOURCE_EXHAUSTED (429)` on its pinned model and errored out on the two fallbacks. No findings outstanding from the rounds that did run.

Full server suite passes (`cd server && NODE_ENV=test npm test`); the only failure is the pre-existing `services/updateExecutor.test.js` local-env flake, which fails on `main` too.

PR #3946, opened 2026-08-13
fix: guard SongBook attachment mutations against the 'failed' presence sentinel (#3900)

### LINKED ISSUE #3900: [SongBook] Deleting attachment when presence lookup failed crashes with TypeError
### Problem
In `client/src/pages/SongBookViewer.jsx:171`, when `listSongAttachments` rejects, `attachments` state is set to the string `'failed'`. Later, inside `onDeleteAttachment` (`client/src/pages/SongBookViewer.jsx:338-349`), `setAttachments` runs:
```js
setAttachments((prev) => (res?.attachments || []).map((meta) => ({
  ...meta,
  present: prev?.find((a) => a.filename === meta.filename)?.present ?? false,
})));
```
Because `prev` is the string `'failed'`, calling `prev?.find(...)` attempts `'failed'.find(...)`, which throws `TypeError: prev.find is not a function`.

### Impact
If attachment presence fails to fetch on page load, attempting to delete any attachment crashes the deletion handler with an uncaught JavaScript error, leaving the UI state broken without deleting the attachment from view.

### Fix
In `SongBookViewer.jsx:343`, guard `prev` to verify it is an Array before calling `.find()`, using `(Array.isArray(prev) ? prev : []).find(...)`.

Files to touch:

### Acceptance Criteria
- [ ] Deleting an attachment when `attachments === 'failed'` does not throw `TypeError: prev.find is not a function`.
- [ ] Attachment list state updates cleanly regardless of prior attachment load status.

atomantic:
## Summary

`SongBookViewer` keeps attachment state as a three-state value: `null` (not fetched), the string `'failed'` (presence lookup errored — the record's own synced list is still shown, with presence unknown), or an array. Both mutation handlers assumed an array:

- **Delete** ran `prev?.find(...)` on the sentinel and threw `TypeError: prev.find is not a function`, so the deleted row stayed on screen and the handler died mid-update.
- **Upload** ran `[...(prev || [])]`, spreading `'failed'` into six single-character rows.

Both now check `Array.isArray` first. Delete leaves `present` undefined when the prior state was the sentinel — presence is genuinely unknown after a failed lookup, so stamping `present: false` would render a false "not on this machine" pill. Upload falls back to the record's own synced attachment list so the visible entries survive the append instead of being wiped down to the new file.

## Test plan

- `cd client && npx vitest run src/pages/SongBookViewer.test.jsx` — 33 passed, including two new regression tests under `attachment mutations after a failed presence lookup (#3900)`.
- Bypass probe: with the `SongBookViewer.jsx` fix stashed, both new tests fail (`TypeError: prev?.find is not a function`, and 8 list items instead of 3).
- `cd client && npm test` — full suite green apart from one unrelated load-dependent flake (a different test file failed on each of two runs: `SeriesSettings` provider select, then `Catalog` album-load retry; neither touches this diff).

Closes #3900


atomantic:
**Review substitution:** the configured reviewer for this run (`agy` / antigravity) could not review — the account quota is exhausted (`Individual quota reached`, resets in ~4h) and a headless invocation also fails with `bubbletea: error opening TTY`. `codex` was run as an independent substitute reviewer instead.

**Round 1 — `codex`: CHANGES_REQUESTED.** One P2, and it was correct: after a failed presence lookup, an *upload* replaced the `'failed'` sentinel with an array whose synced entries had no resolved `present` field, so the next *delete* treated that array as fully known and its `?? false` stamped every survivor with a false "not on this machine" pill.

Fixed in `ccc99ada5`: the delete reducer now carries over presence only where it is an explicit boolean, leaving genuinely-unresolved entries unknown (rendered as links). Added the upload-then-delete regression test the reviewer asked for — it fails against the round-1 reducer and passes now.

**Round 2 — `codex`: no remaining findings.** "Independent runtime, state-lifecycle, and producer/consumer contract passes all agree: the prior P2 is fixed and there are no remaining correctness findings in the diff." It could not execute the suite itself (its read-only sandbox blocks Vitest's temp writes) and then hung in a wait loop without printing its final verdict line, so the run was stopped after that conclusion; it made no changes to the worktree.

Tests: `cd client && npx vitest run src/pages/SongBookViewer.test.jsx` — 34 passed. Full `cd client && npm test` green apart from one unrelated load-dependent flake (a different file failed on each of two runs).

PR #3947, opened 2026-08-13
fix: re-sync the backup cron on settings save so post-boot enablement schedules (#3910)

### LINKED ISSUE #3910: [Docs Drift] Backup scheduler cron registration skipped when enabled post-boot
## Problem
`docs/BACKUP.md:33` states that the scheduled backup cron handler in `server/services/backupScheduler.js` re-reads settings on every run, so `destPath`, `excludePaths`, `disabledDefaultExcludes`, and `enabled` all take effect on the next scheduled run without a restart.

However, `server/services/backupScheduler.js:18-30` checks `if (settings.backup?.enabled === false)` or `if (!settings.backup?.destPath)` when `startBackupScheduler()` is executed during server bootstrap (`server/services/bootstrap.js:77`). If backup was disabled or `destPath` was unconfigured when PortOS booted, `startBackupScheduler()` returns early without calling `schedule()`. Consequently, configuring `destPath` or enabling backup in the Settings UI post-boot does not schedule any backup jobs until the server process is restarted.

## Impact
Users who enable backup or configure a backup destination directory in the Backup Settings UI after server boot will expect daily backups to execute automatically as documented, but no scheduled backup will run until the server is restarted.

## Fix
Update `docs/BACKUP.md:33` to clarify that enabling backups or setting `destPath` when backup scheduling was previously inactive requires a server restart for the cron job registration to take effect.

Files touched:
- `docs/BACKUP.md`

## Acceptance Criteria
- [ ] `docs/BACKUP.md` accurately documents that enabling backups or setting `destPath` when initially disabled/unset requires a server restart for cron job registration to occur.
- [ ] Documentation notes that dynamically re-reading settings on scheduled runs only applies to active scheduled backup handlers.

atomantic:
## Summary

Issue #3910 was filed as docs drift, but the documented behavior is the correct one — the code is the bug, so this fixes the code and documents the (now true) behavior.

`startBackupScheduler()` ran once during boot and returned early when `backup.enabled === false` or `backup.destPath` was unset. Nothing re-ran it, so a user who enabled backups or chose a destination folder in Backup settings after the server was already running got **no scheduled backup at all** until the process restarted. The cron expression had the same problem — editing it was ignored until restart.

Registration now lives in `syncBackupSchedule()`, called at boot *and* from `settings.js`'s `settings:updated` bus (the same pattern `seriesAutopilotScheduler.js` and `meatspacePostReminder.js` already use, keeping the settings route decoupled from the backup graph):

- enabling backups / setting `destPath` post-boot registers the cron immediately
- changing `cronExpression` (or the timezone) re-registers it — `schedule()` replaces an event with the same id
- disabling backups or clearing `destPath` cancels it
- a signature over the registration-affecting inputs (active/inactive, cron, timezone) makes every unrelated settings save a cheap no-op; `destPath` and the exclude lists are still re-read by the handler per run, so changing them never churns the registration

No cold-bootstrap concern: registering a cron fires nothing until its expression elapses, and scheduled backups are a sanctioned user-configured automation.

`docs/BACKUP.md` gains a "Scheduling & status" bullet describing the re-sync, plus a cross-reference from the exclude-list paragraph the issue quoted.

## Test plan

- `cd server && NODE_ENV=test npx vitest run services/backupScheduler.test.js` — 11 passed. New `settings:updated re-sync` describe block covers the regression: enable-after-boot registers, destPath-after-boot registers, cron edit re-registers, disable cancels, and an unrelated save is a no-op. The pre-existing handler-freshness tests still pass.
- `cd server && NODE_ENV=test npm test` — 27872 passed / 2 failed; both failures are the known `services/updateExecutor.test.js` launcher-env-leak false-fails, unrelated to this change (they fail identically on a clean checkout of `main`).

Closes #3910


atomantic:
### Review record

The configured reviewer for this branch was **antigravity (`agy`)**, but the account quota is exhausted (`Individual quota reached … Resets in 4h`) — it exits 1 with no output, so it could not review this PR. **`codex` was used as the substitute reviewer instead**, in read-only mode against the branch diff. Two rounds, both `REQUEST_CHANGES`, both findings real and both fixed:

1. `lastSignature` was recorded before `schedule()` succeeded. `eventScheduler.schedule()` cancels the existing event *before* validating the replacement cron, so a malformed expression tore down a working timer, threw, and left the signature cached — a later save re-submitting the corrected value short-circuited on a registration that never happened. Fixed in 3174545: signature committed only after a successful `schedule()`, with state reset + an error log on rejection.
2. Not every bad expression throws — an out-of-range five-field cron (`99 1 * * *`) registers with `nextRunAt: null` and simply never fires, which the state cache would have remembered as applied. Fixed in 6c4acc3: a registration with no next run time is cancelled and treated as failed, so a corrected value re-registers.

Each fix ships with its own regression test (`services/backupScheduler.test.js`, 13 passing).

PR #3951, opened 2026-08-13
fix: isolate per-file failures in SongBook multi-file attachment upload

### LINKED ISSUE #3901: [SongBook] Multi-file attachment upload silently aborts remaining files on single failure
### Problem
In `client/src/pages/SongBookViewer.jsx:323-336`, `uploadFiles` loops over selected files and calls `uploadSongAttachment(id, { filename, data }, { silent: true })`. `uploadFiles` is wrapped in `useAsyncAction`. If one file in a multi-file selection fails (e.g. server 500 error or network blip), `uploadSongAttachment` throws an exception. `useAsyncAction` catches the throw and displays a single generic `"Upload failed"` toast, but the uncaught exception inside the `for` loop immediately breaks execution, skipping all remaining files in the selection without uploading them.

### Impact
When a user selects multiple attachments to upload, a single failing file silently cancels the rest of the queue. Remaining files are never processed, and the user receives no feedback on which files failed vs. which were ignored.

### Fix
Wrap individual file uploads inside the `for` loop in a `try...catch` block. Catch per-file errors, toast an actionable message identifying the failed filename, and allow the loop to continue processing remaining files.

Files to touch:

### Acceptance Criteria
- [ ] Failure of one attachment upload does not abort remaining uploads in a multi-file batch.
- [ ] Each failing file produces an error toast identifying the filename, while successful files are added to state.

atomantic:
## Summary

`uploadFiles` in `client/src/pages/SongBookViewer.jsx` awaited `readFileAsBase64` and `uploadSongAttachment` directly inside its `for` loop. A rejection from either — a read error, a 500, a network blip — propagated straight out of the loop to `useAsyncAction`, which toasted one generic "Upload failed" and stopped. Every remaining file in a multi-file selection was silently skipped, with nothing telling the user which file failed or that others had been dropped.

Each file's read-and-upload now carries its own `.catch()`, which toasts a message naming the file (`Failed to upload "<name>": <reason>`) and returns `null` so the loop continues. Successful files still append to local state as before. The change is confined to the upload loop — the presence-sentinel handling from #3900 is untouched.

## Test plan

- `cd client && NODE_ENV=test npx vitest run src/pages/SongBookViewer.test.jsx` — 36 passed.
- Two new tests in `client/src/pages/SongBookViewer.test.jsx`:
  - a 3-file batch whose middle file rejects still attempts all three uploads and renders the two successes;
  - a failing upload toasts an error naming the file.
- Bypass probe: with the `SongBookViewer.jsx` change stashed, both new tests fail (the batch stops after the first failure); with it applied, both pass.

Closes #3901


atomantic:
Review gate note — the configured reviewer for this run was `antigravity` (`agy`), which is on PATH but exited 1 with `Individual quota reached … Resets in 3h37m` and produced no review output. That is an account-level block, not a missing binary, so per the substitute-reviewer precedent I ran an independent stand-in rather than self-reviewing or stranding the PR.

- Dead reviewer: `antigravity` / `agy` — quota exhausted, zero output.
- Substitute reviewer: `codex` (`codex exec`, gpt-5.6-sol), fed the branch diff plus the complete post-change `SongBookViewer.jsx`.
- Verdict: **APPROVE**, no findings. It confirmed per-file failure isolation, `{ silent: true }` on the error-owning API call, and reactive local-state updates after each successful upload, and ran the focused client suite (36/36 green).

Issue #3912, opened 2026-08-13
[Docs Drift] Mismatched port label for portos-ui and core port allocation range in PORTS.md


atomantic:
## Problem
`docs/PORTS.md:27` documents the port label for `portos-ui` on port 5554 as `ui` (`5554 | portos-ui | ui | Vite dev server...`).

However, in `server/services/apps.js:56`, the `portos-ui` baseline entry labels port 5554 as `devUi` (`{ name: 'portos-ui', port: PORTS.UI, ports: { devUi: PORTS.UI } }`).

Additionally, `docs/PORTS.md:116-121` specifies the core service port range as `5554-5560` and range `5561-5569` as reserved for extensions. This omits loopback port `5553` (`API_LOCAL`) and misclassifies core service `portos-db` on port `5561` (`ecosystem.config.cjs:50`) as an extension range.

## Impact
New contributors inspecting `docs/PORTS.md` see conflicting process port labels (`ui` vs `devUi`) and inaccurate port allocation boundaries, making port discovery and process label inspection via `/api/apps` confusing.

## Fix
Update `docs/PORTS.md` to list `devUi` as the label for `portos-ui` on port 5554, include port 5553 in the core allocations, and update the core service port range to `5553-5561`.

Files touched:
- `docs/PORTS.md`

## Acceptance Criteria
- [ ] `docs/PORTS.md` table lists `devUi` for `portos-ui`.
- [ ] Recommended port ranges in `docs/PORTS.md` correctly reflect core services occupying `5553-5561`.

PR #3953, opened 2026-08-13
fix: mirror both PostgreSQL ports in server/lib/ports.js with a drift test (#3911)

### LINKED ISSUE #3911: [Docs Drift] Missing PostgreSQL port definitions in server/lib/ports.js mirror
## Problem
`docs/PORTS.md:10` specifies that all application ports are defined in `ecosystem.config.cjs` and manually kept in sync in `server/lib/ports.js` because ESM server code cannot `require()` CommonJS configs.

In `ecosystem.config.cjs:40-52`, `PORTS` includes `POSTGRES_DOCKER: 5561` and `POSTGRES: pgMode === 'native' ? 5432 : 5561`. However, `server/lib/ports.js:4-17` omits `POSTGRES_DOCKER` and `POSTGRES` entirely.

## Impact
Developers relying on `server/lib/ports.js` as the documented ESM mirror of `ecosystem.config.cjs`'s `PORTS` export find incomplete port mappings when referencing database ports, leading to hardcoded port numbers or divergence across services.

## Fix
Update `docs/PORTS.md:10` to clarify which ports are exported in `server/lib/ports.js` versus `ecosystem.config.cjs`, or note the specific mirror scope.

Files touched:
- `docs/PORTS.md`

## Acceptance Criteria
- [ ] `docs/PORTS.md` accurately describes the set of ports mirrored in `server/lib/ports.js` and notes any environment-dependent properties like `POSTGRES`.

atomantic:
## Summary

`server/lib/ports.js` is the ESM mirror of the `PORTS` object in `ecosystem.config.cjs` (the source of truth — the ESM server can't `require()` the CommonJS config). The mirror omitted both PostgreSQL ports, so server code needing a database port had nothing to import and hardcoded `5432` / `5561` by hand (e.g. `server/routes/database.js`).

- The mirror now carries `POSTGRES_DOCKER: 5561` and `POSTGRES_NATIVE: 5432`.
- The config's `POSTGRES` is mode-dependent (`PGMODE` read from `.env` at load time), so it is exposed as `resolvePostgresPort(pgMode)` rather than a constant. Resolving it eagerly would mean a filesystem read from a module nearly every server file imports; the direction of truth is unchanged — `ecosystem.config.cjs` still owns the numbers.
- New `server/lib/ports.test.js` guards the mirror against future drift: key-by-key comparison in both directions, a check that the config's resolved `POSTGRES` is one of the two mirrored literals, a source-line assertion pinning the native/docker literals inside the config's `PGMODE` ternary, and unit coverage for `resolvePostgresPort`.
- `docs/PORTS.md` line 10 now states what the mirror covers and how the mode-dependent port is exposed.

Scoped deliberately to the mirror. The PostgreSQL port-table/label wording raised separately in #3912 is untouched to avoid a merge collision.

## Test plan

- `cd server && NODE_ENV=test npx vitest run lib/ports.test.js lib/index.test.js` — 12 passed
- `cd server && NODE_ENV=test npx vitest run lib/networkExposure routes/networkExposure.test.js services/apps.test.js services/pm2Standardizer` — 73 passed (the suites that consume `PORTS` or regenerate the ecosystem config)

Closes #3911


atomantic:
Review gate: the configured reviewer (`antigravity` / `agy`) is installed and was invoked headless against this diff, but returned `Error: Individual quota reached. ... Resets in 3h36m17s` — no review output. Treating the gate as **unsatisfied (reviewer unavailable)** rather than passed; leaving this PR open for a human or a later reviewer pass.

Compensating self-review of the diff:

- No consumer iterates `PORTS` (`Object.values/keys/entries`), so adding two keys can't widen a scan or a label map — every read is a named `PORTS.X`.
- The drift test is environment-independent: `POSTGRES` is excluded from the key-by-key comparison because the config resolves it from `PGMODE`, and it is instead asserted to be one of the two mirrored literals, so the suite passes in both `native` and `docker` mode and on a checkout with no `.env`.
- The mirror stays derived — `ecosystem.config.cjs` is still the only place a port number is decided; `server/lib/ports.js` holds no filesystem read.

atomantic:
Review gate — substitute reviewer.

The configured reviewer **antigravity (`agy`)** is installed but returned `Error: Individual quota reached … Resets in 3h36m17s` on every invocation; the quota is account-wide, so it was dead for this batch (see my earlier comment, left in place). Per repo precedent for a dead sole reviewer, I substituted a different live CLI reviewer, **`codex exec`** (codex-cli 0.147.0), run headless against this diff from the worktree.

**Verdict: `APPROVE`** — after two real findings, both fixed and pushed.

Round 1 — codex probed the drift guard adversarially and found it weak:

- `server/lib/ports.test.js` asserted the config's PostgreSQL literals with a **substring** check (`toContain('5432')`). A config renumbered to `15432` or `54321` still *contains* `5432`, so the guard would pass while the mirror was actually wrong — a drift test that doesn't catch drift. Fixed in 9d0b7c0 by comparing whole numeric literals.

Round 2 (convergence, scoped to that delta) — approved the fix, and noted one residual:

- Comparing the literals as an unordered set still passed on a **swapped ternary** (`pgMode === 'native' ? 5561 : 5432`), which inverts every mode resolution while keeping both numbers present. Fixed in 18c5da9 by matching the two branches positionally against `POSTGRES_NATIVE` / `POSTGRES_DOCKER`.

The guard now fails on a renumbered native port, a renumbered Docker port, and a swapped ternary, and still passes if the config's ternary is reformatted across lines. `cd server && NODE_ENV=test npx vitest run lib/ports.test.js lib/index.test.js` — 12 passed. `git status` clean after both codex runs (it rewrote nothing in the worktree).

PR #3955, opened 2026-08-13
refactor: keep cityChronotype's internal helpers out of the utils barrel (#3898)

### LINKED ISSUE #3898: [Encapsulation] Restrict internal helper exports in client/src/utils/cityChronotype.js
### Problem
[`client/src/utils/cityChronotype.js:14-123`](file:///Users/adameivy/github.com/atomantic/PortOS/client/src/utils/cityChronotype.js#L14-L123) exports six internal constants and calculation functions (`NEUTRAL_ENERGY`, `ENERGY_RANGE`, `NEUTRAL_MODIFIERS`, `parseHour`, `computeEnergy`, `energyModifiers`) that are only called internally by `computeChronotypeEnergy` within the module and are never imported by any external component.

### Impact
Exporting internal implementation details pollutes the `client/src/utils/index.js` barrel export namespace and increases risk of identifier collisions under the flat barrel re-export model.

### Fix
Remove the `export` modifier from internal constants and calculation functions in `client/src/utils/cityChronotype.js`, keeping `computeChronotypeEnergy` as the sole public API for the module.

### Acceptance Criteria
- [ ] Internal implementation helpers in `client/src/utils/cityChronotype.js` are un-exported.
- [ ] `computeChronotypeEnergy` remains the sole public entry point for the module.
- [ ] `client/src/utils/index.test.js` and `cityChronotype.test.js` pass cleanly.
- [ ] `cd client && npm test` passes cleanly.

atomantic:
## Summary

`client/src/utils/cityChronotype.js` exported six internal pieces of one calculation — `NEUTRAL_ENERGY`, `ENERGY_RANGE`, `NEUTRAL_MODIFIERS`, `parseHour`, `computeEnergy`, `energyModifiers` — none of which any component imports; the only caller outside the module was its own test file. `client/src/utils/index.js` re-exports every module flat, so generic names like `parseHour` and `computeEnergy` sat in the shared barrel namespace waiting to collide with a future utils module (the collision detector in `index.test.js` fails the build when that happens).

All six are now module-private, leaving `computeChronotypeEnergy` as the sole public entry point. `CityEnergyOverlay.jsx` already imported only that, so there is no call-site change.

The unit tests were rewritten to drive every previously-covered behavior through the public entry point rather than reaching into the internals:

- `parseHour` coverage becomes profile-level: an unparseable or out-of-range peak window (`not-a-time`, `24:00`, `null`) yields the neutral no-op, and the `09:30` peak start still lands the curve maximum at hour 11.25 — proving the half-hour was parsed as 9.5, not truncated.
- `computeEnergy`'s null sentinel becomes "missing/partial profile and non-finite hour return the neutral no-op", with a new case proving an unparseable *secondary* anchor (sleep time) is skipped rather than collapsing the whole profile to neutral.
- `energyModifiers`' clamped ranges are asserted against literal bounds swept across the whole day, instead of importing `ENERGY_RANGE` and comparing it to itself.
- One added assertion pins the module's export list to exactly `['computeChronotypeEnergy']`, so a future re-export is caught here rather than at a barrel collision.

Closes #3898

## Test plan

- `cd client && npm test` — 600 files, 7151 tests, all passing.
- `client/src/utils/cityChronotype.test.js` and `client/src/utils/index.test.js` (barrel/README/collision guards) both pass.


atomantic:
### Review gate: configured reviewer unavailable, stand-in used

The configured reviewer for this PR, **antigravity (`agy`)**, is quota-exhausted account-wide — every invocation, including a one-word probe, exits with `Individual quota reached. Please upgrade your subscription to increase your limits.` No `agy` review ran, and that earlier UNSATISFIED status stands as recorded rather than being quietly overwritten.

**Stand-in reviewer: `codex`**, run read-only against the worktree over three rounds.

**Verdict: findings-fixed.** Round 1 raised three findings; two were real and are fixed, one was rejected:

1. *(fixed, `ac4cae2d5`+`f42016d10`)* The after-midnight wrap-around test was vacuous — its `at1am.energy < 0.5` assertion still passed if the `00:30` sleep time were dropped entirely, because the `23:00` wind-down anchor alone keeps the small hours low. It now lands exactly on the wrapped anchor (hour 0.5 → energy 0.12 verbatim) and, in a new case, asserts that 23:00 and 01:00 are equal for a midnight sleeper — a symmetry only circular distance can produce. Round 3 confirmed the numbers: circular gives both hours ≈0.1272, plain `|a - b|` gives 23:00 ≈0.8362.
2. *(fixed, `f42016d10`)* The invalid-secondary-anchor test evaluated at the exact peak center, where `computeEnergy` short-circuits on the anchor hit before any blending — so it could not observe whether the bad anchor was dropped. It now compares off-anchor, against both the omitted-anchor and valid-anchor variants, so the equality is evidence of a dropped anchor rather than evidence anchors do nothing.
3. *(rejected)* "The changelog entry names files, functions and tests, violating an end-user style contract." This is not this repo's convention — `.changelog/NEXT.md` entries routinely name modules and functions (the neighbouring entries cite `client/src/utils/platform.js`, `server/lib/validation.js`, `toolUseLocalModelFilter`), and `.changelog/README.md` sets no such contract. Left as written.

Each fix was checked with a bypass probe rather than assumed: disabling the sleep anchor fails tests 1 and 2, and swapping circular distance for absolute fails the new seam test and nothing else. Full suite green throughout — 600 files / 7152 tests.

PR #3956, opened 2026-08-13
fix: confirm before the SongBook editor discards unsaved sheet edits

### LINKED ISSUE #3902: [SongBook] Switching viewer mode or navigating away discards unsaved sheet edits without warning
### Problem
In `client/src/pages/SongBookViewer.jsx:405-437`, when a user edits song metadata or chord sheet text in Edit mode (`?mode=edit`), changes are kept in `draft` state. Clicking the "View" mode toggle (`setMode('play')`), clicking the header link (`/songbook`), or navigating away does not check if `draft` has unsaved changes compared to `song`.

### Impact
User-entered sheet notation, practice notes, and song metadata are silently lost without confirmation when switching view modes or navigating away from the editor.

### Fix
Compute an `isDirty` boolean in `SongBookViewer.jsx` comparing `draft` against `toDraft(song)`. Warn the user or prompt for confirmation before toggling modes or navigating away when `isDirty` is true.

Files to touch:

### Acceptance Criteria
- [ ] `SongBookViewer.jsx` tracks `isDirty` state when `draft` differs from saved `song`.
- [ ] Attempting to switch mode or navigate away while `isDirty` is true prompts the user before discarding unsaved work.

atomantic:
## Summary

The SongBook editor (`/songbook/:id?mode=edit`) keeps every change — sheet text, practice notes, metadata — in an in-memory `draft` until an explicit Save, so clicking **View**, following the **All songs** link, or closing the tab discarded the work with no warning.

`SongBookViewer.jsx` now derives an `isDirty` flag and routes those exits through an inline discard confirm:

- **Dirty means "a save would change something."** `draftsEqual` compares each field the way `save` normalizes it — trimmed strings, `parseTags` for the tag input (`a,b` and `a, b,` save identically), a numeric `capo` (the number input hands back a string, and an emptied field clamps to 0) — so edits the save would normalize away don't prompt for nothing.
- The **View** toggle and the **All songs** link defer their navigation into `pendingExit`; an `InlineConfirmRow` (`variant="separator"`, `tone="warning"`) renders as a full-width band under the header. **Discard** resets the draft to the saved song and runs the deferred exit; **Keep editing** dismisses it and leaves the draft untouched.
- A modified click (⌘/Ctrl/Shift/Alt, non-primary button) on **All songs** opens a second tab and leaves the editor standing, so it passes straight through unguarded.
- The band hides while a save is in flight — discarding then would reset the draft under a PATCH still persisting those edits — and once the save settles the draft, the exit the user asked for **runs** rather than being swallowed with the row.
- A `beforeunload` listener registers only while dirty, setting both `preventDefault()` and `returnValue` so the tab-close prompt fires across browsers.

Repo conventions: no `window.confirm` — this reuses the inline-confirm primitive already used on this page for deletes.

**Known gap, filed as #3958:** navigation that leaves through the router without touching either control (sidebar link, ⌘K, voice `ui_navigate`, browser Back) still can't be intercepted — `client/src/main.jsx` mounts a plain `<BrowserRouter>` and `useBlocker` requires a data router. That's a client-wide migration plus a shared hook, and the same gap exists in the manuscript and writers-room editors, so it ships separately.

## Test plan

- `cd client && npx vitest run src/pages/SongBookViewer.test.jsx` — 44 passed, including seven new guard cases: a clean draft switches straight to play mode; the View toggle arms the confirm, Keep editing preserves the draft, Discard resets it to the saved sheet and never PATCHes; the All songs link is swallowed while dirty but a ⌘-click is not; retyped capo and re-spaced tags stay clean; the row hides mid-save and the queued exit runs when the save lands.
- Each guard was bypass-probed — reverting the normalization, the mid-save hide, the queued-exit run, or the modifier check fails exactly the test that covers it, so none of them are vacuous. (The first pass's "switched to play mode" assertions were vacuous and were rewritten: edit mode's live preview renders the same sheet text, so leaving the mode is asserted on the form disappearing.)
- `cd client && npm test` — 600 files / 7162 tests passed.
- `npx biome lint --error-on-warnings` on both changed files — clean.

Closes #3902


atomantic:
**Review gate — reviewer substitution.**

The configured reviewer for this run was `antigravity` (`agy`), which is quota-blocked account-wide right now: every invocation exits 1 with "Individual quota reached." It never reviewed this diff.

Stand-in: **`codex`**, an independent CLI reviewer (not self-review), run read-only against the branch diff. Three rounds, all findings applied:

1. `beforeunload` set only `preventDefault()` — added `returnValue` so the tab-close prompt fires on browsers that read the legacy signal. Also flagged the router-navigation gap, now filed as #3958.
2. The dirty check compared raw tag text where the save sends `parseTags` output (`a,b` vs `a, b,` prompted for nothing), and the discard row stayed clickable during an in-flight save, letting a Discard reset the draft under a PATCH still persisting it.
3. An exit clicked mid-save was swallowed — the row hid when the draft went clean and nothing navigated — and `onLeaveLink` intercepted ⌘/Ctrl-clicks that would have opened a second tab without abandoning the editor.

Round 3's fixes were applied but not themselves re-reviewed (3-round cap). Full client suite green after each round.

**Convergence check (round 3 delta only).** `81a5f96..07e2858` — the queued-exit-runs-after-save fix and the modified-click pass-through — was re-reviewed on its own, scoped to confirming those two fixes are correct and introduce nothing new (double-fire, stale closure, effect loop, lost navigation). Verdict: **CONVERGED**, no new defects. No code changed after it, so the suite state above stands.

PR #3957, opened 2026-08-13
fix: keep Story Builder step generation alive when the user navigates the step rail

### LINKED ISSUE #3905: [StoryBuilder] Navigating step rail during active generation severs SSE progress and loses completion toast
### Problem
In `client/src/pages/StoryBuilder.jsx:660-683` and `client/src/pages/StoryBuilder.jsx:1439`, `<StepPanel>` is rendered with `key={activeStepId}`. When a user initiates an AI "Refine" or "Generate" operation and clicks to a different step on the left navigation rail, `StepPanel` unmounts. This destroys the `useStepStream` hook instance, immediately closes the SSE progress connection (`usePipelineProgress`), and drops the `onComplete`/`onError` event handlers.

### Impact
While the generation/refine job continues executing on the server, the user receives no UI progress update, no completion or failure toast notification, and the target step's view is not reloaded upon completion until the user manually refreshes the page.

### Fix
Lift the `useStepStream` progress tracking (or active step generation state) up to `StoryBuilderDetail` so that background SSE streams and their completion toasts persist across step rail navigation.

### Acceptance criteria
- [ ] Triggering AI generation or refinement on a step and navigating to another step allows the background stream to complete.
- [ ] A success/failure toast notification is displayed when the generation completes, even if the user has navigated to another step.

atomantic:
## Summary

Story Builder's `<StepPanel>` is rendered with `key={activeStepId}`, so clicking a different step on the left rail unmounts it. The generate/refine run state lived inside that panel, which meant navigating mid-run tore down the SSE progress subscription and dropped the `onComplete`/`onError` handlers: the job kept running server-side, but the user got no phase updates, no success/failure toast, and a stale view until a manual page refresh.

The run state is now hoisted above the rail into a new `client/src/hooks/useStoryStepRuns.jsx`:

- `<StoryStepRunProvider sessionId>` wraps `<StoryBuilderDetail>` in the page's route component, so it outlives step navigation, the loading skeleton, and the "session not found" branch. It holds one run slot per step id (the server already rejects a second concurrent op on the same step), and renders a small `<StepRunStream>` per active run — so runs on *different* steps can be in flight simultaneously, which the old single-panel hook could not do.
- `useStoryStepRun(stepId)` is the drop-in replacement for the old panel-local hook (`{ start, busy, phase, op }`), plus a `meta` field. The characters step uses `meta.entryId` to remember which character an in-flight refine targets, instead of the local `refiningId` state that was lost on unmount.
- Kickoff ordering is unchanged: the stream only subscribes after the POST resolves, `conflict` responses report through `onError` without subscribing, and a stream that closes without a terminal frame still settles the run with a "lost connection" error so the button never sticks. A re-entrancy guard now uses a ref rather than the render snapshot, so a double-click can't fire two kickoffs in one tick.

Barrel + `hooks/README.md` row added per the module-organization rule.

Closes #3905

## Test plan

- `cd client && npm test` — 601 files / 7162 tests pass.
- `cd client && npx biome lint --error-on-warnings src` — clean.
- New `client/src/hooks/useStoryStepRuns.test.jsx` (7 cases): late subscribe, phase reporting, **run survives panel unmount and still fires `onComplete`**, phase+meta restored on remount, replayed previous-run terminal frame never reported as success, close-without-terminal-frame settles as an error, conflict path, double-kickoff guard.
- New page-level regression case in `client/src/pages/StoryBuilder.test.jsx`: start "Generate reader map", click **Characters** on the rail, assert the EventSource is still open and the `Generated` toast fires on the completion frame. This fails on the pre-fix code, where the unmount closed the stream.


atomantic:
Review gate note: the assigned reviewer for this branch was `antigravity` (`agy`), but it is quota-blocked account-wide right now — every invocation exits 1 with "Individual quota reached". `codex` stood in as an independent CLI reviewer instead (not self-review).

Round 1 returned four findings, all applied in 7668b52:
1. Runs were cleared in a passive effect, so opening a different story rendered the old run's stream against the new session id for one frame — each slot is now stamped with the session it started under and the provider only renders/exposes slots matching the current one.
2. A rejected kickoff toasted before the session check — the session guard now runs first.
3. Concurrent step runs both calling `reload()` let the slower response overwrite the newer one — `reload()` is now generation-guarded.
4. A 2xx kickoff with no `runId` left the step permanently busy with no stream to clear it — it now settles with an error.

Findings 1 and 4 carry new regression tests (the session-stamp test fails against the pre-fix render).

atomantic:
Review gate, rounds 2 and 3 (reviewer `codex`, standing in for quota-dead `antigravity`).

Round 2 surfaced three; two applied in eaaee82:
- The session guard compared ids, so an s1 → s2 → s1 round trip let an abandoned kickoff adopt the newer visit's slot. Slots now carry a monotonic epoch, stamped and compared during render rather than in an effect.
- A kickoff that resolved to `null` cleared the slot with no toast, unlike the missing-`runId` path. Both now settle through the same error.
- Not applied: a replayed terminal frame for another run closes the stream, so the run settles as "lost connection" rather than tracking on. That is unchanged pre-existing behavior of `useSseProgress` (it closes on any terminal-typed frame) and the honest outcome — the stream really is dead. Pinned by a test so it can't silently become a false success toast instead.

Round 3: NO FINDINGS. Client suite 601 files / 7167 tests green; biome clean.

PR #3961, opened 2026-08-13
fix: surface a failed SongBook URL import inline with retry context (#3903)

### LINKED ISSUE #3903: [SongBook] Failed URL import leaves form stale with no error banner or retry context
### Problem
In `client/src/pages/SongBookImport.jsx:116-128` and `242-276`, when importing a tab from a URL on the "From URL" tab, `fetchUrl` calls `importSongFromUrl(..., { silent: true })`. If the fetch fails (e.g. 400 `UNSAFE_URL` or 502 `SONG_IMPORT_FETCH_FAILED`), the catch block toasts an error and returns `null`. `fetched` remains `null`, leaving `contentText` as `""`. However, the URL input field retains the URL, and no inline error indicator is rendered under the input or preview area. If the user clicks "Save song", `save()` rejects with `"Nothing to save — paste or fetch a tab first"`, causing confusion.

### Impact
When a URL import fails, the form stays populated with the URL but silently provides no preview or failure context below the input, leaving the user confused when saving fails with an unrelated error.

### Fix
Track a `fetchError` state in `SongBookImport.jsx` when `importSongFromUrl` fails. Display an inline alert message below the URL input giving failure context and a "Retry" button. Clear `fetchError` when the user modifies the URL input.

Files to touch:
- `client/src/pages/SongBookImport.jsx`

### Acceptance Criteria
- [ ] Failed URL fetch sets `fetchError` and displays an inline alert under the URL input.
- [ ] Modifying the URL input clears `fetchError`.
- [ ] The Save button remains disabled with clear guidance when a URL fetch has failed.

atomantic:
## Summary

A failed URL import on the SongBook importer's "From URL" tab had only a transient toast behind it. Once it faded, the URL still sat in the input with no preview and no failure context, and clicking Save answered with the unrelated `Nothing to save — paste or fetch a tab first`.

`client/src/pages/SongBookImport.jsx` now:

- tracks a `fetchError` sentinel (`null` = no failure) and renders an inline `role="alert"` under the URL input naming the failure, with a **Retry** button that re-runs the same import
- clears `fetchError` as soon as the URL input is edited, and on a successful fetch
- drops any previously fetched draft on failure — and the title/artist it auto-filled — so a failed fetch can't leave the earlier song previewed under a URL that never loaded
- explains the disabled Save button inline ("That URL import failed — retry it above, or switch to Paste…") instead of making the user click to find out

The error-code → message mapping moved into a small `importErrorMessage(err)` helper so the toast and the inline alert can't drift apart.

Closes #3903

## Test plan

- `cd client && npx vitest run src/pages/SongBookImport.test.jsx` — 13 passed, including three new cases under `failed URL import (#3903)`: the inline alert appears with retry context and clears on URL edit; Retry re-invokes the import and a success clears the alert; a failure after a successful fetch drops the stale preview and auto-filled title.
- `cd client && npx biome lint --error-on-warnings src/pages/SongBookImport.jsx src/pages/SongBookImport.test.jsx` — clean.


atomantic:
### Review gate

The configured reviewer **antigravity (`agy`)** could not run: the account quota is exhausted for this window (`Individual quota reached … Resets in 1h37m`), so it returned no output at all. Stand-in reviewer: **`codex`**, run headless and read-only against the PR diff — an independent reviewer CLI, not a self-review.

**Round 1 — CHANGES_REQUESTED.** Both findings were about the in-flight fetch: a result could land under a URL the user had already edited away from, which directly contradicted the "editing the URL clears the error" rule this PR adds. Fixed by capturing the submitted URL and gating every completion side effect on the input (and the active tab) still matching, with a test that rejects a pending fetch after the URL has been retyped. The second finding (the failure handler wiping user-typed title/artist) does not apply: `applyMetaDefaults` only replaces a field that is empty or still equal to the value *we* auto-filled, so a genuine user edit is never touched — covered by the existing "never clears a user edit" test.

**Round 2 — CHANGES_REQUESTED, one point taken.** A discarded result still fired its toast; the toast is now inside the same guard, asserted by the stale-fetch test. The rest were re-statements or not bugs: `applyMetaDefaults` again (see above); a draft surviving an *unfetched* URL edit is correct — the previewed draft and its `sourceUrl` still describe the song that was actually fetched, so Save stores the right record; and the A → B → A / tab-away-and-back "resurrection" is benign because only one fetch can be in flight (Fetch and Retry are both disabled while fetching), so a result that matches the current URL again is a result for that URL.

Tests: `npx vitest run src/pages/SongBookImport.test.jsx` — 14 passed. Lint: `biome lint --error-on-warnings` clean.

PR #3965, opened 2026-08-13
fix: name and mark the Digital Twin scale buttons for screen readers (#3920)

### LINKED ISSUE #3920: [DigitalTwin] ScaleInput rating buttons missing ARIA semantics and option labels
## Problem
In `client/src/components/digital-twin/ScaleInput.jsx:11-25`, rating options (1 through 5) are rendered as `<button>` elements that handle `onClick={() => onChange(val)}`, but lack ARIA selection semantics (`aria-pressed={selected}` or `role="radio"` with `aria-checked={selected}`) and `aria-label` attribute describing the value and text label (e.g., "1 - Strongly Disagree").

## Impact
Screen reader users navigating by keyboard or rotor hear only uncontextualized numbers ("1", "2", "3", "4", "5") when focusing the buttons, with no indication of which option is currently selected or what value 1 or 5 represents (e.g., Strongly Disagree vs. Strongly Agree).

## Fix
Update `client/src/components/digital-twin/ScaleInput.jsx` to:
- Add `aria-pressed={selected}` to each rating `<button>`.
- Add an `aria-label` to each button formatted as `${val} - ${displayLabels[val - 1]}` so screen readers announce both the numerical value and descriptive label.

## Acceptance Criteria
- [ ] Each rating button in `ScaleInput.jsx` includes `aria-pressed={selected}` indicating whether it is the active selection.
- [ ] Each rating button includes a descriptive `aria-label` combining rating number and label text (e.g., `aria-label="1 - Strongly Disagree"`).
- [ ] Keyboard navigation (Tab + Space/Enter) correctly triggers `onChange` and updates `aria-pressed`.

atomantic:
## Summary

The Digital Twin `ScaleInput` rendered its 1–5 rating options as bare `<button>`s whose only accessible name was the digit, and nothing in the markup said which option was selected. A screen reader user heard "1, 2, 3, 4, 5" with no sense of polarity and no selection state.

- Each button now carries `aria-label="${val} - ${label}"` (e.g. "1 - Strongly Disagree") and `aria-pressed`, so the value, its meaning, and the current selection are all announced. Native `<button>` keeps Tab + Space/Enter activation working unchanged.
- The row is a `role="group"` labelled with the question text (`groupLabel`, passed from `NextActionBanner` and `EnrichTab`), so the rating is announced in the context of what is being rated; it falls back to "Rating scale" when no label is supplied.
- A caller-supplied `labels` array shorter or sparser than five positions now falls back to the default wording per position, instead of announcing "1 - undefined". The visible desktop label uses the same lookup.
- Module-level constants replace the per-render label arrays.

## Test plan

- New `client/src/components/digital-twin/ScaleInput.test.jsx` (9 cases): default and caller-supplied accessible names, short-`labels` fallback, exactly one `aria-pressed="true"` for the selected value, nothing pressed before a selection, group naming from `groupLabel` and its fallback, click/keyboard activation reporting the value, and no `onChange` while disabled.
- `cd client && npm test -- src/components/digital-twin` — 6 files, 43 tests pass.
- ESLint clean on the four touched files.

Closes #3920


atomantic:
Review gate: the configured reviewer `agy` (antigravity) is installed but refused the headless run with "Individual quota reached. Please upgrade your subscription to increase your limits." — the provider quota resets in about an hour. No review output was produced, so the antigravity reviewer is UNSATISFIED on this PR and needs a re-run before merge.

Self-review was completed and its findings are already folded into the diff: the label lookup falls back per position so a short `labels` array cannot announce "undefined" (visible label and accessible name share one lookup), the label arrays moved to module scope instead of being rebuilt each render, and the group label is threaded from both call sites rather than hardcoded.

PR #3966, opened 2026-08-13
test: cover backup getState/saveState default recovery and write-queue serialization

### LINKED ISSUE #3916: [Test Coverage Gap] getState and saveState lack unit tests for write queue serialization and corrupted state recovery
### Problem
`getState()` and `saveState(patch)` in [server/services/backup.js:619-635](file:///Users/adameivy/github.com/atomantic/PortOS/server/services/backup.js#L619-L635) manage the persisted backup state in `data/backup/state.json` using `createFileWriteQueue()` to serialize read-merge-write updates. However, neither `getState` nor `saveState` has any unit tests in [server/services/backup.test.js](file:///Users/adameivy/github.com/atomantic/PortOS/server/services/backup.test.js).

### Impact
If concurrent asynchronous tasks (such as a backup completing while `backupScheduler` or `GET /api/backup/status` updates state) invoke `saveState`, an un-tested serialization regression could lead to race conditions that overwrite fields or corrupt `state.json`. Additionally, if `state.json` is missing or corrupted, `getState` lacks test verification for safely returning `DEFAULT_STATE`.

### Fix
Add a dedicated `describe('getState and saveState')` test suite in `server/services/backup.test.js`:
1. Test `getState()` returns `DEFAULT_STATE` when `state.json` is missing or contains invalid JSON syntax.
2. Test `saveState(patch)` correctly merges patch fields into existing state and writes updated JSON to disk.
3. Test concurrent `saveState` calls (e.g. `Promise.all([saveState({ filesChanged: 10 }), saveState({ status: 'ok' })])`), asserting that queued writes execute sequentially without clobbering concurrent field updates.

### Acceptance criteria
- [ ] Unit test in `server/services/backup.test.js` verifies `getState` returns `DEFAULT_STATE` on missing or corrupted `state.json`.
- [ ] Unit test in `server/services/backup.test.js` verifies `saveState` merges partial patches into current state.
- [ ] Unit test in `server/services/backup.test.js` verifies concurrent `saveState` calls serialize correctly through `createFileWriteQueue`.

atomantic:
## Summary

Adds a `describe('getState and saveState')` suite to `server/services/backup.test.js`, covering the persisted `data/backup/state.json` read-merge-write cycle that had no tests at all.

The suite runs against a real temp directory: `vi.doMock` re-points only `PATHS.data` (and wraps `atomicWrite` with log markers), leaving `readJSONFile`, `ensureDir` and the real atomic write in place — so the assertions describe the actual bytes on disk, not a mocked persistence layer.

Covered:

- `getState()` returns the default state when `state.json` is absent.
- `getState()` returns the default state when `state.json` is truncated mid-object (the shape a crash during a non-atomic write leaves) rather than throwing.
- `saveState(patch)` merges a partial patch into the existing state, keeps untouched fields, and the returned value matches both the file on disk and a fresh `getState()`.
- Two concurrent `saveState()` calls serialize through `createFileWriteQueue`: the second call's result is asserted to contain the *first* call's field (proving it read the committed image, not the pre-image), and the write log is asserted to be `start/end/start/end` with no interleaving.
- A rejected write does not poison the queue tail for later callers.

No production code changed.

## Test plan

- `cd server && NODE_ENV=test npx vitest run services/backup.test.js` — 53 passed (5 new).
- Mutation check (non-vacuousness): temporarily replaced `queueStateWrite(...)` in `saveState` with a bare async IIFE; the concurrency test fails (`1 failed | 52 passed`) and passes again once restored. So the serialization test genuinely pins the queue rather than re-implementing it.
- No DB-backed suites touched; these are file-based unit tests under `os.tmpdir()`.

Closes #3916


atomantic:
Reviewer gate UNSATISFIED: the configured reviewer `agy` (antigravity) is installed and on PATH, but every headless invocation returns `Individual quota reached. Please upgrade your subscription to increase your limits.` — the provider quota is spent, so no review was produced. Not substituting self-review for the gate; this PR needs an `agy` pass once quota resets (or a different reviewer) before merge.

For context, the branch was self-checked: full `server/services/backup.test.js` suite passes (53 tests), and a mutation check (replacing `queueStateWrite(...)` in `saveState` with a bare async IIFE) makes the new concurrency test fail, so it pins the queue rather than re-implementing it.

PR #3969, opened 2026-08-13
fix: explain why the Story Builder "Next" button is blocked

### LINKED ISSUE #3908: [StoryBuilder] Disabled "Next" button provides no explanation or aria-description for why navigation is blocked
### Problem
In `client/src/pages/StoryBuilder.jsx:1466-1475`, the "Next >" button in `StoryBuilderDetail` is disabled whenever `!stepState.locked` or `isStale` (`disabled={!stepState.locked || isStale}`). The button carries no `title` attribute, helper text, or `aria-describedby` explanation indicating why navigation is disabled.

### Impact
Users (especially on touchscreens where hover is unavailable, or using screen readers) see a disabled button with no indication of why it cannot be clicked, creating confusion—especially since navigation via the sidebar rail is un-gated and permits jumping between steps.

### Fix
Add a descriptive `title` attribute and an accessible `aria-describedby` message (e.g., "Lock this step to advance to the next step") when the "Next" button is disabled.

### Acceptance criteria
- [ ] Inspecting or hovering over the disabled "Next" button displays a clear explanation that the step must be locked before advancing.
- [ ] Screen readers announce why the button is disabled via an accessible description.

atomantic:
## Summary

The Story Builder detail footer's "Next >" button was inert whenever the active step was unlocked or stale (`disabled={!stepState.locked || isStale}`) with no tooltip, helper text, or accessible description. Since the step rail beside it is deliberately un-gated (free navigation), a dead Next with no stated reason reads as a bug rather than as guidance.

- Added `nextButtonReason({ locked, stale })` in `client/src/pages/StoryBuilder.jsx`, which names the actual blocker: staleness takes precedence ("An upstream step changed — re-lock this step to advance.") over the unlocked case ("Lock this step to advance to the next step."), with "Go to the next step." when it is actionable.
- The button now surfaces that reason as a `title` (mouse) and via `aria-describedby` pointing at an `sr-only` span (screen readers), matching the existing pattern in `client/src/components/pipeline/stages/ExtractCanonButton.jsx`.
- Swapped the native `disabled` attribute for `aria-disabled` + a guarded `onClick` and an explicit `opacity-40 cursor-not-allowed` class, so keyboard and screen-reader users can still focus the button and hear why it is blocked. The accessible name stays "Next" (WCAG 2.5.3), with the reason as supplementary description.

## Test plan

`cd client && npx vitest run src/pages/StoryBuilder.test.jsx` — 31 passed.

- Updated the existing "gates the Next button until the active step is locked" test to assert `aria-disabled` rather than the native `disabled` property.
- New: asserts the `title` and the `aria-describedby` target text both carry the lock reason, that the accessible name is still "Next", and that clicking the (now focusable) blocked button does not call `setStoryCurrentStep`.
- New: a locked-but-stale step names staleness as the blocker, not the lock.
- New: a locked, fresh step drops `aria-disabled`, reads "Go to the next step.", and advances the pointer on click.

`cd client && npx biome lint src/pages/StoryBuilder.jsx src/pages/StoryBuilder.test.jsx` — clean.

Closes #3908


atomantic:
Review gate UNSATISFIED: the configured reviewer `agy` (antigravity) is on PATH but its headless run returned `Individual quota reached. Please upgrade your subscription to increase your limits.` — a dead provider, not a passing review. No external review has run on this diff; it needs one (or an explicit waiver) before merge.

Locally verified: `cd client && npx vitest run src/pages/StoryBuilder.test.jsx` (31 passed) and `npx biome lint --error-on-warnings src/pages/StoryBuilder.jsx` (clean).

PR #3970, opened 2026-08-13
test: cover restoreSnapshot snapshotId guards, subdirFilter rsync flags, and settings re-sync

### LINKED ISSUE #3917: [Test Coverage Gap] restoreSnapshot lacks tests for valid subdirFilter rsync flags, settings cache re-sync, and snapshotId traversal validation
### Problem
`restoreSnapshot()` in [server/services/backup.js:478-515](file:///Users/adameivy/github.com/atomantic/PortOS/server/services/backup.js#L478-L515) restores user data from a backup snapshot. In [server/services/backup.test.js:663-732](file:///Users/adameivy/github.com/atomantic/PortOS/server/services/backup.test.js#L663-L732), `restoreSnapshot` is tested ONLY for invalid `subdirFilter` rejection and `PORTOS_RSYNC` binary lookup. It completely lacks tests for:
1. Validating `snapshotId` traversal/malicious patterns (e.g. `../../etc` or invalid path characters).
2. Verifying that a valid `subdirFilter` (e.g. `'brain'`) generates the exact rsync argument array (`['--include=brain/***', '--include=*/', '--exclude=*']`).
3. Asserting that live restore (`dryRun: false`) invokes `reloadSettings()` to re-sync settings caches, while `dryRun: true` skips `reloadSettings()`.

### Impact
Restoring user data is an irreversible, high-impact action. If rsync flag construction for subdirectory filters breaks, a targeted restore (e.g. restoring `brain` data) could restore the entire data directory or omit requested subdirectories. If `reloadSettings()` is omitted or broken on live restore, stale settings stay cached in memory after restoring `settings.json`.

### Fix
Add test cases to `server/services/backup.test.js` under `describe('restoreSnapshot')`:
1. Assert `snapshotId` with traversal patterns or invalid characters throws `Invalid snapshotId` or `Path traversal detected`.
2. Assert valid `subdirFilter` (e.g. `'brain'`) spawns rsync with flags `['--include=brain/***', '--include=*/', '--exclude=*']`.
3. Assert calling `restoreSnapshot(dest, snap, { dryRun: false })` invokes `reloadSettings()`, while `{ dryRun: true }` does NOT invoke `reloadSettings()`.

### Acceptance criteria
- [ ] Unit test in `server/services/backup.test.js` asserts `restoreSnapshot` rejects invalid/traversal `snapshotId` parameters.
- [ ] Unit test in `server/services/backup.test.js` asserts valid `subdirFilter` generates the exact `--include` and `--exclude` rsync arguments.
- [ ] Unit test in `server/services/backup.test.js` asserts `reloadSettings` is called when `dryRun: false` and skipped when `dryRun: true`.

atomantic:
## Summary

`restoreSnapshot()` in `server/services/backup.js` rsyncs a snapshot back over the user's live `data/` directory — an irreversible, high-impact action. The existing suite covered only `subdirFilter` rejection and which rsync binary gets spawned. This adds tests for the three uncovered contracts named in #3917:

- **`snapshotId` validation** — ids carrying a path separator, whitespace, or shell metacharacters are rejected by the character allow-list before `spawn` is reached. Separately, a bare `..` passes that allow-list (it is made only of allowed characters) and must be caught by the `resolve()`/`relative()` traversal check — that case is asserted on its own so the second guard has independent coverage.
- **`subdirFilter` rsync flags** — a valid filter produces the exact, ordered argument array (`--archive`, `--itemize-changes`, `--itemize-changes`, `--dry-run`, `--include=<dir>/***`, `--include=*/`, `--exclude=*`, src, dest). Asserted as an exact array rather than `arrayContaining` because order is load-bearing: `--exclude=*` must come last or a targeted restore silently widens into a full-tree restore. A companion test asserts no include/exclude flags are emitted when no filter is given.
- **Settings cache re-sync** — `reloadSettings()` is called exactly once after a live restore (`dryRun: false`), and not called for a dry run, for the default (dry-run) invocation, or when rsync exits non-zero on a live restore.

No production code changed. `./settings.js` is partially mocked (only `reloadSettings` overridden) so the test never touches a real `settings.json` or emits socket events.

## Test plan

- `cd server && NODE_ENV=test npx vitest run services/backup.test.js` — 63 passed.
- Mutation-probed each new assertion against a deliberately broken `backup.js` (changes reverted afterward; `git diff` clean):
  - dropping `flags.push('--exclude=*')` → the exact-args test fails.
  - changing `if (!dryRun) await reloadSettings()` to an unconditional call → the dry-run and default-invocation tests fail.
  - deleting the `relative()`/`isAbsolute()` traversal guard → only the bare-`..` bypass-probe test fails.

Closes #3917


atomantic:
Review gate UNSATISFIED: the configured reviewer `agy` (antigravity) is on PATH but its headless run exits 1 with `Individual quota reached. Please upgrade your subscription to increase your limits.` — a dead provider, not a passing review. Not merging on that basis.

Self-review notes on this test-only diff: mutation-probed each new assertion against a deliberately broken `backup.js` (reverted afterward) — dropping `--exclude=*` fails the exact-args test, making `reloadSettings()` unconditional fails the dry-run tests, and deleting the `relative()`/`isAbsolute()` traversal guard fails only the bare-`..` bypass probe. No production code changed.

PR #3973, opened 2026-08-13
fix: gate catalog-bundle pushes on live rows in any block, not just ingredients (#3926)

### LINKED ISSUE #3926: [PeerSync] Missing schema version check gate for catalog ingredient relations and tags when federating catalog bundles
### Problem
In `server/services/sharing/peerSyncReceive.js:172-184`, `applyIncomingPush` checks `catalogBundle` to decide whether the `catalog` schema version gate should be included in `relevantCategories`:

```javascript
if (record.deleted !== true && isPlainObject(catalogBundle) &&
    Array.isArray(catalogBundle.ingredients) &&
    catalogBundle.ingredients.some((i) => i?.deleted !== true)) {
  for (const c of (RECORD_KIND_SCHEMA_CATEGORIES['cat-ingredient'] || ['catalog'])) relevantCategories.add(c);
}
```

If a `catalogBundle` is sent carrying `refs`, `relations`, `tags`, or `media` (introduced in catalog schema versions 4 through 8, see `server/lib/schemaVersions.js:271-317`), but `ingredients` is empty or absent, `relevantCategories` omits `'catalog'`.

Because `'catalog'` is omitted from `relevantCategories`, `scopeVersionDiff` filters out the `catalog` category ahead-version entry. The receiver bypasses the schema-version gate and calls `applyCatalogBundle(catalogBundle)` on an older peer whose database schema cannot handle the newer `relations`, `tags`, or `media` structures.

### Impact
Upgrading one PortOS install to a newer catalog schema and syncing catalog bundles (e.g. relation edges or tag hierarchies) without ingredients to an older peer bypasses version gating. The older peer attempts to parse and insert unknown schema structures, leading to unhandled database errors or silent data loss.

### Fix
Update `server/services/sharing/peerSyncReceive.js` so that `catalogBundle` presence checks for any live entries across `ingredients`, `refs`, `relations`, `tags`, or `media` when deciding to add the `catalog` category to `relevantCategories`.

### Acceptance Criteria
- [ ] `applyIncomingPush` in `server/services/sharing/peerSyncReceive.js` adds `'catalog'` to `relevantCategories` if `catalogBundle` contains non-deleted entries in `ingredients`, `refs`, `relations`, `tags`, or `media`.
- [ ] Pushing a `catalogBundle` containing relations or tags to a peer with an older `catalog` schema version is rejected with HTTP 409 `ERR_SCHEMA_VERSION_AHEAD`.
- [ ] Unit tests cover `catalogBundle` version gating when `ingredients` is empty but `relations`/`tags` are present.

atomantic:
## Summary

`applyIncomingPush` in `server/services/sharing/peerSyncReceive.js` decided whether the `catalog` schema-version gate applied by inspecting `catalogBundle.ingredients` only. Any bundle whose live rows sat in another block therefore left `catalog` out of `relevantCategories`, `scopeVersionDiff` filtered the ahead-entry away, and `applyCatalogBundle` ran on a receiver whose catalog schema can't interpret the payload.

That is reachable today, not just in theory: `buildCatalogBundleForRef` already emits ingredient-free bundles (`refs` only, when a universe's ref rows outlive its ingredient rows), and catalog schema v4–v8 added the `relations`, `tags`, `media` and `catalogTypes` blocks a newer sender can bundle with no ingredients at all.

The gate now counts a live (non-tombstone) row in **any** array-valued block of the bundle rather than an enumerated key list — so a block introduced by a catalog version *newer than this receiver* (which an older receiver cannot name by definition) gates too, which is exactly the payload the gate exists to stop. Such a push is rejected with the existing 409 `PEER_SYNC_SCHEMA_VERSION_AHEAD`.

Tombstone-only bundles stay exempt: those rows are `id`+`deleted`+`deletedAt`+`updatedAt` at every catalog version, and gating them would stall federated catalog deletes the moment one peer upgrades ahead (same reasoning as the existing tombstone-record exemption).

No schema-version bump — this fixes the enforcement of the existing `catalog` gate; the wire shape is unchanged.

## Test plan

- `cd server && NODE_ENV=test npx vitest run services/sharing/peerSync.test.js` — 246 passed.
- New cases in `peerSync.test.js`: a parameterized set covering `relations`-only, `tags`-only, `media`-only, `refs`-only and unknown-future-block bundles from a schema-ahead sender, each asserting the `PEER_SYNC_SCHEMA_VERSION_AHEAD` code, the exact `ahead` entry, and that `applyRemoteChanges` never ran; plus an equal-version relations-only bundle that still applies (the widened gate is not a blanket rejection), and a tombstone-only bundle from a schema-ahead sender that still applies.
- Bypass probe: with the source change reverted and the tests kept, exactly the five new schema-ahead cases fail — confirming they exercise the fix rather than re-implementing it.
- No DB-backed suites touched.

Closes #3926


atomantic:
Review gate UNSATISFIED: the configured reviewer `antigravity` (`agy`) is installed but its headless run exited 1 with `Individual quota reached. Please upgrade your subscription to increase your limits. Resets in 42m` — no review was produced. Re-run `agy --dangerously-skip-permissions --model "Gemini 3.5 Flash (High)" --print-timeout 30m -p <review prompt>` against this diff once the quota resets before merging.

PR #3977, opened 2026-08-13
Give a failed Quota Burn initial load a named cause and a Retry button

### LINKED ISSUE #3930: [QuotaBurn] Initial load failure silently trapped on un-retryable error screen
### Problem
In `client/src/pages/QuotaBurn.jsx:78` and `client/src/pages/QuotaBurn.jsx:198`, `load()` issues `api.getQuotaBurn(refresh, { silent: true }).catch(() => null)`. When the initial request fails (such as a backend startup delay, network disconnect, or 500 error), `.catch(() => null)` swallows the error and `data` evaluates to `null`. `load()` returns without populating `config`, and `.finally(() => setLoading(false))` sets `loading` to `false`. On line 198, `if (!config)` triggers an early return of `<div className="p-6 text-gray-400">Quota burn is unavailable — the server did not return a plan.</div>`.

### Impact
Because line 198 returns before rendering the page header, the header containing the "Refresh quota" button and page title is unrendered. The user is left trapped on a static error screen with no feedback on why the fetch failed and no retry mechanism short of manually reloading the entire browser tab.

### Fix
In `client/src/pages/QuotaBurn.jsx`, track initial fetch failures in a local `error` state. When `!config` and `error` is truthy, render an explicit error banner containing the error details and a "Retry" button that invokes `load()`.

### Acceptance criteria
- [ ] Failed initial `getQuotaBurn` request sets an explicit `error` state in `QuotaBurn.jsx`.
- [ ] Unavailable state renders an error banner and a working "Retry" button.
- [ ] Clicking "Retry" re-triggers `load()` and clears the error state on successful response.

atomantic:
## Summary

A failed first read on the Quota Burn page left the user on a dead end. `load()` did `.catch(() => null)`, so a backend startup delay, a dropped connection, or a 500 all landed as "no data": `config` stayed null, `loading` flipped false, and the `if (!config)` branch returned a single grey line *before* the page header — so "Refresh quota", the one control that could re-read, was never rendered. Nothing named the cause and nothing short of reloading the browser tab recovered.

- The read's failure is now kept in its own `loadError` state: `null` means the read has never failed, a string is the message from the read that did. That keeps "the request failed" distinct from "the server answered and has no plan" — collapsing them into `!config` is what made the state un-diagnosable.
- Any successful read clears it, including the pending-quota polls, so a stale message can't outlive the condition.
- The unavailable branch now renders the shared `Banner` (tone `error`) carrying that message plus a working Retry that re-invokes `load()`. When there is no error to name it still offers Retry with "The server did not return a plan." — that state is equally stuck.
- The header's "Refresh quota" and the new Retry both go through one small `reload(refresh)` helper rather than two copies of the set-loading/load/finally dance.

The GET keeps `{ silent: true }`: the page owns this error's UI, so per `client/src/CLAUDE.md` the toast must not also fire. No `alert`/`confirm`.

## Test plan

`cd client && npm test -- src/pages/QuotaBurn.test.jsx` — 29 passed, including two new cases:

- a rejected initial `getQuotaBurn` renders the banner with the failure message, and clicking Retry re-reads, paints the plan, and removes both the banner and the message;
- a 200 that carries no config still renders Retry alongside the no-plan text.

Closes #3930


atomantic:
Review round 1 (antigravity, headless) raised two findings.

1. **Stale error when a read resolves with nothing** — applied in 356863530. `if (!data) return` sat before `setLoadError(null)`, so a read that *succeeded* with a falsy body kept the previous attempt's network error on the banner. The catch now returns an explicit `READ_FAILED` symbol and the early return keys on that sentinel rather than falsiness, so "the read threw" and "the read answered with nothing" clear the error state differently. Covered by a new test.

2. **A failed refresh/poll on an already-loaded page is silent** — not applied here. It is real, but pre-existing (the old code was `.catch(() => null)` with `silent: true`, equally silent) and outside this PR's scope, which is the initial-load trap #3930 describes. Filed as #3979.

30 client tests pass.

PR #3986, opened 2026-08-13
fix: surface job skill save and preview failures in the Prompt Manager (#3937)

### LINKED ISSUE #3937: [UX Audit] [PromptManager] Job skill save failures are swallowed silently with no user error toast
## Problem
In [`client/src/pages/PromptManager.jsx:292-298`](file:///Users/adameivy/github.com/atomantic/PortOS/client/src/pages/PromptManager.jsx#L292-L298), `saveJobSkill` executes `apiSaveJobSkill` with `{ silent: true }` and an empty `.catch(() => {})`:

```javascript
const saveJobSkill = async () => {
  setSaving(true);
  await apiSaveJobSkill(selectedJobSkill, jobSkillContent, { silent: true }).catch(() => {});
  setSaving(false);
};
```

Passing `{ silent: true }` instructs `request()` to omit default error toast notifications, while `.catch(() => {})` swallows the rejected promise without calling `toast.error()`.

## Impact
If `apiSaveJobSkill` fails (due to a network timeout, authorization failure, or backend error), no toast or error banner is presented to the user. The "Save" button briefly shows a disabled saving state and then re-enables normally, misleading the user into believing their changes were persisted when they were actually lost.

## Fix
In `client/src/pages/PromptManager.jsx`:
- Catch errors from `apiSaveJobSkill` and display a user-visible toast error notification:
  ```javascript
  const ok = await apiSaveJobSkill(selectedJobSkill, jobSkillContent, { silent: true })
    .then(() => true)
    .catch((err) => { toast.error('Failed to save job skill: ' + err.message); return false; });
  if (ok) { toast.success('Job skill saved'); }
  ```

## Acceptance criteria
- [ ] If `apiSaveJobSkill` fails on save, an error toast displaying the failure message appears.
- [ ] Successful saves display a success confirmation toast or feedback.
- [ ] The `saving` flag is reset in all execution paths.

atomantic:
## Summary

`saveJobSkill` in `client/src/pages/PromptManager.jsx` fired the PUT with `{ silent: true }` and an empty `.catch(() => {})`. A network blip, an auth failure, or a 500 discarded the edit while the Save button just re-enabled — indistinguishable from a save that worked.

- A failed save now toasts `Failed to save job skill: <message>`; a successful one confirms with `Job skill saved`. The `saving` flag clears on both paths, so Save is always retryable.
- `previewJobSkill` on the same panel had the same hole (`.catch(() => ({}))` left the preview blank with no cause); it now toasts the failure and leaves the existing preview alone.
- Both keep `{ silent: true }` so the toast fires from one layer only, matching the silent-vs-toasting convention in `client/src/CLAUDE.md` and the `saveStage` / `saveVariable` handlers directly above.

## Test plan

- `cd client && npm test -- src/pages/PromptManager.test.jsx` — 33 passed.
- New `PromptManager job skill save feedback` suite covers: rejected save toasts the message and does not toast success; Save re-enables after a failure; a successful save toasts and sends `(name, content, { silent: true })`; a rejected preview toasts instead of blanking the panel.

Closes #3937


atomantic:
Review pass (antigravity/`agy`, diff-only): one finding, dismissed as a false positive. It claimed `../components/ui/Toast` is never mocked in `PromptManager.test.jsx`; the `vi.mock('../components/ui/Toast', ...)` block is pre-existing at line 43 (outside the diff hunk the reviewer saw), and the suite passes 33/33.

PR #3992, opened 2026-08-13
perf: batch normalizeTags into two queries instead of one per tag (#3943)

### LINKED ISSUE #3943: [Catalog] N+1 per-label database reads during normalizeTags taxonomy mapping
### Problem

`normalizeTags` in `server/services/catalogDB/tags.js:27-50` maps raw tag inputs to canonical tags by iterating through raw labels in a `for` loop. For each label, it executes `INSERT INTO catalog_tags (id, label) VALUES ($1, $2) ON CONFLICT DO NOTHING RETURNING label`. When a tag already exists (the common case), `ins.rows[0]` is empty, forcing a second `SELECT label FROM catalog_tags WHERE id = $1` query per tag label.

### Impact

Creating or updating ingredients or committing extracted scraps with M tags issues between M and 2M individual database queries per ingredient. When committing a batch of N ingredients with M tags each, this results in up to 2*N*M SQL round-trips.

### Fix

In `server/services/catalogDB/tags.js`, batch tag normalization by querying all requested tag IDs in a single `SELECT id, label FROM catalog_tags WHERE id = ANY($1)` query. Bulk insert any missing tags in a single statement and return the canonical labels list.

### Acceptance criteria

- [ ] `normalizeTags` resolves any list of tag labels using at most 2 batched SQL queries regardless of tag count.
- [ ] Canonical tag creation and casing preservation rules remain unchanged.
- [ ] `catalogDB.tags.test.js` passes cleanly.

atomantic:
## Summary

`normalizeTags` (`server/services/catalogDB/tags.js`) mapped raw tag labels to canonical `catalog_tags` rows one label at a time: an `INSERT … ON CONFLICT (id) DO NOTHING RETURNING label` per label, plus a follow-up `SELECT label … WHERE id = $1` whenever the row already existed — the common case. An ingredient carrying M tags therefore cost between M and 2M round-trips on every create, every tag-touching update, and every committed scrap.

The list is now resolved in at most two statements regardless of tag count:

1. `SELECT id, label FROM catalog_tags WHERE id = ANY($1)` reads every pre-existing row in one hop.
2. A single multi-row `INSERT … VALUES (…),(…) ON CONFLICT (id) DO NOTHING RETURNING id, label` creates whatever is missing — skipped entirely when nothing is, so the all-existing path costs one query.

A row that conflicts on step 2 (created between the two statements) falls back to the incoming label rather than spending a third round-trip; under the single-user trust model that window is not reachable in practice.

Canonical semantics are unchanged: blank tags dropped, casing/whitespace variants collapsed to one row via `canonicalTagKey`, order of first appearance preserved, first-seen casing wins within a call, and a stored label still beats the incoming casing across calls (`Noir` over a later `NOIR`). Passing `{ client }` still runs both statements on the caller's transaction.

## Test plan

- `server/services/catalogDB.tags.test.js` — the in-memory `catalog_tags` mock now interprets the batched `ANY(...)` select and the multi-row insert; all existing dedup / casing / ordering assertions pass unchanged, plus three new ones: 40 tags issue exactly 2 queries, an all-existing list issues exactly 1, and `{ client }` routes both statements to the transaction client instead of the pool.
- `cd server && NODE_ENV=test npx vitest run $(ls services/catalog*.test.js routes/catalog*.test.js)` — 16 files passed, 199 tests passed, 3 DB-backed files skipped (no test DB in this worktree; not run against the real `portos` database).

Closes #3943


atomantic:
Reviewer `antigravity` (`agy`) run against the diff: no correctness, injection, or dead-code findings. One Low edge-case note — a multi-row insert of more than 32,767 *missing* tags in a single call would exceed Postgres's 65,535-parameter limit (2 params per tag). Won't-fix: `normalizeTags` is called with one ingredient's tag list, which is bounded by the form and by bulk-import row shape, and chunking would break the issue's acceptance criterion of at most 2 batched queries. The batched `SELECT` binds a single array param and is unaffected.

PR #3997, opened 2026-08-13
show the arc verifier the world category canon it judges groundedness against

atomantic:
## Summary

The series arc verifier judged groundedness against `worldCanonText` only — the named-canon trunks, which stay empty until prose mints entities into them. A Universe Builder world actually defines its factions, locations, vehicles, artifacts and interfaces as category variations, so the verifier was handed a block labelled "world canon" that omitted the entire built world and reported locked faction cards as invented / "absent from the World Canon". No arc edit can close a finding like that.

That stalls the autopilot rather than merely annoying it: the foundation gate's structure repair reverts its whole rewrite whenever verification leaves any blocker, so one permanent false positive discards every real fix beside it. The sibling arc-resolve and volume-verify prompts already rendered this block; only the arc verifier was blind, which is why that verify/resolve pair could never converge.

This change:

- Renders the category and composite canon in `pipeline-arc-verify.md` and states that a listed entity is grounded even without a named-canon record.
- Adds migration 263 to upgrade installs still on the prior shipped body, and resyncs the earlier migrations that track this file plus the drift baseline.
- Exports `renderVerifyIssueLeaf` / `renderVerifySeasonFields` and adds `verifyPromptContract.test.js` so the checklist and the rendered context stay in agreement — every backticked record field the prompt cites must be renderable, and the verify prompt must render every world block its paired resolve prompt does.

## Test plan

- [x] `cd server && npm test -- services/pipeline/arcPlanner/verifyPromptContract.test.js`
- [x] `cd server && npm test -- ../scripts/migrations/263-arc-verify-world-category-canon.test.js ../scripts/setup-data-drift.test.js`
- [x] `cd server && npm test -- services/pipeline/arcPlanner/`
- [x] Shipped `pipeline-arc-verify.md` MD5 matches `NEW_SHIPPED_MD5` (`9f32e91bd33b97d30e1cbb2e697f4fc3`)
- [ ] CI green on this PR


atomantic (commented):
Requesting Copilot review

atomantic:
@copilot please review this pull request.

PR #4086, opened 2026-08-14
add a swappable prompt conditioner to MiniMax H3 video renders

atomantic:
## Summary

MiniMax H3 does not use its Qwen3-VL-32B conditioner as a language model: the DiT reads the **unnormalized hidden state after language layer 49** and feeds it straight into `condition_proj`. Layers 50–63, the final norm and `lm_head` are never evaluated — the MLX port doesn't even load them.

That makes the conditioner substitutable. Any checkpoint carrying the same embedding + language layers 0–49 + vision tower produces a conditioning signal of the same shape, so swapping it changes **how the model reads a prompt** without touching the diffusion weights, the VAEs or the sampler.

Video Gen now offers that choice under the Model field, defaulting to the stock conditioner so nothing changes unless you pick something else:

| Option | What it is | Extra download |
|---|---|---|
| **Stock** | the conditioner inside `MiniMaxAI/MiniMax-H3`'s `FL2VA/text_encoder/` | none |
| **Ultra-Heretic uncensored** | `ethanfel/Qwen3-VL-32B-Ultra-Heretic-H3-ComfyUI-INT8-ConvRot`, bf16 — Qwen3-VL-32B-Instruct abliterated with Heretic v1.2.0, repackaged for H3 | ~48 GB, one pinned file |

### How the swap works

The pinned MLX runtime is verified clean before every render, so the swap lives entirely outside it:

1. **A composed checkpoint root** — a tree of symlinks into the upstream `FL2VA/` snapshot with only `text_encoder/` replaced, so the runtime's own `from_pretrained()` loads it unmodified. Built under `~/.portos/minimax-h3-encoder-shims/`, deliberately outside the checkout (anything inside would read as untracked in the pin verification).
2. **A key-prefix rewrite** — the ComfyUI packaging flattens the namespace (`model.layers.N.…` / `visual.…`) while the port's loader matches the HF one. `install_key_prefix_map()` wraps `_wanted` so keys are translated before it sees them, then delegates every real decision back to the pinned implementation. It raises loudly if a future pin drops that hook rather than mis-loading.
3. **A synthesized final norm** — the substitute omits `norm.weight` (correct: H3 reads the state *before* it) but the port builds the whole module tree and refuses to load with any parameter missing. The runner writes a ones-filled companion shard the loader's glob picks up, keyed in the substitute's own namespace so the prefix map rewrites it like any other key.

Only the bf16 variant is usable — the repo's INT8 ConvRot and NVFP4/AWQ variants use ComfyUI's learned-rotation quantization the MLX loader can't dequantize, and the `50_63` tails belong to a ComfyUI node PortOS doesn't implement. The download lane is therefore scoped to that one pinned file rather than a repo snapshot (~130 GB for ~48 GB of usable weights).

Options live in code (`server/lib/videoTextEncoders.js`) rather than the media-models registry, mirroring `icLoraWeights.js`: a stale `data/media-models.json` must never be able to name a checkpoint this build's runner has no key map for. No seed file or migration needed.

## Test plan

- **Verified the drop-in against the installed pin**, not just in theory: read the substitute's safetensors header via an HTTP range request, applied the prefix map, and diffed the result against the parameter set the real loader builds. The file's 902 tensors plus the synthesized norm map onto exactly the 903 parameters expected — 552 language, 351 vision, **zero missing, zero extra, zero skipped**. Re-ran after the review pass; still exact.
- `server`: 1361 files / 28,274 tests pass. `client`: 616 files / 7,386 tests pass. `biome lint` clean.
- New coverage: registry semantics + the client projection (`videoTextEncoders.test.js`), a client/server parity guard on the `stock` sentinel (`videoTextEncoders.parity.test.js`), the runner's arg validation / key remap / shim composition / norm synthesis (`generate_minimax_h3.test.js`), the download+repair+status lane (`videoGen.integrity.test.js`), argv threading and the not-cached 400 (`local.test.js`), pre-staging rejection (`prepareParams.test.js`), form state and remix/resume round-trip (`useVideoGenForm.test.jsx`), and the picker (`TextEncoderPicker.test.jsx`).
- An unswapped render posts the same body, builds the same argv, and persists the same job params as before this knob existed — asserted rather than assumed.
- Docs: `docs/features/video-text-encoders.md`, indexed from `docs/README.md`, including how to validate a candidate conditioner before downloading tens of GB.

## CI fixes rolled in

CI was red on `main` itself before this branch, on jobs unrelated to the feature. Three fixes, each verified:

- **`client/src/pages/VideoGen.terms.test.jsx`** — the prompt-enhancement work mounted `useProviderModels` under `VideoGen`, but this file mocks `services/api` wholesale and had no `getProviders`, so the hook threw out of a passive effect. The assertions still passed; the three unhandled rejections failed the job. **The client job is green again.**
- **`server/services/settings.test.js`** — forced `access()` to throw `ENOTDIR` via a path whose parent is a file, but Windows reports that path as `ENOENT`, flipping the assertion. Now drives the non-ENOENT branch through a stubbed `access()`, so it checks the branch it names on every platform. `settings.js` is unchanged; verified the test still fails when the `ENOENT` guard is removed.
- **`server/services/updateChecker.test.js`** — asserted a `'/mock/package.json'` literal against a path the source builds with `join()`. Previously masked by the settings failure ahead of it.

The Windows job is still red on a **pre-existing** chain — it bails at the first failing file, so each fix unmasks the next. The next one is `lib/icloudFile.test.js` (a macOS ubiquity-container symlink test), untouched by this branch. Tracked in #4088 rather than chased here.


atomantic:
### Note: this branch also fixes an unrelated Windows CI break

`main` was already red on the **Windows server unit tests** job (5+ consecutive runs), so this PR could not go green on its own merit without addressing it.

The cause is in `server/services/pipeline/audioMux.test.js`: it splices an absolute path into a `new RegExp(...)`, and its escaping character class `[.+^${}()|[\\]\\\\]` **closed early on its own `\\]`** — so it matched "a metacharacter followed by two literal backslashes" and escaped nothing at all.

It went unnoticed because a POSIX temp path has no backslashes and its unescaped `.` still matches itself. On Windows the path is `C:\Users\RUNNER~1\AppData\Local\Temp\…`, which the unescaped splice turns into the escape sequences `\U`, `\R`, `\A`, `\T` — so the assertion could never match.

Fixed with the standard `[.*+?^${}()|[\]\\]` escape, extracted as a named helper. Verified against a simulated Windows path that the pattern now matches **and** still rejects a wrong suffix or a different directory — a fix here must not make the assertion vacuously true.

Also in the same commit: `build_encoder_shim` now passes `target_is_directory` when symlinking. No-op on POSIX, but load-bearing on Windows, where a directory linked as a file symlink can't be traversed — the VAEs, tokenizer and processor are all directories.

Both are unrelated to the feature, but the repo's review policy is to fix a real issue regardless of whether the branch introduced it.

atomantic:
### CI triage (first run after the audioMux fix)

The audioMux fix worked — Windows got past it. What remains is unrelated to this branch:

| Job | Result | Cause |
|---|---|---|
| `lint` | ✅ | |
| `test (24.x)` (Linux server) | ✅ | |
| `DB tests and server smoke` | ✅ | |
| `Plan test impact` | ✅ | |
| `Client tests and build` | ❌ | Reports **`Test Files 617 passed (617)`** — every file passed. The job failed on an unhandled rejection logged *after* the suite: `No "getProviders" export is defined on the "../services/api" mock`, from `src/hooks/useProviderModels.js:67`. Untouched by this branch. |
| `Windows server unit tests` | ❌ | `ENOSPC: no space left on device` on the runner, plus an unrelated `lib/tuiPromptRunner.test.js` assertion. Neither file is in this diff. |

`main` was already failing CI before this branch (5+ consecutive runs), so these are not regressions introduced here. Locally on this exact commit: server **1361 files / 28,276 tests pass**, client **617 files / 7,388 tests pass**, `biome lint` clean.

Applying the standard single conservative re-run of the failed jobs on the **same SHA** (`b73b9eb`) to separate flake from real — `ENOSPC` is definitionally infrastructure.

PR #4189, opened 2026-08-14
fix: re-anchor agy's --print <prompt> at the end of the argv so trailing extraArgs can't strand it

### LINKED ISSUE #4110: Antigravity CLI: trailing extraArgs land after spliced prompt, not before --print
## Problem

`agy --print` takes the prompt as `--print`'s VALUE, so `ensureAntigravityPrintArgs` (`server/lib/antigravity.js:160-177`) puts `--print` last in the built argv, and `prepareAntigravityPrompt` (`server/lib/antigravity.js:191-205`) splices the prompt in as the value immediately after the trailing `--print` marker.

But `executeCliRun` in `server/lib/cliProviderRun.js:93` builds:

```js
const builtArgs = [...buildCliArgs(effectiveProvider), ...(Array.isArray(extraArgs) ? extraArgs : [])];
```

`buildCliArgs` → `buildVendorCliArgs` → `antigravityCliArgs` (`server/lib/providerVendors.js:140-142`) calls `ensureAntigravityPrintArgs`, whose output already ends in `--print`. Any non-empty `extraArgs` therefore get appended *after* that trailing `--print`.

`prepareCliPrompt` (`server/lib/providerVendors.js:357-360`) then dispatches to `prepareAntigravityPrompt`, which finds the LAST print flag in the argv and splices the prompt in right after it (`server/lib/antigravity.js:196-203`, `idx+1`). Since `--print` is not the last token once `extraArgs` is non-empty, the prompt lands *before* the extraArgs, not at the end.

Final argv shape for a non-empty `extraArgs` case:

```
… --dangerously-skip-permissions --print <PROMPT> <extraArg1> <extraArg2> …
```

The prompt is still correctly delivered as `--print`'s value, but the extraArgs following it become stray positional arguments that `agy` may reject or misinterpret.

## Current risk

Low today: no caller passes `extraArgs` for an antigravity provider in the shipped paths (Creative Director plan via runner/direct, Run Prompt, ask, vision have all been validated end-to-end with empty `extraArgs`). This is a latent footgun for the next caller that adds `extraArgs` for an antigravity-routed run.

## Fix shape (pick one)

1. Strip/relocate the trailing `--print` marker in `executeCliRun` (or in `ensureAntigravityPrintArgs`'s consumer) so any `extraArgs` are inserted *before* it, keeping `--print` — and therefore the prompt — as the true last tokens.
2. Have `prepareAntigravityPrompt` move the `--print <prompt>` pair to the very end of the argv after splicing, rather than just inserting after the flag's current position.

## Affected code

- `server/lib/antigravity.js` — `ensureAntigravityPrintArgs`, `prepareAntigravityPrompt`
- `server/lib/cliProviderArgs.js` — `prepareCliPrompt` re-export
- `server/lib/providerVendors.js` — `antigravityCliArgs`, vendor `preparePrompt` dispatch
- `server/lib/cliProviderRun.js` — `executeCliRun` (`builtArgs` concatenation, line ~93)

atomantic:
## Summary

`ensureAntigravityPrintArgs` deliberately leaves a bare `--print` marker as the last token of an agy argv, because `agy --print` takes the prompt as the flag's VALUE (it never reads stdin). But `runCliProviderPrompt` (`server/lib/cliProviderRun.js`) builds `[...buildCliArgs(provider), ...extraArgs]` — so with a non-empty `extraArgs` the marker is no longer last, and `prepareAntigravityPrompt`'s in-place splice produced:

```
… --dangerously-skip-permissions --print <PROMPT> <extraArg1> <extraArg2> …
```

The prompt still reached agy as `--print`'s value, but every extraArg became a stray positional agy may reject or misinterpret.

This takes the issue's **fix shape 2**: `prepareAntigravityPrompt` now lifts the print flag out of its current position and re-appends `<flag> <prompt>` as the final two tokens, so the pair is last no matter what got concatenated on. That is more local than rewriting the concatenation site, and it holds for any future caller that appends to a built agy argv — not just the one that does today.

Details:

- The flag's original **spelling** is preserved when relocating (`-p` / `--prompt` / `--print`), so a user-baked short form still reaches agy as written.
- The **empty-`extraArgs`** shape is byte-identical to before — the existing spawn paths (Creative Director, Run Prompt, ask, vision, `services/imageGen/agy.js`) are unaffected.
- `server/lib/aiToolkit/internal/antigravity.js` needs no mirror change: it has no prompt preparer, only the arg builders.
- Comments at both ends now name the real concatenation site. The issue attributed it to `executeCliRun`; that function (`server/services/runner.js`) takes no `extraArgs` — `runCliProviderPrompt` is the one that concatenates.

## Test plan

- `cd server && NODE_ENV=test npx vitest run ../server/lib/antigravity.test.js ../server/lib/cliProviderArgs.test.js ../server/lib/cliProviderRun.test.js` — 88 passed.
- New regression coverage:
  - `server/lib/antigravity.test.js` — with non-empty `extraArgs` appended past the marker, the argv ends with `--print <PROMPT>`, every extraArg precedes it, and exactly one print flag survives the move; the empty-`extraArgs` shape is unchanged; the flag spelling is preserved.
  - `server/lib/cliProviderArgs.test.js` — the same assertion one level up, through the real `prepareCliPrompt` vendor dispatch, over an argv built exactly the way `runCliProviderPrompt` builds it (`[...buildCliArgs(provider), ...extraArgs]`), plus the unchanged no-extraArgs shape.
- Bypass probe: reverting `prepareAntigravityPrompt` to the old in-place splice fails all 3 new assertions, confirming they are not vacuous.
- Wider suites unaffected: `server/lib` (310 files / 7705 tests) green; `server/services/imageGen`, `agentCliSpawning`, `slashdoInvocation`, `tuiHandshake`, `runner`, `askService`, `codeReview`, `cos-runner` all green.

Closes #4110

### DIFF
--- a/.changelog/next/fixed-issue-4110.md
+++ b/.changelog/next/fixed-issue-4110.md
@@ -1,0 +1,1 @@
+- Antigravity CLI runs with extra arguments no longer strand them after the prompt — `--print <prompt>` is re-anchored as the final argv pair at spawn time
--- a/server/lib/antigravity.js
+++ b/server/lib/antigravity.js
@@ -201,23 +201,25 @@ const NOOP_CLEANUP = () => {};
  * no-op for the already-trailing case.
  *
  * The flag's original SPELLING is preserved (`-p` / `--prompt` / `--print`) so a
- * user-baked short form still reaches agy as they wrote it.
+ * user-baked short form still reaches agy as they wrote it — the LAST spelling
+ * wins, matching which flag agy itself would have honored.
+ *
+ * EVERY print flag is lifted, not just the last one. `ensureAntigravityPrintArgs`
+ * strips baked print flags from the provider's saved argv, but the `extraArgs`
+ * concatenated on afterwards are never stripped — so leaving an earlier `-p`
+ * behind would have agy consume the token following IT as the prompt.
  *
  * @param {string[]} args - argv as built by ensureAntigravityPrintArgs
  * @param {string} prompt - the full prompt text
  * @returns {{ args: string[], useStdin: false, cleanup: () => void }}
  */
 export function prepareAntigravityPrompt(args = [], prompt = '') {
-  const out = [...args];
-  // Find the LAST print flag so the prompt lands as its value even if the argv
-  // carries stray tokens (it shouldn't, post-ensureAntigravityPrintArgs).
-  let idx = -1;
-  for (let i = out.length - 1; i >= 0; i--) {
-    if (ANTIGRAVITY_PRINT_FLAGS.includes(out[i])) { idx = i; break; }
+  let flag = '--print';
+  const out = [];
+  for (const arg of args) {
+    if (ANTIGRAVITY_PRINT_FLAGS.includes(arg)) flag = arg;
+    else out.push(arg);
   }
-  // No print flag at all → add one. Otherwise lift the existing marker out of
-  // its current position so it can be re-appended as the final pair.
-  const flag = idx === -1 ? '--print' : out.splice(idx, 1)[0];
   out.push(flag, prompt);
   return { args: out, useStdin: false, cleanup: NOOP_CLEANUP };
 }
--- a/server/lib/antigravity.js
+++ b/server/lib/antigravity.js
@@ -78,8 +78,9 @@ export function parseAntigravityModelList(stdout) {
 
 // agy print flags. Unlike the old Gemini CLI (which read the prompt from stdin),
 // `agy --print`/`-p`/`--prompt` takes the prompt as the flag's VALUE and does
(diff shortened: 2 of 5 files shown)


atomantic:
Review gate: `agy` headless, 2 rounds. Round 1 flagged that only the LAST print flag was lifted — extraArgs are concatenated on after `ensureAntigravityPrintArgs` has stripped the provider argv, so a second `-p`/`--prompt` can still arrive and would be left bare for agy to read the following token as the prompt. Fixed in d0b7ba2 with a single filtering pass (also simpler than the index splice it replaced) plus a regression test. Round 2: VERDICT: CLEAN.

PR #4204, opened 2026-08-14
refactor: extract the duplicated rigged-GLB fit-to-height math into fitModelToHeight (#4119)

### LINKED ISSUE #4119: Extract shared rigged-GLB avatar boilerplate (3 consumers)
## Problem

Three components each independently reimplement the same rigged-GLB avatar boilerplate:

- `client/src/components/cos/MuseCoSAvatar.jsx`
- `client/src/components/cos/MiniCharacterCoSAvatar.jsx`
- `client/src/components/city/PlayerAvatar.jsx`

Specifically, each duplicates:

1. **Load + clone.** `useGLTF(url)` + `useMemo(() => SkeletonUtils.clone(gltf.scene), [gltf.scene])`, with the identical copy-pasted "rebind SkinnedMeshes" rationale comment.
2. **Fit-to-height.** `new THREE.Box3().setFromObject(scene)` → `getSize`/`getCenter` → `scale = TARGET_HEIGHT / Math.max(size.y, 1e-3)`. Muse centers vertically; Mini and Player rest feet on the ground — the fit logic has already drifted between consumers.
3. **Crossfade.** A `fadeTo(clipName, ...)` callback that crossfades between animation actions (e.g. `MuseCoSAvatar.jsx` line 84, `PlayerAvatar.jsx` line 109).

## Fix shape

- `useRiggedGLTF(url)` hook in `client/src/hooks` — load + clone + `withInPlaceClips` + `useAnimations`, returning `{ scene, actions, names, mixer }`.
- `fitModelToHeight(scene, { targetHeight, feetOnGround })` pure helper in `client/src/utils` — parameterize the vertical-center-vs-feet-on-ground divergence explicitly instead of leaving it implicit per consumer.
- A low-level `crossfade(actions, activeRef, name, { timeScale, fade })` primitive is a weaker candidate for extraction — Muse's version carries montage `once`/`reps` branching that Mini/Player don't need — so only extract it if it stays simple after the other two are pulled out.

## Why deferred

Deferred from the CyberCity player-swap PR: touches two CoS avatar components well outside that diff's scope, and each new public module needs its barrel entry + README row + index-test coverage per the repo's Module Organization convention (`server/lib/README.md` pattern, mirrored for `client/src/hooks`/`client/src/utils`).

atomantic:
## Summary

Three rigged-GLB avatars — `MuseCoSAvatar`, `MiniCharacterCoSAvatar`, and the CyberCity `PlayerAvatar` — each hand-rolled the same fit-to-height math: measure a `Box3`, divide `targetHeight` by `size.y`, then re-anchor the scene. This extracts it into `fitModelToHeight` in `client/src/utils/modelFit.js` and points all three at it.

The three copies had already drifted, so the helper parameterizes the divergence instead of unifying it silently. **Framing is unchanged for every consumer:**

| Consumer | Call | Anchoring |
|---|---|---|
| Cyber Muse | `{ targetHeight: 1.9, yOffset: 0.05 }` | bounding-box center, nudged up — a portrait frame with no ground contact |
| Mini-character | `{ targetHeight: 2.6, feetOnGround: true, yOffset: GROUND_Y }` | feet on the shared `-1.4` ground plane, under the shadow disc |
| CyberCity player | `{ targetHeight: 1.7, feetOnGround: true }` | feet at `y = 0`, the rig owns world position |

Two edges the copies handled inconsistently are now handled once:

- **Idempotency.** Only `PlayerAvatar` reset the scene transform before measuring. Without it, StrictMode's second mount effect measures an already-fitted model, computes `scale ≈ 1`, and blows it back up to source size — a latent bug in the two CoS avatars. The helper always resets, which is a no-op for the first (and only real) fit, since `GLTFLoader`'s scene root is a plain identity-transform `Group`.
- **Degenerate inputs.** A geometry-less object measures as an *empty* `Box3` whose `min` is `+Infinity`; fitting it would fling the object to `-Infinity`. It is now left at identity. A missing or non-numeric `targetHeight` likewise leaves the object untouched and logs, rather than dividing into `NaN` and writing that through `scale`/`position`. The existing `Math.max(size.y, 1e-3)` guard against a zero-height model is kept verbatim and covered.

### Scope

Part 1 of the issue's fix shape — `useRiggedGLTF(url)` — has already landed as `client/src/hooks/useClonedGltf.jsx` (`useGLTF` + `SkeletonUtils.clone` + optional animation transform + `useAnimations`, plus the `GltfPrimitive` that stops R3F disposing shared drei-cache resources). All three components already use it, so renaming it would be churn; no change here.

Part 3 — the low-level `crossfade` primitive — is **deliberately skipped**, as the issue anticipated. With the fit extracted, the three `fadeTo` implementations differ in kind rather than in configuration: Muse branches over `once` / `{ reps: N }` / infinite loop modes to drive its montage, the player short-circuits on "already the active clip" because it is called every frame from `useFrame`, and the mini-character is an effect body with a cleanup `fadeOut` and no active-action ref at all. A primitive spanning all three would take more parameters than the ~10 lines it replaced.

Also drops the now-unused `three` import from `MiniCharacterCoSAvatar`, and registers the new module in the `client/src/utils` barrel + README per the Module Organization convention.

## Test plan

- New `client/src/utils/modelFit.test.js` (15 cases) asserts observable outcomes — where the model's *world* bounding box lands after the fit — rather than restating the formula: rendered height matches `targetHeight`, aspect ratio is preserved, the default anchoring centers the box on the origin, `feetOnGround` puts `min.y` on the ground while still recentering x/z, `yOffset` moves whichever anchor is active, and both degenerate cases stay finite / untouched.
- The idempotency case is checked at call counts 2, 3, and 4. This matters: the un-reset bug *oscillates* with period 2, so a check made only after an odd number of calls passes vacuously — the first draft of this test did exactly that.
- Each assertion was verified to bite by temporarily reverting the corresponding line in the helper (drop the reset → 2 failures; drop the `Math.max` guard → 1; drop the empty-box guard → 1; collapse `feetOnGround` to the centered branch → 2).
- `cd client && npm test` — 630 files / 7590 tests passing, including the `client/src/utils/index.test.js` barrel + README drift check.

One caveat is documented rather than fixed: `Box3.setFromObject` measures in world space while `scale`/`position` are written in local space, so a non-identity parent transform is folded into the fit. Every consumer already behaved that way before the extraction, so changing it would break this PR’s own behavior-preservation contract.

Closes #4119


atomantic:
Review gate: **antigravity (`agy`), headless, 3 rounds — clean.**

Round 1 returned no findings. Round 2 used a sharper prompt that forced it to restate each consumer's *old* scale/position expressions and compare them against what the new `fitModelToHeight` call computes; it confirmed PASS on all three call sites and raised four items. Round 3, on the updated diff, returned no findings.

Addressed in `9e728c5`:

- **Non-finite `targetHeight` wrote `NaN` through the transform.** The options object defaulted to `{}`, so `fitModelToHeight(scene)` divided into `NaN` and silently dropped the model out of the scene. The options object is now required and a non-finite height leaves the object untouched and logs. Covered by four new cases.

Declined, with reasons:

- **World-space measurement vs local-space write.** Real, and now documented in the helper. Not changed: all three consumers parent their scene under a group they drive per frame and already folded that transform in, so "fixing" it would be exactly the silent behavior change this PR is meant to avoid.
- **A zero-height model scales x/z by 2600×.** True, but `Math.max(size.y, 1e-3)` is verbatim what all three consumers already had. Both outcomes are garbage for a degenerate input no real model hits, so the existing guard stays rather than introducing a fourth behavior.
- **Module-scoped scratch `Vector3`s to avoid two allocations.** This runs once per loaded scene inside an effect, not per frame. Two short-lived vectors are not worth module-level mutable state in a helper called a handful of times per session.

PR #4208, opened 2026-08-14
fix: re-derive longevity time horizons on read so goal urgency stops drifting (#4122)

### LINKED ISSUE #4122: Re-derive longevity.timeHorizons on read instead of trusting stale snapshot
## Problem

`computeGoalUrgency` (`server/services/identity/goals.js:65-81`) ranks goal urgency off `timeHorizons.yearsRemaining`, but `getGoalsTree()` (`server/services/identity/goals.js:355`) reads the *persisted* horizons via `loadJSON(LONGEVITY_FILE, …)` (also used at lines 233-240, 284-285, 324-326, 363) — and only `deriveLongevity()` (`server/services/identity/longevity.js:22`) ever consults the current date.

So `yearsRemaining` is frozen as of the last derive (triggered by `setBirthDate` or an explicit `/longevity/derive` call — see the early-return at `longevity.js:18`: `if (existing.derivedAt) return existing;`), and every urgency value in the app silently drifts further from the truth the longer that goes unrun.

## Blast radius

Every consumer that ranks or gates on urgency is affected:
- `getGoalsTree()` (`server/services/identity/goals.js`)
- `jobGates.js:57`
- `goalCheckIn.js:44`
- `identity/insights.js`
- `telegram.js`
- The Character sheet's Life Goals card (reads through `getGoalsTree`)

## Fix shape

Either:
1. Recompute horizons from `birthDate` + today inside `getGoalsTree()` / wherever longevity is read, rather than trusting the stored snapshot — this is cheap arithmetic, not I/O, so there's no cost to doing it on every read. **(Preferred.)**
2. Or, stamp `derivedAt` and re-derive lazily whenever it's older than a day.

Note: the Character sheet's Life Goals card (#2675) is a read-only mirror of the goals service and must not re-derive longevity on its own — this fix belongs in the goals/longevity service layer, not the Character card.

Related to #2675.

atomantic:
## Summary

`timeHorizons.yearsRemaining` was only ever recomputed by `deriveLongevity()`, which runs on `setBirthDate` or an explicit `/longevity/derive` call (`longevity.js` early-returns the stored snapshot when `derivedAt` is set). Every consumer that ranks or gates on goal urgency therefore read a value frozen as of the last derive, drifting further from the truth each day it went unrun.

- Extracts the date math into one shared `computeTimeHorizons(birthDate, adjustedLifeExpectancy)` in `server/services/identity/markers.js`. Both the persist path (`computeLifeExpectancy`) and the new read path call it, so there is a single copy of the arithmetic rather than two that can diverge.
- Adds `applyFreshTimeHorizons(longevity, birthDate)` and `readLongevity()` to `server/services/identity/longevity.js`. Every longevity read funnels through them: `getGoals` (which is how `jobGates.js`, `goalCheckIn.js` and `telegram.js` reach goal urgency), `createGoal`, `updateGoal`, `getGoalsTree`, `getLongevity`, and the cross-insight rules in `identity/insights.js`.
- The fix stays in the goals/longevity service layer. The Character sheet's Life Goals card is untouched apart from its stale comment — it remains a read-only mirror and does not re-derive on its own.

Degradation is explicit rather than silent:

- No birth date, or no derived life expectancy to measure against, returns the stored snapshot untouched. Blanking it there would turn "we can't recompute" into "you have no horizons left", which is a worse claim than a slightly stale number.
- An unparseable birth date now yields `null` instead of a `NaN`-filled horizons object.
- The refresh in `getGoals` runs after the lazy-migration save, so this read path never triggers a write of its own.

Re-deriving is cheap arithmetic with no I/O, which is why it runs on every read (the issue's preferred option 1).

## Test plan

- New `server/services/identity/longevity.test.js` (15 cases): a snapshot derived a year ago reports a fresh `yearsRemaining` on read; `computeTimeHorizons` returns `null` for a missing/unparseable birth date and for an underived life expectancy; horizons clamp past life expectancy; `getGoalsTree` enriches urgency from today's horizons (0.7 where the stale snapshot would have scored 0) and orders goals by the re-derived value; `getGoals` refreshes urgency in memory without writing; both no-birth-date fallbacks preserve the stored values. Every birth-date fixture is computed relative to now, so none is transcribed.
- `cd server && NODE_ENV=test npm test` — 28751 passed, 252 skipped.
- `cd client && NODE_ENV=test npm test` — 7625 passed.
- `cd client && npm run lint` — clean.

Closes #4122

### DIFF
--- a/.changelog/next/fixed-issue-4122.md
+++ b/.changelog/next/fixed-issue-4122.md
@@ -1,0 +1,1 @@
+- Goal urgency and time horizons are recomputed from your birth date on every read, so `yearsRemaining` no longer stays frozen at whatever the last longevity derive stamped
--- a/client/src/components/character/GoalsCard.jsx
+++ b/client/src/components/character/GoalsCard.jsx
@@ -89,14 +89,12 @@ export default function GoalsCard() {
     // so the two surfaces can never show contradictory urgency. (Enriching the flat endpoint
     // instead would change a shared API's semantics.)
     //
-    // Two honest limits on how fresh that is, both the goals feature's to fix (see #4122),
-    // and both landing on "sorts last" rather than on a wrong number:
-    //   - The tree only re-derives for goals it considers active (`goals.js` gates on a
-    //     strict status), so the status-less goals `isActive` admits keep whatever urgency
-    //     they carry — for MortalLoom-synced goals, none at all.
-    //   - The horizons it derives against are only as fresh as the last deriveLongevity()
-    //     run. Every urgency consumer shares that, and re-deriving from this read-only card
-    //     would write on a read path AND make the sheet disagree with /goals.
+    // The horizons that urgency is derived against are now recomputed server-side on every
+    // read (#4122), so the number is current no matter how long ago deriveLongevity() ran.
+    // One honest limit remains, and it lands on "sorts last" rather than on a wrong number:
+    // the tree only re-derives for goals it considers active (`goals.js` gates on a strict
+    // status), so the status-less goals `isActive` admits keep whatever urgency they carry —
+    // for MortalLoom-synced goals, none at all.
     //
     // silent: this card owns its own error UI (the message below), so letting request() toast
     // as well would surface the same failure twice.
--- a/server/services/identity/goals.js
+++ b/server/services/identity/goals.js
@@ -15,7 +15,16 @@ import {
   loadJSON,
   saveJSON
 } from './store.js';
-import { deriveLongevity } from './longevity.js';
+import { applyFreshTimeHorizons, deriveLongevity } from './longevity.js';
+
+/**
+ * Read the longevity snapshot with `timeHorizons` re-derived against today (#4122).
+ * `goals` is passed in rather than re-read because every caller here already holds it.
+ */
+async function loadLongevityFor(goals) {
+  const longevity = await loadJSON(LONGEVITY_FILE, DEFAULT_LONGEVITY);
+  return applyFreshTimeHorizons(longevity, goals?.birthDate);
+}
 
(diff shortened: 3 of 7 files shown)


atomantic:
Review gate: two rounds with `agy`.

Round 1 findings, all three fixed in c9acc3e:
1. `typeof NaN === 'number'`, so a NaN adjusted life expectancy slipped past `computeTimeHorizons`' guard and produced `yearsRemaining: 0` / `percentLifeComplete: 100` — the exact false reading the null return exists to prevent. Now gated on `Number.isFinite`, with a test.
2. `getGoals` refreshed each goal's urgency but left `data.timeHorizons` stale, so the payload paired fresh urgencies with a stale snapshot that `dataSync` would federate onward. Now refreshed too, with a test.
3. `createGoal` / `updateGoal` loaded longevity a second time after `getGoals` had already refreshed it. They now read `goals.timeHorizons` from the record.

Round 2 raised two more; both evaluated and declined, with reasons:
1. *`updateGoal` recomputes urgency regardless of status, so a non-active goal can persist a value `getGoals` will never refresh.* Pre-existing behavior, and gating it on `status === 'active'` would make things worse for the status-less goals the Character sheet's card admits — today `updateGoal` is one of the few paths that gives them any urgency at all. Out of scope here; the card already documents the limit as "sorts last" rather than a wrong number.
2. *`for (const goal of data.goals)` in the new refresh block can throw on a goals.json with no `goals` key.* Unreachable: the pre-existing lazy-migration loop at `goals.js:203` iterates `data.goals` unguarded 36 lines earlier, so the throw already happens upstream. A guard on the new loop alone would be dead code.

PR #4225, opened 2026-08-15
fix: re-flow the dashboard grid when LayoutEditor's Move up/down reorders widgets (#4132)

### LINKED ISSUE #4132: Make LayoutEditor's Move up/down buttons actually reorder widgets
`client/src/components/dashboard/LayoutEditor.jsx:97` (`move`) reorders the layout's `widgets` array, but `Dashboard.jsx#saveLayout` persists the grid via `reconcileGrid(baseGrid, widgets)`, which iterates the grid and uses `widgets` only for membership — and the renderer reads the grid. So once a layout has a `grid` (every layout, after the first Arrange or the `synthesizeGrid` fallback is saved), those buttons produce no visible change at all. The component's own header calls itself a "keyboard-first editor… reorder uses up/down buttons," so this is a control advertising a capability it lost. The mobile reorder handle now edits the same underlying order through `reflowToOrder`, which makes the contradiction user-facing: two reorder affordances for the same widgets, one inert. (The dashboard's own `client/src/components/dashboard/CLAUDE.md` flags this directly: "LayoutEditor's own Move up/down buttons currently do not move anything once a layout has a grid.")

Fix: have `saveLayout` reflow the grid when the incoming `widgets` order differs from `readingOrderIds(existing.grid)` (both exported from `DashboardGrid.jsx` already).

This was NOT rolled into the mobile-drag change because layouts arranged before `saveGridEdit` started syncing `widgets` to reading order can have the two already out of sync, so a naive "orders differ → reflow" would re-pack a carefully arranged desktop layout the first time someone renames it or toggles a widget. Needs either a one-shot migration that stamps `widgets = readingOrderIds(grid)` onto every stored layout (`server/services/dashboardLayouts.js`, `scripts/migrations/`) before the comparison goes live, or a reorder signal from LayoutEditor that says "this save is a reorder" instead of inferring it.

Surfaced by the /simplify altitude review.

atomantic:
## Summary

LayoutEditor's Move up/down buttons reordered the layout's flat `widgets` array, but the dashboard renders from the layout's `grid` — and `reconcileGrid` iterates the grid, using `widgets` only for membership. So the buttons were inert on any layout that had a grid, which is every layout after the first Arrange (or after the `synthesizeGrid` fallback gets saved). Meanwhile the mobile reorder handle edits the same underlying order through `reflowToOrder`, leaving two reorder affordances for the same widgets with one of them doing nothing.

Fixed with an **explicit reorder signal** rather than the migration alternative the issue also floated:

- `reconcileGrid(grid, visibleIds, { reorder })` — when `reorder` is set, the reconciled grid is re-flowed to the list order through the existing `reflowToOrder`. Defaults to `false`, so the renderer and the add/remove path are unchanged.
- `LayoutEditor` passes `reordered` on `save` and `commitDuplicate`; `Dashboard.jsx` forwards it into `reconcileGrid`.

Why explicit instead of inferring it from `widgets` disagreeing with `readingOrderIds(grid)`: layouts arranged before `saveGridEdit` started syncing the two can already be out of sync, so inferring would re-pack a carefully arranged desktop layout the first time someone renamed it or toggled a widget. The alternative — a one-shot migration stamping `widgets = readingOrderIds(grid)` onto every stored layout — would have to be right on every install with no undo if it wasn't; the explicit signal keeps the change client-side and never consults the divergence at all. Rationale recorded on the issue.

`reordered` is computed at save time against an `unmovedOrder` ref: the order the draft would have had if the user had never pressed Move. `add` appends to it, `remove` filters it, `move` deliberately leaves it alone, and the rehydrate effect resets it on layout switch. That baseline can't be reconstructed from the widget list after the fact — nothing in the list records which widgets were added in this session or in what order — which is exactly why the first two attempts (a plain draft-vs-stored comparison) missed a widget added and then moved, a swap between two widgets both added in the session, and any reorder in a layout that started empty. Keeping the baseline also means a move up followed by a move back down nets out to "not a reorder", so an unrelated save still can't re-pack an arranged grid.

Also updates the note in `client/src/components/dashboard/CLAUDE.md` that documented these buttons as broken.

## Test plan

- New `client/src/components/dashboard/LayoutEditor.test.jsx` (9 cases). Flagged as a reorder: Move up, Move down, a widget added then moved up, a swap between two widgets added in the same session, a reorder in a layout that started with no widgets, and "Save as new…" carrying the signal through to `onDuplicate`. Not flagged: a rename, a remove-plus-add, and a move up followed by a move back down.
- New `reconcileGrid` cases in `DashboardGrid.test.jsx`: an arranged grid whose coordinates deliberately disagree with its widget list comes back byte-identical with the options omitted, `{}`, `{ reorder: undefined }`, and `{ reorder: false }`; with `{ reorder: true }` the reading order matches the widget list (heights preserved), including a widget added in the same save.
- Both new reorder-detection tests were verified as real regression guards by re-running them against the previous implementation — they fail there.
- `cd client && npx vitest run src/components/dashboard/` — 71 passed, run repeatedly.
- `cd client && npm test` — 7679 tests. The full-suite run has failures in 13 unrelated files (`Layout`, `Catalog`, `PostDrillConfig`, …); they are 5–14s timeouts from other work saturating this machine, not this branch. A sample of them (`Layout.test.jsx`, `PostDrillConfig.test.jsx`, `Catalog.test.jsx`) passes 102/102 in isolation on this branch, and nothing in the dashboard tree fails either way.
- `cd client && npx biome lint --error-on-warnings src/components/dashboard src/pages/Dashboard.jsx` — clean.
- No `server/` files touched.

Closes #4132


atomantic:
## Review gate — antigravity (`agy`), 3 rounds

**Round 1** — no findings on a plain "review this diff" pass; a second, more pointed pass at the same diff found that `isReordered()` filtered both lists down to the widgets they shared, so a widget added in the session and then moved up read as "not a reorder" and the grid auto-placed it at the bottom. Fixed in 7c07d79, with a regression test.

**Round 2** — found that reconstructing the baseline from the widget list can't work at all: after a move, nothing in the list records which widgets were added this session or in what order. Two cases still fell through — swapping two widgets both added in the session, and any reorder in a layout that started empty (no stored order to compare against). Fixed in 0f355ee by keeping the baseline explicitly in an `unmovedOrder` ref that `add`/`remove` update and `move` deliberately does not. Two more regression tests; both verified to fail against the previous implementation.

**Round 3** — three findings, one applied and two declined:

- *Applied* (95b42be): the rehydrate effect assigned `cur.widgets` to both the state and the ref with no `?? []`, so a layout without a widget list would leave the baseline `undefined`.
- *Declined — not a bug*: "removing a widget after moving another erases the displacement." The baseline models exactly what the grid's reading order will be after a non-reorder save (stored order, minus removals, plus additions appended), so when it matches the draft there is genuinely nothing to re-flow. The one case where information is lost is a layout whose `widgets` array already disagrees with its grid — and consulting that divergence is precisely what this issue rules out. Filtering the baseline differently doesn't recover it either.
- *Declined — pre-existing and inert*: the stale-closure note on `add()`'s `setWidgets([...widgets, id])`. The same render's `widgets` also backs the duplicate check and the widgets cap right above it, so converting only the setter would half-fix an issue that user clicks can't trigger (each click re-renders). Out of scope for this fix.

No merge from me — leaving that to the orchestrator.

PR #4253, opened 2026-08-15
fix: portal the shared Drawer overlay to <body> (#4150)

### LINKED ISSUE #4150: Make the shared Drawer overlay portal-capable (glass-theme trap)
## Section: Hand-rolled `fixed inset-0` overlays trapped by `backdrop-filter` cards (glass themes)

Root cause: the theme rule at `client/src/index.css` makes every bordered/rounded `.bg-port-card` a containing block for `position:fixed` descendants on "glass" themes (e.g. `lumen-glass-day`). Any hand-rolled `fixed inset-0` overlay rendered inside such a card is then sized to the card, not the viewport.

`client/src/components/Drawer.jsx` (~line 87, `className="fixed inset-0 z-40 bg-black/60"`) renders its backdrop inline, not portaled. It's the shared config-drawer primitive used by many pages, so any instantiation nested under a bordered `.bg-port-card` is trappable on glass themes.

## Fix shape

Audit its call sites or make the primitive portal its overlay. Precedent for the fix: `GalleryImagePicker.jsx` and the shared `Modal`'s `usePortal` option.

This is the highest-value item of the three related overlay bugs (Drawer is the shared primitive; the other two — MediaLightbox, ProcessesTab — are single call sites).

atomantic:
## Summary

`client/src/components/Drawer.jsx` rendered its `fixed inset-0 z-40 bg-black/60` backdrop and its `fixed inset-y-0 right-0` panel inline. Both are `position:fixed`, so a `filter` / `backdrop-filter` / `transform` ancestor becomes their containing block — and on the "glass" themes `client/src/index.css` gives every bordered/rounded `.bg-port-card` a `backdrop-filter`. Any Drawer instantiated under such a card was therefore sized to the *card*, not the viewport. Drawer is the shared config-drawer primitive with ~20 call sites, so the trap was reachable from many pages.

- **The Drawer overlay is now always portaled to `<body>`.** Portaling is unconditional rather than an opt-in flag (the shape the issue's `Modal` `usePortal` precedent uses) so no call site has to remember it. React still routes events and context through the React tree, so deeply nested Drawers keep working unchanged.
- **New `portalClassName` prop** for the one thing a portal genuinely leaves behind: a page-scoped CSS theming root written as *descendant* selectors. `index.css`'s City HUD palette remap is written as `.cybercity-themed <utility>`, and `CitySettingsDrawer` renders inside that root, so it now passes `portalClassName="cybercity-themed"` to keep its themed ink/surfaces.
- Documented the rule in the Config-drawer section of `client/src/CLAUDE.md`.

### Call-site audit

Grepped all 20 Drawer importers for a dependency on the backdrop being an in-tree DOM sibling:

- No `useClickOutside`/DOM-ancestor listener wraps a Drawer (React portals preserve React-tree event propagation anyway).
- No component-scoped CSS targets Drawer internals through a page ancestor; every theme rule in `index.css` is `html[data-port-theme]`-rooted, except the `.cybercity-themed` remap handled above.
- The only two call-site tests that scope through `container` (`CitySettingsDrawer.test.jsx`, `TaskConfigDrawer.test.jsx`) assert the dialog is *absent* while closed, which a portal does not change.
- `.always-dark` inside `EditAppDrawer` is on an element within the drawer body, so it travels with the portal.

## Test plan

- `client/src/components/Drawer.test.jsx` gains a `portaled overlay` block: the backdrop **and** the panel escape a `backdropFilter` ancestor into `<body>`; backdrop-click + Esc dismissal still fire through the portal; the default portal has no wrapper element; `portalClassName` wraps both the panel and the backdrop in the themed div. Queries are scoped per element (`getByTestId('glass-card')`, `parentElement` identity), never by index into a text match.
- Verified the three behavioral assertions **fail** against the pre-fix `Drawer.jsx` (`3 failed | 17 passed`) and pass after.
- `cd client && npm test` → 643 files / 7843 tests passing.
- `cd server && npx vitest run services/agentPromptBuilder` → 175 passing (the only suite that reads nested `CLAUDE.md` files).

Closes #4150


atomantic:
## Review loop

**`claude` (round 1)** — 2 low-severity findings, both triaged:

1. *Vacuous backdrop assertions in `Drawer.test.jsx`* — valid. `Node.contains(null)` returns `false`, so the "backdrop escaped the glass card" assertion would have passed even if the backdrop were dropped entirely. **Fixed** in 50e739d — the test now asserts `expect(backdrop).not.toBeNull()` first. Also dropped the `document.body.contains(dialog)` line, which was true pre-fix too (RTL's container is itself attached to `<body>`).
2. *`Drawer.jsx`'s portal call duplicates `Modal.jsx`'s three lines* — **declined as out of scope.** The two call shapes differ (Modal gates on a `usePortal` flag; Drawer is unconditional with an optional wrapper element), so a shared helper would have to carry both branches to save three lines, and retrofitting it into `Modal` would touch ~30 modal call sites for no behavior change. The SSR guard is kept rather than dropped so the two overlay primitives stay symmetrical.

Its call-site audit independently confirmed the manual one: no `<form>` ancestor at any of the ~20 Drawer call sites, the single `useClickOutside` + `Drawer` co-importer scopes its outside-click to an unrelated menu ref, `.cybercity-themed` is the only page-scoped descendant-selector root containing a Drawer, and z-order is unchanged.

**`codex review --base origin/main`** — no findings. It traced the portal behavior, call sites, the scoped City theming, focus/scroll behavior, and CSS inheritance, and ran the full client suite.

Tests green after the review fix: `cd client && npm test` → 643 files / 7843 tests.

PR #4258, opened 2026-08-15
fix: portal MediaLightbox's fixed overlay out of glass-theme cards (#4151)

### LINKED ISSUE #4151: Portal MediaLightbox's fixed overlay (glass-theme trap)
## Section: Hand-rolled `fixed inset-0` overlays trapped by `backdrop-filter` cards (glass themes)

Root cause: the theme rule at `client/src/index.css` makes every bordered/rounded `.bg-port-card` a containing block for `position:fixed` descendants on "glass" themes (e.g. `lumen-glass-day`). Any hand-rolled `fixed inset-0` overlay rendered inside such a card is then sized to the card, not the viewport.

`client/src/components/media/MediaLightbox.jsx` (~line 247, `className={\`fixed inset-0 z-50 bg-black/90 flex items-center justify-center \${overlayPad}\`}`) is not portaled. Trappable when opened from inside a themed gallery card.

## Fix shape

Portal its overlay. Precedent for the fix: `GalleryImagePicker.jsx` and the shared `Modal`'s `usePortal` option.

atomantic:
## Summary

On "glass" themes, the rule in `client/src/index.css` that gives every bordered/rounded `.bg-port-card` a `backdrop-filter` also makes that card the containing block for `position: fixed` descendants. `MediaLightbox` renders a hand-rolled `fixed inset-0` overlay — it deliberately opts out of `<ui/Modal>` (viewport-edge chevrons + its own Esc cascade), so it never inherited Modal's `usePortal` escape. Opening the lightbox from a themed gallery tile therefore sized the overlay to the card instead of the viewport.

The overlay now renders through `createPortal(…, document.body)`, the same escape `GalleryImagePicker` and `FolderPicker` get via `Modal`'s `usePortal`. Nothing else about the component changes: the focus trap, scroll lock, Esc cascade, and the nested refine/prompt-from modals all still hang off the same React subtree, so events and context are unaffected.

The two file-top comments that document the Modal opt-out are updated so the portal is discoverable from the same place as the rest of the opt-out contract.

## Test plan

- New `MediaLightbox overlay portal` test renders the lightbox inside a `.bg-port-card` ancestor carrying `backdrop-filter` and asserts the `role="dialog"` overlay is **not** inside that ancestor, is parented to `<body>`, and is absent from the component's own rendered subtree. Verified it fails against the pre-fix (non-portaled) render and passes after.
- The four existing video-element tests queried the overlay through `render()`'s `container`, which no longer holds it; they now go through a shared `videoEl()` document query.
- `cd client && npx vitest run src/components/media/ src/pages/MusicVideo.test.jsx src/a11yConventions.test.js` — 15 files / 162 tests pass (covers every in-repo suite that touches the lightbox).
- `cd client && npm test` — 645 files / 7866 tests pass.
- `cd client && npx biome lint --error-on-warnings` on both changed files — clean.

Closes #4151

### DIFF
--- a/.changelog/next/fixed-issue-4151.md
+++ b/.changelog/next/fixed-issue-4151.md
@@ -1,0 +1,1 @@
+- Media lightbox no longer gets trapped inside a gallery card on glass themes — its overlay now portals to the document body so it covers the full viewport
--- a/client/src/components/media/MediaLightbox.jsx
+++ b/client/src/components/media/MediaLightbox.jsx
@@ -1,4 +1,5 @@
 import { useEffect, useRef, useState } from 'react';
+import { createPortal } from 'react-dom';
 import {
   X, Copy, Sparkles, Film, Image as ImageIcon, Download, Eraser, Wand2,
   ChevronLeft, ChevronRight, Maximize2, Minimize2, Star, Box, ScanEye,
@@ -33,8 +34,9 @@ import { formatDateTime, formatDateNumeric } from '../../utils/formatters';
 // (A mobile tap-to-open bottom-sheet drawer existed pre-ed0e4859 and was
 // removed because it covered the image area in fullscreen.)
 // Opting out means owning the dialog semantics Modal would have supplied:
-// the overlay carries role="dialog"/aria-modal and runs useFocusTrap itself,
-// and `a11yConventions.test.js` allowlists it on that basis.
+// the overlay carries role="dialog"/aria-modal, runs useFocusTrap itself, and
+// portals to <body> in place of Modal's `usePortal` (see the render below);
+// `a11yConventions.test.js` allowlists it on that basis.
 
 const NOTE_MAX = 2000;
 const NOTE_DEBOUNCE_MS = 500;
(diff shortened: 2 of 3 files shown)


atomantic:
### Review pass — `antigravity` (`agy` 1.1.13), 2 rounds, headless against the branch diff plus the complete post-change files

**Round 1 — CHANGES_REQUESTED, 3 findings.** Round 2 (with the sole caller, the sub-modal, and the focus-trap hook supplied) — **APPROVE, no findings.** Disposition:

1. **Redundant `onKeyDown` on the overlay short-circuits the documented Esc cascade** (`refineOpen` → `promptFromOpen` → `fullScreen` → `onClose`). Focus is trapped inside the overlay, so React dispatches the synthetic handler before the window listener and `onClose()` fires unconditionally — Esc in full screen dismisses the lightbox instead of leaving full screen.

   Genuine, but **pre-existing and not a regression from this diff** — the diff adds `createPortal` and edits two comments; it touches neither Esc path, and React dispatches from its container either way, so handler order is identical before and after. Filed as #4260 with the call already made (remove the prop; verified `biome lint --error-on-warnings` stays clean without it, since the `<div>` carries `role="dialog"`) rather than widening this PR past one logical change. The reviewer agreed on round 2.

2. **Backdrop `onClick` could bubble to a clickable parent tile.** Not a finding. React portals preserve React-tree bubbling, which is exactly what happened pre-portal; the portal strictly *removes* native DOM bubbling to non-React ancestors and adds none. `MediaPreview.jsx` is the only caller and mounts the lightbox as a page-level sibling, never inside a clickable tile. Withdrawn on round 2.

3. **Focus trap would fight the portaled sub-modals.** Not a finding. `PromptRefineModal` / `PromptFromMediaModal` render `<Modal>` *without* `usePortal`, so they stay inside the overlay and inside `overlayRef`; `useFocusTrap` also binds its Tab listener to the container and only sets focus once on activation rather than continuously reclaiming it. Withdrawn on round 2.

No code changes resulted from the review, so the branch is unchanged since the PR opened.

**Guard probe:** the new portal test was run against a temporarily un-portaled render and fails there (`dialog.parentElement`, the `.bg-port-card` containment check, and the empty-subtree check all trip), so it is not vacuous.

PR #4259, opened 2026-08-15
refactor: dedup identical opus-tier rates in modelPricing EXACT_RATES (#4163)

### LINKED ISSUE #4163: Dedup identical opus-tier rate rows in modelPricing.js EXACT_RATES
In `server/lib/modelPricing.js`, all five `claude-opus-*` rows in `EXACT_RATES` (~lines 39-43) carry an identical `[5.0, 25.0]` rate pair:

```js
'claude-opus-5': [5.0, 25.0],
'claude-opus-4-8': [5.0, 25.0],
'claude-opus-4-7': [5.0, 25.0],
'claude-opus-4-6': [5.0, 25.0],
'claude-opus-4-5': [5.0, 25.0],
```

Confirmed still the case. Because of this, the `{ test: /opus/i, rateModel: 'claude-opus-5' }` line in `FAMILY_RULES` (~line 97) must be re-pointed at every model bump purely for zero behavioral effect — it only relabels the reported `rateModel` for unrecognized opus ids to whichever pointer key currently has the rates, since every opus row already resolves to the same numbers.

## Fix shape

Either:
1. Extract a shared `const OPUS_TIER_RATES = [5.0, 25.0]` and reference it from every `claude-opus-*` row plus the `FAMILY_RULES` fallback, so the pointer bump becomes unnecessary (the fallback can point at the constant directly instead of a specific model id), or
2. Let a family rule yield rates directly (`{ test: /opus/i, rates: [5.0, 25.0] }`) instead of pointing at a peer `EXACT_RATES` key.

Either retires the per-bump `FAMILY_RULES` edit that migration 206 (and every opus bump before it) had to make for no behavioral change.

Low priority — listed for completeness, not urgency. Surfaced alongside the migration-skeleton factoring item on the opus-5 PR.

atomantic:
## Summary

All five `claude-opus-*` rows in `EXACT_RATES` (`server/lib/modelPricing.js`) carried the same hand-copied `[5.0, 25.0]` literal, and the `{ test: /opus/i, rateModel: 'claude-opus-5' }` entry in `FAMILY_RULES` had to be re-pointed at the newest id on every opus bump — migration 206 and every bump before it — purely to relabel the reported `rateModel`, since every opus row already resolved to the same numbers.

The tier now owns one `OPUS_TIER_RATES` pair plus a newest-first `OPUS_MODEL_IDS` list that generates its `EXACT_RATES` rows, and the family rule takes its pointer from the head of that list. **An opus bump is now a one-line prepend** — no second edit site.

### Which fix shape, and why

The issue offered two shapes. I took shape 1 (shared constant, pointer derived rather than hand-written) instead of shape 2 (a family rule yielding `rates` inline), because `rateModel` is load-bearing downstream and shape 2 could not preserve it:

- `client/src/pages/UsagePage.jsx` prints `Priced as ${m.rateModel}` in the per-model tooltip.
- `attributedModel()` in `server/services/usageReconciler.js` compares two ids' resolved `rateModel` to decide whether a recorded launch model and a transcript model are the *same family*, and deliberately treats `null` as "not recognized, therefore not a match".

An inline-`rates` rule would have to report `rateModel: null` (breaking that comparison outright) or a synthetic label like `claude-opus` (which would stop matching the exact-table label, so `opus` and `claude-opus-5` would no longer reconcile as one model). Keeping the pointer a real `EXACT_RATES` key avoids both while still retiring the per-bump edit — and it now cannot go stale or dangle, since it is derived from the list rather than typed.

**No pricing or behavior change.** A differential run of `resolveModelRates` and `isFreeModelId` over 954 provider/model pairs (every listed id, CLI shorthands, Bedrock-prefixed and suffixed variants, unlisted opus generations, local `family:tag` ids, and the fallback paths) returns byte-identical results before and after. No client mirror of this module exists — it is server-only.

### Tests

`server/lib/modelPricing.test.js` gains an `opus tier` block pinning the three properties that made the hand-maintained pointer safe:

- every listed opus generation bills at the same tier rates (`matched: 'exact'`);
- an opus id the table has never heard of (`claude-opus-9`, `global.anthropic.claude-opus-7-…-v1:0`) still prices at $5/$25 without a pointer bump — the invariant this refactor exists to guarantee;
- the reported family label is itself a non-null, exactly-resolvable table id at the same rates, and every opus id agrees on it — the property `usageReconciler` and the usage tooltip depend on.

The existing `opus` shorthand assertion now checks the *rates* rather than hardcoding `claude-opus-5`, so the suite needs no edit at the next bump either; the new label test covers what that assertion used to.

## Test plan

- `cd server && NODE_ENV=test npx vitest run lib/modelPricing.test.js services/usageReconciler.test.js` — 89 passed.
- `cd server && NODE_ENV=test npm test` — 29127 passed. Four files reported failures under this run, all pre-existing load flakes from concurrent suites (a different, non-overlapping set failed on a second run); `services/loras.test.js`, `services/imageGenQuota.test.js`, `routes/settings.secretsStrip.test.js`, and `routes/imageGen.watermark.test.js` all pass in isolation (93 passed) and none touch pricing.
- Differential harness comparing pre- and post-change `resolveModelRates`/`isFreeModelId` across 954 pairs: 0 mismatches.

Closes #4163

### DIFF
--- a/.changelog/next/changed-issue-4163.md
+++ b/.changelog/next/changed-issue-4163.md
@@ -1,0 +1,1 @@
+- Deduped the identical Claude Opus rate rows in the usage-cost table: the tier now shares one rate pair and the opus family rule derives its pointer, so an opus model bump no longer needs a second hand-edit (#4163)
--- a/server/lib/modelPricing.js
+++ b/server/lib/modelPricing.js
@@ -31,16 +31,29 @@
 
 export const PRICING_AS_OF = '2026-07-12';
 
+/**
+ * Every shipped Claude Opus generation bills at the same published rate, so the
+ * tier owns one [input, output] pair rather than a hand-copied literal per row.
+ * `OPUS_MODEL_IDS` is ordered NEWEST FIRST: its head is the id the `/opus/i`
+ * family rule reports for opus ids the table doesn't list, so an opus bump is a
+ * single prepend here — the family rule no longer has to be re-pointed by hand
+ * (it was, at every bump through migration 206, for zero behavioral effect).
+ */
+const OPUS_TIER_RATES = [5.0, 25.0];
+const OPUS_MODEL_IDS = [
+  'claude-opus-5',
+  'claude-opus-4-8',
+  'claude-opus-4-7',
+  'claude-opus-4-6',
+  'claude-opus-4-5',
+];
+
 /** USD per 1M tokens: [input, output]. Exact model-id matches. */
 const EXACT_RATES = {
   // Anthropic
   'claude-fable-5': [10.0, 50.0],
   'claude-mythos-5': [10.0, 50.0],
-  'claude-opus-5': [5.0, 25.0],
-  'claude-opus-4-8': [5.0, 25.0],
-  'claude-opus-4-7': [5.0, 25.0],
-  'claude-opus-4-6': [5.0, 25.0],
-  'claude-opus-4-5': [5.0, 25.0],
+  ...Object.fromEntries(OPUS_MODEL_IDS.map((id) => [id, OPUS_TIER_RATES])),
   'claude-sonnet-5': [2.0, 10.0], // intro pricing through 2026-08-31 ($3/$15 after)
   'claude-sonnet-4-6': [3.0, 15.0],
   'claude-sonnet-4-5': [3.0, 15.0],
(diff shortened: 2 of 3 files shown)


atomantic:
## Review log — antigravity (`agy`), 2 rounds

**Round 1 — 2 findings, both applied in e790d697a:**

1. *Shared array reference.* `Object.fromEntries(OPUS_MODEL_IDS.map((id) => [id, OPUS_TIER_RATES]))` handed all five opus keys a reference to one array, so a future edit to a single opus row would have silently rewritten the whole tier. Each row now clones the pair, matching the hand-written rows around it.
2. *Weak test assertion.* The family-label test asserted only `expect.any(String)`. Tightened — but to `toMatch(/^claude-opus-\d/)` rather than the suggested hardcoded `toBe('claude-opus-5')`, since hardcoding the current head id would reintroduce exactly the per-bump edit this PR removes. The exact-resolution assertion immediately after is what actually pins the pointer to a real table key.

**Round 2 — `NO FINDINGS`** on the updated diff.

Re-verified after the fixes: `lib/modelPricing.test.js`, `services/usageReconciler.test.js` and `lib/index.test.js` pass (94 tests), and the differential harness still reports 0 mismatches across 954 provider/model pairs.

Issue #4170, opened 2026-08-14
Give CollapsibleText a children-based variant and migrate AgentCard onto it


atomantic:
`client/src/components/cos/tabs/AgentCard.jsx`'s `TaskDescription` (~line 80) is a second "long task text with a Show more toggle" implementation, one directory over from the shared `client/src/components/ui/CollapsibleText.jsx`. It uses a `text.length > 200` heuristic + `max-h-[3.5rem]` + gradient fade instead of measured `line-clamp-2`, and its toggle carries no `aria-expanded`/`aria-controls`.

Not consolidated when `CollapsibleText` was introduced because `AgentCard` renders *markdown blocks* via `MarkdownOutput`, which `line-clamp` can't clamp.

Decision: add a `children`-based max-height variant to `CollapsibleText` (an alternative clamp strategy for content that can't use CSS `line-clamp`, e.g. rendered markdown), then migrate `AgentCard`'s `TaskDescription` onto it and delete the bespoke implementation. This unifies the collapse affordance and its accessibility (`aria-expanded`/`aria-controls`) everywhere, at the cost of visibly changing `AgentCard`'s current fade treatment — call that out in the PR for review.

Deferred from the PR that clamped CoS task text.

Migrated from PLAN.md by /do:replan --issues.

atomantic:
#4265 was merged while its review round was still in flight, so two of the three review findings did not land with it. They ship in follow-up PR #4272:

- `hasChildren` counted `[]` / `''` as children, so a caller mapping an empty list got an empty capped box with its `text` fallback silently dropped.
- The collapsed container is a scroll container with no visible scrollbar, so tabbing to a link below the cap scrolled the preview out of reach with no way back, while `aria-expanded="false"` claimed content that was fully in the tab order was hidden. Focus inside now expands, gated on real overflow.

The third finding (effect deps missing the mode flag) landed on main in 6605399; #4272 adds the regression test it shipped without. Leaving this issue closed — #4272 is the remaining work.

PR #4272, opened 2026-08-15
fix: finish the CollapsibleText children-mode review fixes (#4170)

atomantic:
## Summary

Follow-up to #4265, which shipped the `children`-based max-height variant for `CollapsibleText` and migrated `AgentCard`'s `TaskDescription` onto it. That PR was merged while its review round was still running, so two of the three review findings didn't make it in. They ship here.

(The third — the effect deps missing the mode flag, which left the observer bound to the detached element after a `<p>` → capped `<div>` swap — landed on `main` in 6605399.)

### 1. `hasChildren` counted an empty list as children

`children != null && children !== false` is true for `[]` and `''`. A caller writing `<CollapsibleText text={fallback}>{items.map(…)}</CollapsibleText>` over an empty list got an empty capped box with its `text` fallback silently dropped — the exact opposite of what the "children wins over text" rule exists to protect against.

### 2. Focus inside the clipped region had no way back out

The collapsed container is `overflow-hidden`: a scroll container with no visible scrollbar. Children in this mode are precisely the content that holds focusable things — `MarkdownOutput` emits `<a target="_blank">`, ``, and `<pre className="overflow-x-auto">`. Tabbing to a link below the cap made the browser scroll the capped box to reveal it, and from there the top of the description was gone with no scrollbar, no wheel target, and no code path resetting `scrollTop`. Meanwhile the toggle asserted `aria-expanded="false"` over content that was fully present in the accessibility tree and the tab order.

Focus inside the capped region now expands it, which puts the target on screen and makes the `aria-expanded` claim honest. It's gated on real overflow: content that fits is never clipped, so expanding it would only mint a no-op "Show less" into the tab order.

### Also

Adds the children-mode regression test for the deps fix — 6605399 shipped that fix without one, so nothing currently guards it.

The `text` / line-clamp path is untouched; `TaskItem.jsx` (4 call sites) and `Review.jsx` (2) pass no children and are unaffected.

## Test plan

- `client/src/components/ui/CollapsibleText.test.jsx` — 3 new cases: children arriving after mount get measured (guards 6605399), an empty child list falls back to `text` and its line clamp, focus inside a clipped preview expands it, and focus inside a *fitting* preview does not.
- `cd client && npx vitest run src/components/ui/CollapsibleText.test.jsx src/components/cos/tabs/AgentCard.test.jsx` — 34 passed.
- Each new assertion was verified to fail against the pre-fix component before the fix was applied.
- `cd client && npm test` — 647 files, 7892 tests, all green on the branch this was cut from.
- `cd client && npm run lint` — clean (biome, `--error-on-warnings`).

Refs #4170, #4265


atomantic:
Review loop closed (codex, 3 rounds).

**Fixed in this PR:**
- `[null]` / `[false, undefined, '  ']` counted as children by array length, so `items.map(i => i.show ? <Row/> : null)` over an all-hidden list rendered an empty capped box and silently dropped the `text` fallback. Now counted via `Children.toArray` + a whitespace-string filter (587c0ae).
- `onFocus` gated on the `isOverflowing` state flag, which a passive effect populates *after* commit — a descendant with `autoFocus` takes focus during commit, reads a stale `false`, and stays stranded inside the clipped scroll region. Now measured live off the ref; the duplicated `scrollHeight > clientHeight + 1` predicate is extracted to one `overflows()` helper (c5e2bb8).

Both regressions have a test verified to fail against the pre-fix component.

**Declined — empty React fragments (`<></>`) count as a child.** Correct as reported, but a fragment is a legitimate child, and deciding whether an arbitrary subtree renders nothing means recursing fragments *and* evaluating components — a component returning `null` is undecidable without rendering it. No caller passes a bare empty fragment; the array case above was the one that actually occurs.

PR #4276, opened 2026-08-15
fix: wait for Claude TUI startup readiness

### LINKED ISSUE #4180: Give tuiPromptRunner the same startup-dialog dismissals as the CoS agent path
## Problem

`executeTuiRun` (`server/lib/tuiPromptRunner.js`) still gates its paste on the **old** idle heuristic — "saw output, then went quiet for `READY_IDLE_THRESHOLD_MS`", with a blind `sendPrompt('fallback')` at `PASTE_DEADLINE_MS`. It has no equivalent of the CoS agent path's positive input-ready gate or its startup-dialog dismissals.

That means any first-run/opt-in modal the provider CLI paints will be treated as "the banner finished repainting", and the prompt gets pasted into the dialog, which ignores it. The run then burns its full timeout and finalizes on whatever the screen happened to say.

The agent path (`server/services/agentTuiSpawning.js`) has already had to learn this lesson twice:

- Claude Code's folder-trust gate (`TUI_TRUST_PROMPT_PATTERN`)
- Claude Code v2.1.233's auto-mode offer (`TUI_AUTO_MODE_PROMPT_PATTERN`) — four agents died `paste-not-rendered` on 2026-08-14 before it was handled

`tuiPromptRunner` handles **neither**, so a one-shot TUI run in a folder claude has not seen, or on a CLI version showing a new opt-in, fails the same way with no dismissal.

## Why it was left out of the fix

The auto-mode fix scoped itself to the CoS agent spawner, which is where the incident happened and the only consumer of `createInputReadyTracker`. Extending `tuiPromptRunner` is a real behavior change to a separate execution path (it would newly suppress the blind-paste fallback), so it wants its own change and its own test pass rather than riding along.

## Proposed work

1. Build the shared tracker in `tuiPromptRunner` too: `createInputReadyTracker(...)` from `server/lib/tuiHandshake.js`, fed from the existing `onData` handler with both the raw and ANSI-stripped chunk.
2. Add the two dismissal branches to the `readyTimer` interval, mirroring `agentTuiSpawning.js` — `needsTrust` → `\r` (accept "Yes, I trust"); `needsAutoModeChoice` → `\x1b[B\r` (decline, then `ackAutoModeChoice()`), each behind a one-shot latch.
3. Decide the fallback posture explicitly. The agent path refuses to blind-paste and fails `tui-not-ready` at the deadline. `tuiPromptRunner` currently always blind-pastes. **Recommendation:** keep the blind-paste fallback for non-claude providers, but suppress it while `needsTrust`/`needsAutoModeChoice` is armed — pasting into a known-live modal is never right, and the hard timeout is still the backstop.
4. Tests in `server/lib/tuiPromptRunner.test.js`: a run that paints the trust gate and one that paints the auto-mode offer both dismiss it and then paste. Verify each test bites by reverting the branch.

## Acceptance

- A one-shot TUI run started in an untrusted folder dismisses the trust gate and submits its prompt.
- A run that hits the auto-mode offer declines it (option 2 — never rewrite the user's global permission default) and submits its prompt.
- No blind paste is sent while a known dialog is on screen.

Refs #4181 (the CoS-agent-side fix for the same class).

atomantic:
## Summary

- Use shared PTY readiness tracking for direct Claude one-shot runs.
- Dismiss folder-trust and auto-mode prompts before pasting, with a safe timeout if no input prompt appears.
- Cover split control sequences and preserve the non-Claude fallback behavior.

## Test plan

- `cd server && NODE_ENV=test npm test -- tuiPromptRunner.test.js tuiHandshake.test.js`
- `cd server && NODE_ENV=test npx vitest run services/agentTuiSpawning.test.js`

Closes #4180

### DIFF
--- a/.changelog/next/fixed-issue-4180.md
+++ b/.changelog/next/fixed-issue-4180.md
@@ -1,0 +1,1 @@
+- Claude one-shot prompts now wait for startup dialogs to clear before sending.
--- a/server/lib/tuiHandshake.js
+++ b/server/lib/tuiHandshake.js
@@ -342,6 +342,10 @@ export function createInputReadyTracker({ readyTextPattern = null, directLaunch 
   let needsAutoModeChoice = false;
   let autoModeAnswered = false;
   let tail = '';
+  // node-pty can split `ESC[?2004h` across reads. Keep only the trailing
+  // prefix of that exact toggle so the next chunk can complete it without
+  // treating unrelated escape traffic as a readiness signal.
+  let rawTail = '';
   return {
     // Ready once the TUI has RE-ENABLED bracketed-paste mode after the launch
     // shell turned it off to run the command — for claude that means its input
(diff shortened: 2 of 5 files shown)


atomantic:
Review gate blocked: `claude` is installed but not authenticated (`Not logged in · Please run /login`), so the required Claude CLI review could not run. Codex review completed cleanly. Please authenticate Claude and rerun the required review before merging.

PR #4279, opened 2026-08-15
test: pin declared mirror parity (#4175)

### LINKED ISSUE #4175: [repo-study-night-street] Enforce that every declared server/client mirror has a parity pin, and backfill the four that have none
**Provenance:** proposed from a study of the captured reference repo `StarKnightt/night-street` (MIT licence), studied 2026-08-14. Clean-room: no code, config or prose was copied — only the technique is described, reimplemented here against PortOS's own modules.

## The technique worth borrowing

That project names "the transcribed constant" as one of the three bug classes that cost it the most: a value duplicated into a second file, sitting under a comment *asserting* the two cannot disagree, with nothing enforcing it. The duplicate reads correct, typechecks, and holds only because two numbers happen to agree today — so the failure arrives later, silently, in a build that still looks plausible. The rules they took away are the useful part: a documented duplicate must be **findable**, and the coupling must be **proved** by a check, because a comment claiming lock-step is exactly the artifact that survives the drift.

## Why PortOS wants it

PortOS already got the hard half right. `server/lib/mirrorParity.js` exists, and eight or so mirrors are pinned by `<name>.mirror.test.js` / `<name>.parity.test.js` suites that extract the mirrored declarations from both files and diff them with comments stripped. That machinery is good and does not need replacing.

What is missing is the **enumeration guard** — nothing checks that a module *declared* a mirror actually has a pin. `client/src/lib/README.md` opens its list by saying these modules "must be kept byte-for-byte in sync with their server counterpart. The server copy is authoritative; the matching server test file is the contract." For four of them that contract does not exist:

- `client/src/lib/seasonStructure.js` — header says "Mirror of server/lib/seasonStructure.js"; no test on either side reads both copies.
- `client/src/lib/shotGrammar.js` — mirrors the `SHOT_TYPES` / `SCREEN_DIRECTIONS` controlled vocabularies, explicitly so a hand-set client value passes the server's `storyboardShotSchema`. Drift here means the storyboards editor offers a value the server then rejects. Unpinned.
- `client/src/lib/appIdentity.js` — mirrors `PORTOS_APP_ID`. Unpinned.
- `client/src/lib/issueLength.js` — header says "kept in lock-step" with the server copy, and even repeats the working bounds as a second hand-copy inside the same file. Unpinned.

This is the transcribed-constant shape precisely: a comment asserting lock-step, no mechanism. And because the gap is *coverage* rather than *capability*, the next mirror added will be unpinned too unless something enumerates them — the same reasoning that already makes `server/lib/index.test.js` fail when a new lib module is missing from the barrel or the README.

## Proposal

1. **A meta-guard test** that parses the mirror rows out of `client/src/lib/README.md` (and the equivalent server-side section), resolves each declared `server/lib/<name>.js` ↔ `client/src/lib/<name>.js` pair, and fails when no test file reads both sides. Message names the missing pin and points at `mirrorParity.js` as the tool to write it with. This makes the README the enumerable registry it is already trying to be.
   - Detect "reads both sides" structurally (a test importing/reading both paths), not by filename convention — three naming conventions are already in use (`.mirror.test.js`, `.parity.test.js`, and pins living inside a plain `<name>.test.js`, e.g. slashdoCatalog), and a filename rule would report false gaps.
   - The guard must fail **loudly on its own failure to parse** rather than quietly finding zero mirrors — a guard that silently matches nothing is the same absent-term bug in a different costume (see `project` precedent: guard tests need a bypass probe). Include a probe asserting the guard actually fires on a synthetic unpinned pair.
2. **Backfill the four missing pins** with the existing `mirrorParity` helpers.
3. Fold `issueLength.js`'s internal re-statement of its own bounds into the pinned region so the in-file duplicate is covered too.

**Fix:**
- New `server/lib/mirrorCoverage.test.js` (guard) — server-side, since the server suite already globs across the tree and the mirror pins live there.
- New pins: `server/lib/seasonStructure.mirror.test.js`, `server/lib/shotGrammar.mirror.test.js`, `server/lib/appIdentity.mirror.test.js`, `server/lib/issueLength.mirror.test.js`, all built on `server/lib/mirrorParity.js`.
- Possibly a one-line clarification in `client/src/lib/README.md` if the guard needs a stable row marker to parse.

**Test plan:** the four new pins pass against the current tree (or reveal existing drift, which is the point); the coverage guard passes on the tree and demonstrably fails against a synthetic declared-but-unpinned pair.

**Scope:** small — five test files, no production-code change unless a pin uncovers real drift.

atomantic:
## Summary

- Add a catalog-driven parity-coverage guard for direct server/client mirrors.
- Pin the previously unguarded season structure, shot grammar, app identity, and issue-length contracts.
- Extend scene-prompt coverage to compare its client mirror declaration-by-declaration.

## Testing

- `NODE_ENV=test npm test` (server)

Closes #4175

### DIFF
--- a/.changelog/next/fixed-issue-4175.md
+++ b/.changelog/next/fixed-issue-4175.md
@@ -1,0 +1,1 @@
+- Server and client mirror contracts now detect missing parity coverage before drift reaches users.
--- a/server/lib/appIdentity.mirror.test.js
+++ b/server/lib/appIdentity.mirror.test.js
@@ -1,0 +1,9 @@
+import { describe, expect, it } from 'vitest';
+import { PORTOS_APP_ID as serverAppId } from './appIdentity.js';
+import { PORTOS_APP_ID as clientAppId } from '../../client/src/lib/appIdentity.js';
+
+describe('appIdentity — server/client mirror parity', () => {
+  it('keeps the baseline app id identical', () => {
+    expect(clientAppId).toBe(serverAppId);
+  });
+});
--- a/server/lib/issueLength.mirror.test.js
+++ b/server/lib/issueLength.mirror.test.js
@@ -1,0 +1,55 @@
+import { describe, expect, it } from 'vitest';
+import {
+  CUSTOM_MINUTE_MAX as serverMinuteMax,
+  CUSTOM_MINUTE_MIN as serverMinuteMin,
+  CUSTOM_PAGE_MAX as serverPageMax,
+  CUSTOM_PAGE_MIN as serverPageMin,
+  DEFAULT_LENGTH_PROFILE as serverDefaultProfile,
+  LENGTH_PROFILES as serverProfiles,
+} from './issueLength.js';
+import {
+  CUSTOM_MINUTE_MAX as clientMinuteMax,
+  CUSTOM_MINUTE_MIN as clientMinuteMin,
+  CUSTOM_PAGE_MAX as clientPageMax,
+  CUSTOM_PAGE_MIN as clientPageMin,
+  DEFAULT_LENGTH_PROFILE as clientDefaultProfile,
+  LENGTH_PROFILES as clientProfiles,
+} from '../../client/src/lib/issueLength.js';
+
+describe('issueLength — server/client picker parity', () => {
+  it('keeps the profiles the client displays aligned with server targets', () => {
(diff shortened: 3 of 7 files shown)


atomantic:
Review gate remains unsatisfied: three headless Antigravity attempts returned no usable code-review verdict (one unrelated flag explanation, then two empty responses). The implementation and server suite are green; leaving this PR open for a reviewer retry rather than merging without the configured review.

PR #4281, opened 2026-08-15
feat: coordinate heavy local accelerator jobs

### LINKED ISSUE #4174: [repo-study-night-street] Hold a machine-wide claim for heavy local GPU jobs instead of checking headroom once at spawn
**Provenance:** proposed from a study of the captured reference repo `StarKnightt/night-street` (MIT licence), studied 2026-08-14. Clean-room: no code, config or prose was copied — only the technique is described, reimplemented here against PortOS's own modules.

## The technique worth borrowing

That project had several agents sharing one machine and one GPU. Its rule is that **every tool which renders takes a filesystem lock for the whole duration of its work and releases it even on a hard exit** — not a check at start-up, a *claim held across the run*. Their write-up is blunt about why a start-time check is not enough: two heavy GPU consumers do not fail loudly when they overlap, they each run at half speed and quietly poison every measurement taken while they do. The lock is what makes "one heavy job at a time on this box" true rather than aspirational.

## Why PortOS wants it

PortOS already knows this failure mode and has documented it: `docs/research/2026-06-13-mflux-training-watchdog-panic.md` records GPU watchdog-timeout reboots with ~21 GB of swap live. `server/services/loraTraining/memoryPrep.js` was written in response, and it is good — it unloads resident ollama/LM Studio models, measures real headroom, and refuses to start below `TRAINING_MIN_HEADROOM_GB`.

But it is a **point-in-time check at spawn only**, and it is the *only* one:

- `prepareMemoryForTraining` is imported by exactly one caller — `server/services/loraTraining/index.js:512`. Nothing else in the tree uses it.
- `server/services/imageGen/local.js` (mflux/diffusers on the same unified-memory pool) has an in-process `activeJob` guard, and no awareness whatsoever of an in-flight training run — it contains no reference to loraTraining.
- `server/services/videoGen/local.js` (mlx_video / diffusers LTX) spawns a Python child with no headroom check at all.
- `server/services/imageTo3d/trellis2.js` reasons about headroom for *quality tiering* but does not gate on the pool being free.
- Nothing reclaims the pool when a *later* job starts: a training run that legitimately passed the 24 GB gate at minute 0 can be joined at minute 40 by a local video render, an mflux batch, and a re-resident ollama model — the exact oversubscription that reboots the box, and the one arrangement the current guard cannot see.

Each subsystem defends itself in isolation; nobody owns the machine-wide resource. This is not the multi-*human* concurrency the root CLAUDE.md tells us not to defend against — it is one user's own long-running processes colliding, which that same document explicitly sanctions serializing.

## Proposal

A single machine-scoped claim for "heavy local accelerator work", taken for the duration of a job and released on exit — including hard exit.

- New `server/lib/heavyJobClaim.js`: `claimHeavyLocalJob({ kind, id, timeoutMs })` → `{ ok, holder, release }`. Backed by a lock file under `data/` carrying holder kind, id, pid and start time, so it survives across the several git worktrees that share one machine and (unlike an in-process variable) is visible to a second server process.
- **Stale-claim recovery is mandatory**, or one crashed trainer wedges local generation forever: a claim whose recorded pid is no longer alive is reclaimed, and a claim older than a generous ceiling is reported as stale in the refusal message rather than silently honoured.
- **Two acquisition modes, decided rather than deferred:** an interactive route call (user pressed Generate) gets an immediate refusal naming the current holder — so the UI can say "LoRA training is using the GPU, started 12 minutes ago" instead of hanging — while a background/scheduled caller (task-schedule, media-job hooks, autopilot) waits with a bounded timeout and then fails with the same message. Waiting silently on a user-facing request is the wrong trade; failing a scheduled render that would have succeeded 90 seconds later is also the wrong trade.
- The claim carries a **memory precondition**: `prepareMemoryForTraining`'s unload-and-measure logic generalizes to "reclaim the pool, then report headroom", so every heavy consumer benefits from what today only training gets. That means lifting the reusable part of `memoryPrep.js` out of `loraTraining/` into `server/lib/` and leaving the training-specific budget/tier maths where it is.
- Claim holders surface in the existing status/health surfaces so the user can see what owns the accelerator.

**Fix:**
- New `server/lib/heavyJobClaim.js` (+ `.test.js`), re-exported from `server/lib/index.js` with a `README.md` row.
- Promote the reclaim/measure half of `server/services/loraTraining/memoryPrep.js` to a shared module; keep `TRAINING_MIN_HEADROOM_GB` and the run-sizing logic in `loraTraining/`.
- Take the claim in `server/services/loraTraining/index.js` (around the existing check at ~:512), `server/services/imageGen/local.js#generateImage`, `server/services/videoGen/local.js`, and the `server/services/imageTo3d/trellis2.js` spawn path.
- Release in the same `close`/`exit` handlers that already clear `activeJob`, plus a process-exit sweep so a killed server does not leave a claim behind.

**Test plan:** unit tests for claim/refuse/stale-pid-reclaim/timeout-wait; per-subsystem tests that a second heavy job is refused (interactive) or waits (background) while a claim is held, and that the claim is released when the child exits, including on error.

**Scope:** medium — one new lib, one module promotion, four call sites, unit tests. No schema or on-disk record change beyond a single ephemeral lock file that is safe to delete.

atomantic:
## Summary

- add a machine-wide, file-backed claim for heavy local accelerator work with stale-PID recovery and detached-child handoff
- coordinate LoRA training, local image/video generation, and image-to-3D renders through that claim
- share the local model memory-reclamation preflight across those workloads

## Test plan

- `vitest run server/lib/heavyJobClaim.test.js server/services/loraTraining/memoryPrep.test.js server/services/imageTo3d/models.test.js server/lib/index.test.js --reporter=dot`
- `vitest run server/services/videoGen/local.test.js --reporter=dot`
- `node --check` for changed server modules

Closes #4174


atomantic:
Required Antigravity review could not complete after three attempts. The `agy` CLI is installed, but headless runs were denied the `command` permission and returned no review verdict. The implementation and focused test suites are ready for a reviewer once that permission is configured.

PR #4406, opened 2026-08-16
fix: make OpenCode installer test Windows-safe

atomantic:
## Summary
- Update the OpenCode installer unit test for the safe Windows cmd.exe /c npm.cmd launch path.
- Keep the direct npm expectation for POSIX runners.

## Test plan
- `cd server && npm test -- --run services/opencodeInstaller.test.js lib/bufferedSpawn.test.js`
- `cd server && npm test` attempted; the local PostgreSQL dependency was unavailable, so database-backed suites failed before this test-only change could be broadly validated.

### DIFF
--- a/server/services/opencodeInstaller.test.js
+++ b/server/services/opencodeInstaller.test.js
@@ -35,15 +35,25 @@ describe('OpenCode installer', () => {
 
     spawnOpenCodeInstaller({ spawnImpl });
 
-    expect(spawnImpl).toHaveBeenCalledWith(
-      'npm',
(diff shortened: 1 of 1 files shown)


atomantic:
## Review gate status

The configured Ollama review did not return a verdict. The initial HTTP request to port 5555 received an empty reply; local transport checks confirmed 5555 is HTTPS and 5553 is the HTTP mirror. Retrying the review through port 5553 ended with `ollama request failed: This operation was aborted`. No approval was inferred, so this PR remains open and unmerged.

PR #4426, opened 2026-08-16
fix: release local media claim when setup fails

### LINKED ISSUE #4364: Local video/image gen: heavy-job claim leaks if setup throws before spawn
## Problem

In both `server/services/videoGen/local.js` (~line 1553-1566) and `server/services/imageGen/local.js` (~line 548-554), the local-accelerator job flow claims a heavy-job lock (`claimHeavyLocalJob`), defines a `releaseHeavyClaim` closure, then runs several setup steps that can throw *before* `spawnDetached` — e.g. `loadHistory()` (video), reading settings via `hfChildEnv()`, resizing/temp-file prep — none of which are wrapped in a `try/finally` that calls `releaseHeavyClaim()`. `releaseHeavyClaim` is only invoked from later handlers (video: lines ~1634, 1748, 1837; image: ~584, 760) that live further down the job lifecycle, past the risky setup window.

If any of that intervening setup throws, execution never reaches a release call site, and `heavyClaim` stays held by the live server process. Every subsequent local accelerator job (video or image) is then rejected as busy (`HEAVY_LOCAL_JOB_BUSY`) until the server restarts — a single transient failure (e.g. an unreadable history file, a settings-read error) can wedge all local generation.

Caught by an automated codex review pass during the v2.44.0 release (rated P2).

## Decision

Wrap the claim in a `try { ...setup...; spawnDetached(...) } catch (err) { await heavyClaim.release(); throw err }` (or a `finally` that only releases on the non-happy-path — the happy path's release already happens correctly via the later handlers) in both `videoGen/local.js` and `imageGen/local.js`, so any throw between the claim and the point where a terminal handler takes over releases the claim before propagating.

## Affected files

- `server/services/videoGen/local.js` (claim at ~1553, setup through ~1580+ before spawn)
- `server/services/imageGen/local.js` (claim at ~548, same pattern)

## Acceptance criteria

- [ ] A thrown error anywhere between `claimHeavyLocalJob` and the point a terminal (success/failure) handler takes ownership of `releaseHeavyClaim` still releases the claim before the error propagates, in both video and image generation
- [ ] Regression test: simulate a setup-step throw (e.g. mock `loadHistory`/settings-read to reject) after a successful claim, and assert the claim is released (a subsequent claim attempt succeeds) rather than left held

atomantic:
## Summary
- Release the machine-wide heavy-job claim when local video or image setup fails before child handoff.
- Preserve terminal-handler claim ownership after a successful handoff.
- Add regression coverage for the video setup failure path.

## Test plan
- `npm test -- --run services/videoGen/local.test.js`
- `npm test -- --run services/imageGen/local.test.js`
- `npm test -- --run lib/heavyJobClaim.test.js`

Closes #4364

### DIFF
--- a/.changelog/next/fixed-issue-4364.md
+++ b/.changelog/next/fixed-issue-4364.md
@@ -1,0 +1,1 @@
+- Local video and image generation release heavy-job claims when setup fails before child handoff. (#4364)
--- a/server/services/imageGen/local.js
+++ b/server/services/imageGen/local.js
@@ -553,16 +553,24 @@ export async function generateImage({ pythonPath, prompt = '', negativePrompt = 
   }
   const releaseHeavyClaim = () => heavyClaim.release()
     .catch((err) => console.error(`❌ Image generation claim release [${jobId.slice(0, 8)}]: ${err.message}`));
-  const memoryReport = await prepareLocalMemory();
-  if (memoryReport.unloaded.length) console.log(`🧹 Image generation [${jobId.slice(0, 8)}] freed ${memoryReport.unloaded.length} resident model(s)`);
+  let proc;
+  let claimHandedOff = false;
+  try {
+    const memoryReport = await prepareLocalMemory();
+    if (memoryReport.unloaded.length) console.log(`🧹 Image generation [${jobId.slice(0, 8)}] freed ${memoryReport.unloaded.length} resident model(s)`);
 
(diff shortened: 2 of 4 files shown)


atomantic:
Review found a blocking claim-ownership race: if process handoff fails after the image/video child starts, the catch path releases the heavy-job claim while the child can continue running. Terminate and await the spawned child before releasing, or retain ownership through its terminal event. Leaving this PR open.

PR #4427, opened 2026-08-16
fix: Ollama context-window management

### LINKED ISSUE #4366: Ollama context-window management: loopback alias mismatch, multi-model max, and stale latch after daemon restart
## Problem

An automated codex review pass during the v2.44.0 release surfaced three related correctness gaps in Ollama's context-window management, all rated P2:

**1. Loopback aliases aren't normalized before comparing daemon endpoints** (`server/lib/ollamaContext.js`, `hostOf()` ~line 61-71 / `isSameOllamaDaemon()` ~line 91-94):
```js
function hostOf(baseUrl) {
  ...
  const url = URL.parse(withScheme)
  return url ? url.host.toLowerCase() : null
}
```
This lowercases the host but never maps `127.0.0.1` / `[::1]` to `localhost` (or vice versa). If a provider's configured base URL uses `127.0.0.1:11434` while PortOS's managed daemon reports `localhost:11434` (or the reverse), `isSameOllamaDaemon()` returns `false` even though they're the same daemon — `ensureOllamaAgentContext()` then wrongly classifies the local provider as remote and never applies its configured `numCtx`/warning.

**2. `getRuntimeContextLength()` takes `Math.max` across ALL resident models, not the selected one** (`server/services/ollamaManager.js` ~line 437-444):
```js
async function getRuntimeContextLength() {
  ...
  const windows = (Array.isArray(data?.models) ? data.models : [])
    .map((m) => Number(m?.context_length))
    .filter((n) => Number.isFinite(n) && n > 0)
  return windows.length ? Math.max(...windows) : null
}
```
When `/api/ps` reports multiple resident models with different `context_length`s, this can report e.g. 131K even though the model the current provider actually selected is loaded at 4K. `ensureContextWindow()` then sees the max satisfies a 64K target and skips the reload, so the harness can still hit a context overflow on the selected (smaller-window) model.

**3. `appliedContextLength` latch doesn't account for daemon restarts** (`server/services/ollamaManager.js` ~line 469-471):
```js
if (Number(appliedContextLength) >= target) {
  return { applied: false, reason: 'already-applied', contextLength: target, runtimeContextLength: appliedContextLength }
}
```
If Ollama is restarted outside PortOS while the server process stays up, `appliedContextLength` still holds the *previous* daemon's applied window. If the new daemon becomes reachable before any availability failure is observed, this check trusts the stale latch and returns `already-applied` without actually reapplying the target window to the new daemon.

## Decision

- (1) Normalize `127.0.0.1`/`[::1]`/empty-host to `localhost` (or a canonical form) inside `hostOf()` before comparing.
- (2) Have `getRuntimeContextLength()` (or its caller) look up the *selected* model's `context_length` from `/api/ps` specifically, rather than `Math.max` across all resident models — or, if the selected model can't be identified from `/api/ps`, require every relevant resident model to individually meet the target before treating the window as satisfied.
- (3) Track daemon identity (e.g. a PID or a boot-time/session token learnable via `/api/version` or process tracking) alongside `appliedContextLength`, and invalidate the latch when the daemon identity changes — or simply re-verify the live runtime context via `getRuntimeContextLength()` before trusting the latch, rather than trusting an in-memory value across an external restart.

## Affected files

- `server/lib/ollamaContext.js` (`hostOf`, `isSameOllamaDaemon`)
- `server/services/ollamaManager.js` (`getRuntimeContextLength`, `ensureContextWindow`, `appliedContextLength`)

## Acceptance criteria

- [ ] `isSameOllamaDaemon('http://127.0.0.1:11434', 'http://localhost:11434')` (and the `[::1]` equivalent) returns `true`
- [ ] With two resident models at different context lengths, `ensureContextWindow(target)` correctly reloads when the *selected* model's window is below target, even if another resident model's window is above it
- [ ] After an external Ollama restart, `ensureContextWindow(target)` reapplies the target window instead of trusting a stale `appliedContextLength` latch
- [ ] Tests covering each of the three scenarios above

atomantic:
## Summary

- Canonicalize Ollama loopback aliases so local daemon ownership is recognized consistently.
- Select the requested model's runtime context window, with a conservative multi-model fallback.
- Revalidate the context latch after a resident daemon disappears, and pass selected models through all Ollama agent launch paths.

## Test plan

- `cd server && npm test -- --run lib/ollamaContext.test.js services/ollamaManager.test.js services/ollamaAgentContext.test.js services/agentCliSpawning.test.js services/agentTuiSpawning.test.js services/runner.test.js`
- Result: 6 test files, 253 tests passed.

Closes #4366

### DIFF
--- a/.changelog/next/fixed-issue-4366.md
+++ b/.changelog/next/fixed-issue-4366.md
@@ -1,0 +1,1 @@
+- Ollama context-window management now canonicalizes loopback aliases, respects the selected model's runtime limit, and recovers after daemon restarts.
--- a/server/lib/ollamaContext.js
+++ b/server/lib/ollamaContext.js
@@ -63,11 +63,24 @@ function hostOf(baseUrl) {
   if (!raw) return null
   // A bare `host:port` (the OLLAMA_HOST convention) is not a valid URL, so give
   // it a scheme before parsing rather than string-comparing the two spellings.
-  const withScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(raw) ? raw : `http://${raw}`
+  const withScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(raw)
+    ? raw.replace(/^(https?:\/\/):(?=\d|$)/i, '$1localhost:')
+    : /^:\d+(?:\/.*)?$/.test(raw)
+      ? `http://localhost${raw}`
+      : `http://${raw}`
   // `URL.parse` returns null on invalid input instead of throwing (Node ≥22.1;
   // the repo's engine floor is 22.12), so no try/catch is needed here.
   const url = URL.parse(withScheme)
-  return url ? url.host.toLowerCase() : null
+  if (!url) return null
+
+  // Ollama treats loopback aliases as the same local daemon. Compare a
+  // canonical hostname plus the parsed port so 127.0.0.1, [::1], and an empty
+  // OLLAMA_HOST hostname cannot make a local provider look remote.
+  const hostname = url.hostname.replace(/^\[|\]$/g, '').toLowerCase()
+  const canonicalHostname = ['127.0.0.1', '::1', 'localhost'].includes(hostname)
+    ? 'localhost'
+    : hostname
+  return `${canonicalHostname}${url.port ? `:${url.port}` : ''}`
 }
 
 /**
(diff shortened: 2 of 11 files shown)


atomantic:
Review found a blocking restart-loop risk: treating an empty  model list as proof that the Ollama daemon was replaced also fires after normal model eviction and immediately after a context-window restart. Use reliable daemon identity/health evidence instead of model absence. Leaving this PR open.

atomantic:
Correction: review found a blocking restart-loop risk. Treating an empty Ollama resident-model list as proof that the daemon was replaced also fires after normal model eviction and immediately after a context-window restart. Use reliable daemon identity or health evidence instead of model absence. Leaving this PR open.

PR #4429, opened 2026-08-16
fix: persist calendar event selection in URL (#4382)

### LINKED ISSUE #4382: Put calendar event selection in the URL
## Problem / Goal

Calendar tabs are already `/calendar/:tab`, but clicking an event only sets `selectedEvent` in memory. Refresh, a shared link, and voice cannot reopen `EventDetail`. Events already have stable ids.

This breaks the project contract: selection lives in the URL (`client/src/CLAUDE.md`).

## Context

Four views each own the same local state:

- `client/src/components/calendar/AgendaTab.jsx` ~48, ~143, ~191
- `DayView.jsx`, `WeekView.jsx`, `MonthView.jsx` — same `selectedEvent` / `setSelectedEvent` pattern
- Route is `/calendar/:tab` only (`App.jsx`)

Row keys already use `` `${event.accountId}-${event.id}` ``, so ids are not unique across accounts.

Copy the Jira reports pattern (`JiraReports.jsx` ~132–166) / `useUrlParams`.

## Decision

Use a single query param `?event=<accountId>:<id>` so one value is unique and shareable. Close deletes the param. Resolve against the already-fetched list; if the event is not in the list, leave the drawer closed (no new get-by-id API in this issue).

Keep the param on the current tab (`/calendar/agenda?event=…`, `/calendar/week?event=…`, …) so tab switches can keep the selection if the implementer wants, but it is acceptable for the first slice to drop the param on tab change if that is simpler. Decision: **keep the param across tab changes** so a shared week-view link still works after the user flips to agenda.

## Proposed approach

1. Extract a tiny `useCalendarEventParam()` (or inline `useSearchParams` the same way in each view) that reads/writes `event`.
2. On click, set `event` to `${accountId}:${id}`.
3. Drive the existing `EventDetail` drawer from the resolved list item.
4. Close / Escape deletes the param.

## Acceptance criteria

- [ ] `/calendar/agenda?event=<accountId>:<id>` opens that event's drawer on a fresh load when the event is in the fetched list
- [ ] Refresh keeps the drawer open
- [ ] Closing the drawer removes `event` from the URL
- [ ] Agenda, day, week, and month views share the same param contract
- [ ] Existing calendar tests updated

## Out of scope

A new get-event-by-id API. Changing `/calendar/:tab` routing. Recurring-event expansion.

atomantic:
## Summary
- Persist calendar event selection in the `event` query parameter using the account and event IDs.
- Restore the EventDetail drawer from the URL across agenda, day, week, and month views.
- Preserve the selection query parameter when switching calendar tabs and clear it when closing the drawer.

## Test plan
- `cd client && npm test`
- `cd client && npm run lint`

Closes #4382

### DIFF
--- a/.changelog/next/fixed-issue-4382.md
+++ b/.changelog/next/fixed-issue-4382.md
@@ -1,0 +1,1 @@
+- Calendar event drawers now reopen from shareable URLs and survive refreshes.
--- a/client/src/components/calendar/AgendaTab.jsx
+++ b/client/src/components/calendar/AgendaTab.jsx
@@ -6,6 +6,7 @@ import socket from '../../services/socket';
 import EventDetail from './EventDetail';
 import { formatTimeOfDay as formatTime, formatWeekdayDate } from '../../utils/formatters';
 import BrailleSpinner from '../BrailleSpinner';
+import useUrlParams from '../../hooks/useUrlParams';
 
 const RSVP_STYLES = {
   accepted: 'bg-port-success/20 text-port-success',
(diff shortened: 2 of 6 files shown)


atomantic:
Review found incomplete deep-link hydration: day/week/month views only search their current visible-range result, so a shared event URL outside that range does not reopen its drawer. Load the referenced event by account/id or encode/restore the relevant range. Leaving this PR open.

PR #4440, opened 2026-08-17
feat: add agent context contract evaluator (#4413)

### LINKED ISSUE #4413: [repo-study-blume-agent-eval] Add black-box contract evaluation for the agent context surface
## Provenance

Studied repository: haydenbleasel/blume
License: MIT License
Study date: 2026-08-16

This is a clean-room technique proposal. Reimplement the behavior against PortOS modules; do not copy Blume source, configuration, prose, or assets.

## Rationale

Blume's evaluation runner makes agent-facing quality testable as a black box: cases declare a question, expected facts, routes/scope, severity, and skip state; each case runs against a fresh minimal context with only the documented read tools; results distinguish pass, fail, error, and skip and can enforce a threshold. A separate judging step can explain failures, but the contract does not depend on the judge being trusted with the project filesystem.

PortOS already evaluates generated Digital Twin behavior in `server/services/digital-twin-testing.js`, `digital-twin-values-testing.js`, `digital-twin-adversarial-testing.js`, and `digital-twin-multi-turn-testing.js`. Those suites measure model responses, not whether a read-only agent context surface returns the right safe records, honors scope, omits forbidden fields, or remains honest when data is stale. That gap becomes important if #4411 is implemented.

## Fix

Fix: after the agent-context contract exists, add a declarative, fixture-backed evaluator (for example `server/services/agentContextEval.js` plus `server/services/agentContextEval.test.js` and an explicit CLI or route) that invokes the context tools through their public contract. Cases should cover search/list/get/navigation, localization or profile scope if supported, redaction/forbidden-field assertions, result and token limits, unknown-record behavior, and stale-snapshot signaling. Use sanitized fake records and a fresh isolated fixture set per case; never read the live Brain, Privacy Vault, or a real workspace during the suite.

Produce machine-readable JSON with per-case pass/fail/error/skip, source/tool pointers, and a configurable failure threshold suitable for focused tests and CI. An optional reader/judge pass may add explanations only from an explicit user-triggered provider/model selection; it must not run at boot, batch-fill a cache, or make a cold-bootstrap LLM call. Add regression cases for the privacy boundary and schema/manifest parity so advertised tools cannot drift from runtime behavior.

Estimated scope: heavy model reasoning and high effort for the contract schema, fixture isolation, privacy assertions, and CI/report integration; this is a follow-on to #4411, not a replacement for PortOS's existing Digital Twin suites.

atomantic:
## Summary

- add a declarative, fixture-backed black-box evaluator over the public agent-context manifest and tool-call contract
- isolate every case from live PortOS data and report machine-readable pass/fail/error/skip results, severity, source/tool pointers, and a configurable failure threshold
- add explicit source freshness and approximate-token budgets to contract schema v2, plus privacy-field allowlisting and coverage for manifest/runtime parity
- expose the evaluator as `npm run --silent eval:agent-context -- --failure-threshold 0` and document its CI behavior

## Test plan

- `npm run --silent eval:agent-context -- --failure-threshold 0`
- `npm test --prefix server -- --run services/agentContextEval.test.js services/agentContextMcp.test.js lib/agentContextValidation.test.js routes/agentContextMcp.test.js`
- `node --check scripts/agent-context-eval.js`
- `node --check server/services/agentContextEval.js`
- `node --check server/services/agentContextMcp.js`

Closes #4413


atomantic:
Merge blocked by required CI. The one allowed same-SHA retry also failed in Windows server unit tests at server/services/systemResources.test.js:204: expected status `unavailable`, received `ready`. CI Gate is consequently red. PR left open for follow-up.

PR #4507, opened 2026-08-17
fix: keep wiki note selection in URL (#4380)

### LINKED ISSUE #4380: Put wiki note selection in the URL
## Problem / Goal

Opening a wiki note is stored in `location.state`, not the URL. Refresh, a shared link, ⌘K, and voice cannot restore the note. Browser Back can also re-fire a stale `openNote` from history state.

This breaks the project contract: selection lives in the URL (`client/src/CLAUDE.md`).

## Context

Vault selection is already `?vault=` (`Wiki.jsx`). Note open is not:

- `client/src/components/wiki/tabs/OverviewTab.jsx` ~65 — `navigate('/wiki/browse', { state: { openNote: note.path } })`
- `SearchTab.jsx` and `GraphTab.jsx` do the same
- `BrowseTab.jsx` ~41–46 reads `location.state?.openNote` and calls `handleSelectNote`

`api.getNote(vaultId, path)` already exists. `useUrlParams` / `useSearchParams` is the pattern (vault selection on this page, Jira reports elsewhere).

## Decision

Use a `note` query param next to the existing `vault` param. Encode the path (`encodeURIComponent`) because wiki paths contain slashes. Closing the note deletes the param. Do not keep `location.state` as a fallback.

## Proposed approach

1. In `BrowseTab`, read `note` from the query string (preserve `vault`). When it is set and differs from the open note, call `handleSelectNote(decodeURIComponent(note))`.
2. `handleSelectNote` / the close-note path write/delete the `note` param (`replace: true` on close so Back does not re-open).
3. Replace every `state: { openNote }` navigation with a browse URL that includes `?note=`.
4. Drop the `location.state` effect.

## Acceptance criteria

- [ ] `/wiki/browse?vault=…&note=<encoded-path>` opens that note on a fresh load
- [ ] Refresh keeps the same note open
- [ ] Closing the note removes `note` from the URL
- [ ] Overview / Search / Graph no longer pass `state: { openNote }`
- [ ] Existing wiki tests updated for the query-param contract

## Out of scope

Changing vault selection. Adding a new notes API. Encoding anything other than the path.

atomantic:
## Summary

- Store wiki note selection in the URL alongside the selected vault.
- Open notes from Overview, Search, and Graph with the `note` query parameter.
- Clear the note parameter when closing the mobile note view, including browser-safe replacement navigation.

## Test plan

- `npm test -- --run src/pages/Wiki.test.jsx src/components/wiki/tabs/BrowseTab.test.jsx`
- `npm run build`

Closes #4380

### DIFF
--- a/client/src/components/wiki/tabs/BrowseTab.jsx
+++ b/client/src/components/wiki/tabs/BrowseTab.jsx
@@ -1,5 +1,5 @@
 import { useState, useEffect, useCallback, useRef } from 'react';
-import { useLocation } from 'react-router';
+import { useSearchParams } from 'react-router';
 import * as api from '../../../services/api';
 import {FileText, FolderOpen,
   ChevronDown, ChevronRight, ArrowLeft, Tag, Link2, Edit3, Save,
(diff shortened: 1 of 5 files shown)


atomantic:
Required external review was attempted with ollama[hf.co/empero-ai/Qwythos-9B-Claude-Mythos-5-1M-GGUF:BF16]~max=2, but the reviewer returned HTTP 404 because that model is unavailable. No .findings verdict was produced; leaving this PR open for follow-up.

PR #4532, opened 2026-08-18
feat: allow editing media prompts

atomantic:
## Summary
- add validated prompt PATCH endpoints for image sidecars and video history records
- add prompt editing to the shared media lightbox for imported, uploaded, and generated media
- keep the media history list updated after a successful save

## Test plan
- npm test -- --run routes/imageGen.test.js routes/videoGen.test.js (server)
- npm test -- --run components/media/MediaLightbox.test.jsx components/media/MediaCard.test.jsx (client)
- npm run lint (client)

### DIFF
--- a/client/src/components/media/MediaLightbox.jsx
+++ b/client/src/components/media/MediaLightbox.jsx
@@ -106,6 +106,7 @@ export default function MediaLightbox({
   hasNext = false,
   annotation = null,
   onAnnotationChange,
+  onPromptChange,
   variantGroup = null,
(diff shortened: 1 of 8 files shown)


atomantic:
The configured local Ollama reviewer was invoked with model qwen3.8:27b-mlx and low effort, but the review endpoint returned no body after retries (including the local HTTPS endpoint and its HTTP mirror). No findings verdict was available, so this PR is intentionally left open rather than merged without a valid review response.

PR #4537, opened 2026-08-18
fix: recover stale chunks in route errors (#4525)

### LINKED ISSUE #4525: Route missing errorElement lets React Router's default error screen bypass the app's stale-chunk auto-recovery
## Problem / Goal

Users occasionally see a raw browser-native error dump — React Router's own
"Unexpected Application Error!" screen, with a stack trace like
`TypeError: Failed to fetch dynamically imported module: .../assets/ChiefOfStaff-*.js`
— and have to manually reload the page. This should never happen: PortOS already
has a purpose-built system to detect exactly this failure (a stale chunk hash
after a redeploy) and recover automatically. The goal is for this class of error
to always resolve itself with a silent reload, never a visible crash screen.

## Context

`client/src/main.jsx:60-82` wires a single splat route via `createBrowserRouter`
with **no `errorElement`**:

```js
const router = createBrowserRouter([
  { path: '*', element: (<><App /><Toaster ... /></>) },
]);
```

When no `errorElement` is set, React Router falls back to its own built-in
`DefaultErrorComponent`, which literally renders `"Unexpected Application Error!"`
(`client/node_modules/react-router/dist/development/lib/hooks.js:626-644`,
`defaultErrorElement` at line 644). Critically, this internal error boundary lives
*inside* `RouterProvider`'s render tree — more deeply nested than the app's own
class-based `ErrorBoundary` (`client/src/components/ErrorBoundary.jsx`), which
wraps `ThemeProvider`/`RouterProvider` from the outside
(`client/src/main.jsx:86-90`). React resolves an error to the *nearest* ancestor
boundary, so React Router's own default boundary catches render-phase errors
first — the outer `ErrorBoundary`'s `componentDidCatch` (lines 16-17, which
contains the stale-chunk recovery call) never runs for these errors.

This is a real gap in an otherwise complete recovery system. PortOS already has
three cooperating layers for exactly this failure mode:

- `client/src/utils/staleChunkReload.js` — `isStaleChunkError()` (cross-browser
  message matching) and `reloadOnceForStaleChunk()` (build-id-aware one-shot
  guard via `sessionStorage`, purges service-worker caches only when the server
  confirms a different build, then `window.location.reload()`).
- `client/src/utils/lazyWithReload.js` — wraps every lazy-loaded route
  (`lazyWithReload(() => import('./pages/ChiefOfStaff'))`, e.g.
  `client/src/App.jsx:43`) with `importWithRetry` (2 retries w/ backoff) before
  treating a still-failing stale-chunk error as fatal.
- `client/src/main.jsx:22-24` (`vite:preloadError` listener) and `:28-32`
  (`unhandledrejection` listener) also call the same recovery for failures
  outside the React render tree.
- `client/src/components/ErrorBoundary.jsx:16-17` — the last-resort catch, same
  recovery call.

These fire correctly for a preload 404 or a rejected promise. The gap is
specifically the **render-phase throw** — when `lazyWithReload`'s retries are
exhausted and it re-throws inside the `<Suspense>` boundary during render, React
Router's own internal boundary (not the app's) catches it first, because the
route has no `errorElement` of its own.

## Proposed approach

- Add an `errorElement` to the splat route in `client/src/main.jsx:60-82` so
  route-level render errors are handled explicitly by app code instead of
  falling through to React Router's default UI.
- The new error element should reuse the existing recovery primitives rather
  than duplicate them: read the thrown error via React Router's `useRouteError()`,
  call `isStaleChunkError(error)` from `staleChunkReload.js`, and on a match call
  `reloadOnceForStaleChunk()` — the same branch `ErrorBoundary.jsx`'s
  `componentDidCatch` already runs. For a non-stale-chunk error, render the same
  (or equivalent) "Something went wrong" fallback UI already defined in
  `ErrorBoundary.jsx`, so there's one visual crash screen, not two divergent
  ones.
- Consider extracting the shared "is this a stale chunk → recover, else render
  fallback" logic out of `ErrorBoundary.jsx` into a small helper both the class
  component's `componentDidCatch`/`render` and the new function-based route
  error element can call, rather than copy-pasting the branch.

## Acceptance criteria

- [ ] The splat route in `client/src/main.jsx` has an `errorElement` so React
      Router's built-in "Unexpected Application Error!" screen can never surface
      to a user.
- [ ] A dynamic-import/chunk-load failure that reaches this boundary (e.g. via
      a test that throws a stale-chunk-shaped error during render) triggers the
      same `reloadOnceForStaleChunk()` recovery as the other existing recovery
      paths, with no visible error dump in the common case.
- [ ] A genuinely different (non-stale-chunk) render error still shows a
      friendly fallback UI matching `ErrorBoundary.jsx`'s existing fallback, not
      a raw stack trace.
- [ ] Existing tests (`client/src/utils/staleChunkReload.test.js`,
      `client/src/utils/lazyWithReload.test.js`) still pass; add coverage for
      the new route-level error element's stale-chunk branch.
- [ ] Manually verified in the dev server: forcing a stale-chunk-shaped throw
      during a route render results in an automatic reload, not a visible crash
      screen.

## Out of scope

- Reworking the single-splat-route / `useBlocker` architecture
  (`client/src/main.jsx:54-59`) — this is an `errorElement` addition to the
  existing route, not a route-tree redesign.
- The socket `build:id` proactive "new version available" hook
  (`server/services/socket.js:208`) — a separate, currently-unwired mechanism
  for prompting a reload out of band; worth its own issue if wanted.
- The app self-update / GitHub-release mechanism
  (`server/services/updateChecker.js`) — unrelated to chunk staleness.

atomantic:
## Summary

- Complete the route-level error fallback's stale-chunk recovery path using the existing reload primitives.
- Trigger the build-aware one-shot reload for render-phase dynamic-import failures while preserving the friendly fallback for other route errors.
- Add regression coverage for stale-chunk detection and recovery invocation.

## Test plan

- `client`: `RouteErrorFallback`, `staleChunkReload`, and `lazyWithReload` tests (29 passing).
- Biome lint and `git diff --check` (passing).

The splat route's explicit `errorElement` is already present on `main` from commit `aa31e47cb`; this PR completes its missing stale-chunk behavior.

Closes #4525

### DIFF
--- a/client/src/components/RouteErrorFallback.jsx
+++ b/client/src/components/RouteErrorFallback.jsx
@@ -1,6 +1,8 @@
 import { AlertTriangle } from 'lucide-react';
+import { useEffect } from 'react';
 import { useNavigate, useRouteError, isRouteErrorResponse } from 'react-router';
 import Banner from './ui/Banner';
+import { isStaleChunkError, reloadOnceForStaleChunk } from '../utils/staleChunkReload';
 
 const getErrorMessage = (error) => {
   if (isRouteErrorResponse(error)) return error.statusText || `Request failed (${error.status})`;
@@ -13,6 +15,10 @@ export default function RouteErrorFallback() {
   const error = useRouteError();
   const navigate = useNavigate();
   const message = getErrorMessage(error);
+
+  useEffect(() => {
+    if (isStaleChunkError(error)) reloadOnceForStaleChunk();
+  }, [error]);
 
   return (
     <div className="min-h-dvh-cap bg-port-bg flex items-center justify-center p-4">
--- a/client/src/components/RouteErrorFallback.test.jsx
+++ b/client/src/components/RouteErrorFallback.test.jsx
@@ -1,8 +1,14 @@
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { render, screen, act } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { createMemoryRouter, Outlet, RouterProvider } from 'react-router';
 import RouteErrorFallback from './RouteErrorFallback';
+import { isStaleChunkError, reloadOnceForStaleChunk } from '../utils/staleChunkReload';
+
+vi.mock('../utils/staleChunkReload', () => ({
+  isStaleChunkError: vi.fn(),
+  reloadOnceForStaleChunk: vi.fn(),
+}));
 
 const renderRouteError = async (error) => {
   const router = createMemoryRouter([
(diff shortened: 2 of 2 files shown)


atomantic:
Required Ollama reviewer was unavailable: the configured model returned API 404 (model not found). Per the review gate, this PR remains open with its claim markers and worktree intact for human follow-up; no merge was attempted.

PR #4549, opened 2026-08-18
fix: guarantee a closing section on MiniMax Music 3 lyric sheets

atomantic:
## Summary

MiniMax Music 3 renders were ending abruptly, with no conclusion — they cut off
mid-phrase at an arbitrary time instead of resolving the song.

The root cause is the model's own architecture, not a duration mis-prompt: MiniMax
Music 3 treats `audio_duration` as a hard *ceiling* (`max_frames = duration * 25`
in the MLX sidecar; `audio_duration` = "maximum requested duration" upstream). The
composition actually resolves where the global LLM emits its end token — "the
language model may stop earlier" — and the only lever PortOS has over *when* that
happens is the **section tags in the lyric sheet**.

The existing "pace MiniMax instrumental renders" fix only auto-built a section
sheet *with* an `[outro]` in the **instrumental fallback** branch. But the common
case — a vocal song where the user supplies their own lyrics — passed those lyrics
**verbatim** through `buildSidecarArgs`, so a draft that ended on `[verse]`/
`[chorus]`/`[bridge]` handed the model no closing cue, and it resolved mid-phrase.

## Changes

- `ensureClosingSection()`: append an `[outro]` cue to whatever sheet reaches the
  model, idempotently — a sheet that already closes on `[outro]` is untouched, and
  the user's own words are never edited (we only ever *append*).
- An `ensureOutro` engine flag on `minimax-music3` and `minimax-music3-mlx` so only
  the MiniMax engines get it; ACE-Step and the non-lyric engines are unaffected.
- Applied at the `buildSidecarArgs` seam, so it covers every caller — the Music
  studio, the pipeline audio stage, and autopilot — not just the studio.
- Detection matches the existing `hasOutro` heuristic in `musicDuration.js`
  (word-boundary, case-insensitive) so the two sides stay in agreement.
- Tests in `musicGen.test.js`: the caller-supplied sheet now closes on `[outro]`;
  a sheet that already has an outro is unchanged; a dedicated `ensureClosingSection`
  block covers idempotency, case-insensitivity, inline tags, and non-string input.

## Test plan

- `npx vitest run server/services/pipeline/musicGen.test.js` — 81 passed.
- `npx vitest run server/lib/musicDuration.test.js server/lib/musicDuration.mirror.test.js server/services/audioGen/local.test.js` — passed.
- The `music.test.js` route suite is un-runnable in this isolated CoS worktree
  (uninstalled deps — `Cannot find package 'express'`); it fully mocks `musicGen`
  and asserts the route's raw user-lyric snapshot, which this change neither touches
  nor is meant to (sheet assembly is the generator's job).

### DIFF
--- a/server/services/pipeline/musicGen.js
+++ b/server/services/pipeline/musicGen.js
@@ -311,15 +311,20 @@ export const ENGINES = Object.freeze({
     // documented section tags. Engines without this field (ACE-Step) accept an
     // empty string and render an instrumental themselves.
     //
-    // The sheet is built per-render rather than fixed, because `audio_duration`
-    // is only a CEILING for this engine — see buildMinimaxInstrumentalLyrics.
-    instrumentalLyrics: buildMinimaxInstrumentalLyrics,
-    // Auto mode sizes that ceiling from lyric words/sections and leaves room
-    // for an ending. MiniMax may still stop earlier by design.
-    autoDuration: true,
-    customModels: false,
-    fixedModelInstall: true,
-    cudaRequired: true,
+     // The sheet is built per-render rather than fixed, because `audio_duration`
+     // is only a CEILING for this engine — see buildMinimaxInstrumentalLyrics.
+     instrumentalLyrics: buildMinimaxInstrumentalLyrics,
+     // Guarantee a closing section on the sheet (see ensureClosingSection) so the
+     // model resolves its ending instead of cutting off mid-phrase at an
+     // arbitrary time. Applied only to this engine — it is the one whose
+     // resolution point is an end-token the user's draft may not cue.
+     ensureOutro: true,
+     // Auto mode sizes that ceiling from lyric words/sections and leaves room
+     // for an ending. MiniMax may still stop earlier by design.
+     autoDuration: true,
+     customModels: false,
+     fixedModelInstall: true,
+     cudaRequired: true,
     executionProfile: 'cuda-bf16-auto-experimental',
     vramProfiles: MINIMAX_MUSIC3_VRAM_PROFILES,
     benchmarkGuide: MUSIC_RENDERER_BENCHMARK_GUIDE,
@@ -338,9 +343,10 @@ export const ENGINES = Object.freeze({
     resolvePython: resolveMinimaxMusic3MlxPython,
     venvDefault: MINIMAX_MUSIC3_MLX_VENV_DEFAULT,
     installEnv: 'INSTALL_MINIMAX_MUSIC3_MLX',
-    healthProbe: 'import mlx; from mlx_audio.music import load',
+     healthProbe: 'import mlx; from mlx_audio.music import load',
     lyrics: true,
     instrumentalLyrics: buildMinimaxInstrumentalLyrics,
+    ensureOutro: true,
     autoDuration: true,
     customModels: false,
     fixedModelInstall: true,
@@ -558,6 +564,40 @@ export function buildMinimaxInstrumentalLyrics(durationSec) {
   return ['intro', ...body, 'outro'].map((tag) => `[${tag}]`).join('\n');
 }
 
+// A section tag alone on (or leading) a line, matching the same detection the
+// duration heuristic uses (`analyzeMusicLyrics`).
+const MINIMAX_LEADING_TAG = /^\[([^\]\r\n]+)\]/;
+// A closing cue, case-insensitive and word-bounded, so `[outro]` and `[outro
+// fade]` count but `[outroduction]` does not — identical to the
+// `hasOutro` signal in server/lib/musicDuration.js.
+const MINIMAX_OUTRO_TAG = /^outro\b/i;
+
+/**
+ * Guarantee a closing section on a MiniMax Music 3 lyric sheet.
+ *
+ * MiniMax resolves its ending wherever the global LLM emits an end token (its
+ * autoregressive loop breaks there — "the language model may stop earlier"), and
+ * the only pacing lever is the sheet's section tags. A sheet whose last section
+ * is a verse/chorus/bridge gives the model no cue for "this is where the song
+ * ends", so it can resolve mid-phrase at an arbitrary time — the "renders end
+ * abruptly, no conclusion" report. Appending an explicit `[outro]` hands the
+ * model a closing structure without touching any section the user wrote.
+ *
+ * Idempotent: a sheet that already carries an outro tag is returned unchanged, so
+ * the instrumental fallback (which already closes with `[outro]`) is untouched and
+ * a user's own `[outro]` is never duplicated.
+ */
+export function ensureClosingSection(lyrics) {
+  const text = typeof lyrics === 'string' ? lyrics : '';
+  const hasOutro = text.split(/\r?\n/).some((raw) => {
+    const tag = MINIMAX_LEADING_TAG.exec(raw.trim());
+    return tag && MINIMAX_OUTRO_TAG.test(tag[1]);
+  });
+  if (hasOutro) return text;
+  const trimmed = text.replace(/\s+$/, '');
+  return trimmed ? `${trimmed}\n[outro]` : '[outro]';
+}
+
 /**
  * Build the `{ bin, args }` for a backend's sidecar. Pure — unit-tested without
  * spawning Python. All sidecars share the same base flag contract
@@ -586,16 +626,20 @@ export function buildSidecarArgs({ engineId = DEFAULT_ENGINE_ID, pythonPath, scr
     '--runtime-dir', runtimeDir ?? engine.runtimeDir,
   ];
   if (engine.lyrics) {
-    // Preserve the caller's string verbatim when they supplied one; substitute
-    // only when it is absent or whitespace, and only for an engine that cannot
-    // accept empty lyrics (see instrumentalLyrics). The substitute may be a
-    // builder that sizes the sheet to the render length.
+     // Preserve the caller's string verbatim when they supplied one; substitute
+     // only when it is absent or whitespace, and only for an engine that cannot
+     // accept empty lyrics (see instrumentalLyrics). The substitute may be a
+     // builder that sizes the sheet to the render length.
     const provided = typeof lyrics === 'string' ? lyrics : '';
     const fallback = typeof engine.instrumentalLyrics === 'function'
-      ? engine.instrumentalLyrics(seconds)
-      : (engine.instrumentalLyrics || '');
-    args.push('--lyrics', provided.trim() ? provided : fallback);
-  }
+       ? engine.instrumentalLyrics(seconds)
+       : (engine.instrumentalLyrics || '');
+     const sheet = provided.trim() ? provided : fallback;
+      // For engines whose resolution point is an end-token the draft may not cue
+      // (see ensureClosingSection), guarantee a closing section on whatever sheet
+      // reaches the model — a caller's or the fallback's — idempotently.
+    args.push('--lyrics', engine.ensureOutro ? ensureClosingSection(sheet) : sheet);
+   }
   return { bin: pythonPath, args };
 }
 
--- a/server/services/pipeline/musicGen.js
+++ b/server/services/pipeline/musicGen.js
@@ -311,20 +311,21 @@ export const ENGINES = Object.freeze({
     // documented section tags. Engines without this field (ACE-Step) accept an
     // empty string and render an instrumental themselves.
     //
-     // The sheet is built per-render rather than fixed, because `audio_duration`
-     // is only a CEILING for this engine — see buildMinimaxInstrumentalLyrics.
-     instrumentalLyrics: buildMinimaxInstrumentalLyrics,
-     // Guarantee a closing section on the sheet (see ensureClosingSection) so the
-     // model resolves its ending instead of cutting off mid-phrase at an
-     // arbitrary time. Applied only to this engine — it is the one whose
-     // resolution point is an end-token the user's draft may not cue.
-     ensureOutro: true,
-     // Auto mode sizes that ceiling from lyric words/sections and leaves room
-     // for an ending. MiniMax may still stop earlier by design.
-     autoDuration: true,
-     customModels: false,
-     fixedModelInstall: true,
-     cudaRequired: true,
+    // The sheet is built per-render rather than fixed, because `audio_duration`
+    // is only a CEILING for this engine — see buildMinimaxInstrumentalLyrics.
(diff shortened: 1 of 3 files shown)


server/services/pipeline/musicGen.js
```diff
@@ -558,6 +564,40 @@ export function buildMinimaxInstrumentalLyrics(durationSec) {
...
+ */
+export function ensureClosingSection(lyrics) {
+  const text = typeof lyrics === 'string' ? lyrics : '';
+  const hasOutro = text.split(/\r?\n/).some((raw) => {
+    const tag = MINIMAX_LEADING_TAG.exec(raw.trim());
+    return tag && MINIMAX_OUTRO_TAG.test(tag[1]);
+  });
+  if (hasOutro) return text;
+  const trimmed = text.replace(/\s+$/, '');
+  return trimmed ? `${trimmed}\n[outro]` : '[outro]';
```
atomantic, line 600: **[CRITICAL]** `ensureClosingSection` does not actually guarantee a closing section, and diverges from the parity it claims with `hasOutro` in `server/lib/musicDuration.js`.

1. The `.some()` scan treats an outro tag **anywhere** in the sheet as satisfying the guarantee. A sheet like `[outro]\n[chorus]\nbig finish` still ends on a chorus — exactly the "renders end abruptly, no conclusion" failure this function exists to fix — but returns unchanged because an outro tag exists earlier in the sheet. Only the *last* tagged section should count.
2. `MINIMAX_OUTRO_TAG.test(tag[1])` never trims the captured tag body, so `[ outro]` / `[outro  ]`-with-inner-padding is not recognized as a closing tag even though `musicDuration.js:44` (`hasOutro`) does trim before testing — the comment above claims these are "identical", but they disagree on padded tags, producing a duplicate `[outro]\n[outro]` on a sheet the duration heuristic already considers closed.

Verified: `ensureClosingSection('[outro]\n[chorus]\nbig finish')` returns the input unchanged (bug 1); `ensureClosingSection('[verse]\nend\n[ outro]')` → `'[verse]\nend\n[ outro]\n[outro]'` (bug 2).

Independently confirmed by all 5 review agents (Surface Scan, Surface Quality, Security, Cross-File Tracing, Cross-File Contract).

```suggestion
export function ensureClosingSection(lyrics) {
  const text = typeof lyrics === 'string' ? lyrics : '';
  const lastTag = text
    .split(/\r?\n/)
    .map((raw) => MINIMAX_LEADING_TAG.exec(raw.trim()))
    .filter(Boolean)
    .at(-1);
  if (lastTag && MINIMAX_OUTRO_TAG.test(lastTag[1].trim())) return text;
  const trimmed = text.replace(/\s+$/, '');
  return trimmed ? `${trimmed}\n[outro]` : '[outro]';
}
```

atomantic (commented):
Reviewed by /do:review — 1 critical, 7 improvements (6 posted inline, 1 out-of-diff).

## Highlights
- **`server/services/pipeline/musicGen.js:590-598`** — `ensureClosingSection` doesn't actually guarantee a closing section: an outro tag anywhere in the sheet (not just the last section) short-circuits the append, so a sheet like `[outro]\n[chorus]\nbig finish` still ends on a chorus — the exact failure this PR exists to fix. It also fails to trim the captured tag body before testing, diverging from the `hasOutro` parity the comment claims with `server/lib/musicDuration.js`, so a padded tag like `[ outro]` gets a duplicate `[outro]` appended. All 5 review agents converged on this independently.
- **Indentation churn** — a large fraction of both changed files' lines (many of them pre-existing, untouched-content lines) are re-indented by one extra space/level relative to their siblings, all 5 agents flagged it, and `server/` has no linter to catch it. Suggested fixes posted at the two clearest examples; the pattern recurs throughout both files (see inline comments for the full list of locations).
- **Misleading test title** (`musicGen.test.js:870`) — "recognizes an inline outro tag and a bare word" pins the opposite of what it asserts.

## Out-of-diff observations
- **`client/src/components/music/MusicGenPanel.jsx:639-641`** (not touched by this PR) — the client still shows "No [outro] section detected... an explicit [outro] marker gives the ending clearer structure" for exactly the two engines (`minimax-music3`, `minimax-music3-mlx`) that this PR makes the server append `[outro]` to unconditionally. The hint is gated on `autoDurationAvailable`, which is true only for those same two engines, so after this PR the hint can never be accurate for the engine it renders on. Consider rewording it (e.g. "PortOS appends a closing cue for you; add your own `[outro]` to control where it lands") or exposing `ensureOutro` through `musicEngineCatalog.js` / `musicEngineCapabilities.js` so the client can gate the hint correctly.
- **Test coverage gap** — `ensureOutro` never appears in `musicGen.test.js` (confirmed via grep); no test exercises the flag on `minimax-music3-mlx` specifically (its one `buildSidecarArgs` test only asserts `--lyrics` is present, never its value), and no test covers the mid-sheet-outro case or a whitespace-padded tag (`[ outro]`) — exactly the two cases the CRITICAL finding above depends on. Worth adding once the fix lands.
- **Auto-duration ceiling computed pre-append** (`musicGen.js:690` calling `recommendMinimaxDurationSec(lyrics, ...)` before `buildSidecarArgs` appends the outro section) — the appended section isn't budgeted into the ceiling it's sized against. Low impact: the unconditional 20s `ENDING_CUSHION_SEC` and the 10s rounding step in `musicDuration.js` absorb the ~3.6s (3s × 1.2 safety multiplier) difference in the common case; only matters near the 300s cap. Not blocking, noted for awareness.
- **Minor**: `musicGen.test.js:779-782` has two assertions (`sheet.startsWith(words)`, `sheet.split('\n').at(-1)`) that are strictly implied by the exact-equality assertion three lines above and can never fail independently of it.

## Coherence check
The PR title and diff are aligned: it adds `ensureOutro` scoped to exactly the two MiniMax Music 3 engine variants (the only engines with `lyrics: true` + `instrumentalLyrics`), and the new `ensureClosingSection` function plus tests match the stated intent. The gap is in the implementation of the guarantee itself (see Highlights above), not in what the PR claims to do.

_Generated by /do:review_

server/services/pipeline/musicGen.js
```diff
@@ -311,15 +311,20 @@ export const ENGINES = Object.freeze({
...
+     // model resolves its ending instead of cutting off mid-phrase at an
+     // arbitrary time. Applied only to this engine — it is the one whose
+     // resolution point is an end-token the user's draft may not cue.
+     ensureOutro: true,
+     // Auto mode sizes that ceiling from lyric words/sections and leaves room
+     // for an ending. MiniMax may still stop earlier by design.
+     autoDuration: true,
+     customModels: false,
+     fixedModelInstall: true,
+     cudaRequired: true,
```
atomantic: **[IMPROVEMENT]** Every added/modified line in this block sits at 5-space indent (including several pre-existing lines re-indented with no functional change) while every sibling property in the same object literal — and the rest of the file — uses 4 spaces. Flagged independently by all 5 review agents. `server/` has no linter configured (only `client/` runs biome), so nothing in CI catches this; it roughly doubles the size of the functional diff and pollutes `git blame`.

This is one instance of a pattern repeated at `musicGen.js:346`, `musicGen.js:629-642`, and throughout `musicGen.test.js` (`244-248`, `769-796`, `838-842`, `863-868`, `875-879`) — worth a single re-indent pass over the whole diff before merge rather than fixing line-by-line.

```suggestion
    // The sheet is built per-render rather than fixed, because `audio_duration`
    // is only a CEILING for this engine — see buildMinimaxInstrumentalLyrics.
    instrumentalLyrics: buildMinimaxInstrumentalLyrics,
    // Guarantee a closing section on the sheet (see ensureClosingSection) so the
    // model resolves its ending instead of cutting off mid-phrase at an
    // arbitrary time. Applied only to this engine — it is the one whose
    // resolution point is an end-token the user's draft may not cue.
    ensureOutro: true,
    // Auto mode sizes that ceiling from lyric words/sections and leaves room
    // for an ending. MiniMax may still stop earlier by design.
    autoDuration: true,
    customModels: false,
    fixedModelInstall: true,
    cudaRequired: true,
```

server/services/pipeline/musicGen.js
```diff
@@ -338,9 +343,10 @@ export const ENGINES = Object.freeze({
     resolvePython: resolveMinimaxMusic3MlxPython,
     venvDefault: MINIMAX_MUSIC3_MLX_VENV_DEFAULT,
     installEnv: 'INSTALL_MINIMAX_MUSIC3_MLX',
-    healthProbe: 'import mlx; from mlx_audio.music import load',
+     healthProbe: 'import mlx; from mlx_audio.music import load',
```
atomantic: **[IMPROVEMENT]** Same indentation drift as above — a whitespace-only re-indent (4→5 spaces) of an otherwise-untouched line.

```suggestion
    healthProbe: 'import mlx; from mlx_audio.music import load',
```

server/services/pipeline/musicGen.js
```diff
@@ -586,16 +626,20 @@ export function buildSidecarArgs({ engineId = DEFAULT_ENGINE_ID, pythonPath, scr
...
   ];
   if (engine.lyrics) {
-    // Preserve the caller's string verbatim when they supplied one; substitute
-    // only when it is absent or whitespace, and only for an engine that cannot
-    // accept empty lyrics (see instrumentalLyrics). The substitute may be a
-    // builder that sizes the sheet to the render length.
+     // Preserve the caller's string verbatim when they supplied one; substitute
+     // only when it is absent or whitespace, and only for an engine that cannot
+     // accept empty lyrics (see instrumentalLyrics). The substitute may be a
+     // builder that sizes the sheet to the render length.
```
atomantic: **[IMPROVEMENT]** This comment ("Preserve the caller's string verbatim ... substitute only when it is absent or whitespace") is now stale for `ensureOutro` engines — line 641 below can append `[outro]` to a caller-supplied sheet, so the string is no longer always preserved verbatim. Also re-indented to 5 spaces (see the block-wide indentation note).

```suggestion
    // Preserve the caller's string verbatim when they supplied one; substitute
    // only when it is absent or whitespace, and only for an engine that cannot
    // accept empty lyrics (see instrumentalLyrics). The substitute may be a
    // builder that sizes the sheet to the render length. `ensureOutro` engines
    // may still append a closing section below.
```

server/services/pipeline/musicGen.js
```diff
@@ -586,16 +626,20 @@ export function buildSidecarArgs({ engineId = DEFAULT_ENGINE_ID, pythonPath, scr
...
-    args.push('--lyrics', provided.trim() ? provided : fallback);
-  }
+       ? engine.instrumentalLyrics(seconds)
+       : (engine.instrumentalLyrics || '');
+     const sheet = provided.trim() ? provided : fallback;
+      // For engines whose resolution point is an end-token the draft may not cue
+      // (see ensureClosingSection), guarantee a closing section on whatever sheet
+      // reaches the model — a caller's or the fallback's — idempotently.
+    args.push('--lyrics', engine.ensureOutro ? ensureClosingSection(sheet) : sheet);
+   }
```
atomantic: **[IMPROVEMENT]** Indentation drift (4/5/6-space mix within one `if` block, closing brace at column 3 for a block opened at column 2) plus a 3-line comment that mostly restates the line beneath it — per CLAUDE.md, comments should lead with the non-obvious *why*, not the *what* the code already says.

```suggestion
    const sheet = provided.trim() ? provided : fallback;
    // ensureOutro engines resolve on an end-token the draft may not cue —
    // guarantee one on whatever sheet reaches the model, idempotently.
    args.push('--lyrics', engine.ensureOutro ? ensureClosingSection(sheet) : sheet);
  }
```

server/services/pipeline/musicGen.test.js
```diff
@@ -820,6 +838,43 @@ describe('buildMinimaxInstrumentalLyrics', () => {
...
+       // "outroduction" has no word boundary after "outro", so it is a different
+       // tag — matching the hasOutro detection in server/lib/musicDuration.js.
+     expect(ensureClosingSection('[outroduction]')).toBe('[outroduction]\n[outro]');
+     expect(ensureClosingSection('\n[final-outro]')).toBe('\n[final-outro]\n[outro]');
+      });
+
+  it('recognizes an inline outro tag and a bare word', () => {
+     expect(ensureClosingSection('no tags at all\n[outro]')).toBe('no tags at all\n[outro]');
+     expect(ensureClosingSection('outro')).toBe('outro\n[outro]');
+    });
```
atomantic: **[IMPROVEMENT]** Test title claims the opposite of what it verifies. The second assertion proves a bare, unbracketed `outro` is *not* recognized (a cue gets appended); "inline" is also inaccurate — `MINIMAX_LEADING_TAG` is `^`-anchored and can never match mid-line, so `[outro]` on its own line (first assertion) is a leading tag, not an inline one. Flagged independently by 4 of the 5 review agents.

```suggestion
  it('treats a leading bracketed tag as a closing section but not a bare unbracketed word', () => {
    expect(ensureClosingSection('no tags at all\n[outro]')).toBe('no tags at all\n[outro]');
    expect(ensureClosingSection('outro')).toBe('outro\n[outro]');
  });
```

server/services/pipeline/musicGen.test.js
```diff
@@ -240,8 +241,11 @@ describe('buildSidecarArgs', () => {
     });
     expect(args).toContain('MiniMaxAI/MiniMax-Music3');
     expect(args.slice(args.indexOf('--duration'), args.indexOf('--duration') + 2)).toEqual(['--duration', '300']);
-    expect(args.slice(args.indexOf('--lyrics'), args.indexOf('--lyrics') + 2)).toEqual(['--lyrics', 'Example lyrics']);
-  });
+      // A sheet with no closing section gets an [outro] cue appended so the model
+      // resolves its ending; the caller's words still lead the sheet.
+    expect(args.slice(args.indexOf('--lyrics'), args.indexOf('--lyrics') + 2))
+         .toEqual(['--lyrics', 'Example lyrics\n[outro]']);
+    });
```
atomantic: **[IMPROVEMENT]** Same indentation drift pattern as `musicGen.js` — the file's convention is 2-space, this block sits at 5/6/9-space with a mismatched closing brace. One representative fix; the same pattern recurs at `769-796`, `838-842`, `863-868`, `875-879`.

```suggestion
    // A sheet with no closing section gets an [outro] cue appended so the model
    // resolves its ending; the caller's words still lead the sheet.
    expect(args.slice(args.indexOf('--lyrics'), args.indexOf('--lyrics') + 2))
      .toEqual(['--lyrics', 'Example lyrics\n[outro]']);
  });
```

Issue #4519, opened 2026-08-17
Add periodic full-reconciliation pass for Brain federation sync to catch silent divergence


atomantic:
## Problem

Brain federation sync (`server/services/brainSync.js`, `syncOrchestrator.js`) is delta-log/cursor-based only. The Instances page (`client/src/pages/Instances.jsx`) shows "N to push" figures, which are BIGSERIAL cursor positions, not verified row-level parity between peers. If a sync cycle silently drops or fails to apply a record (partial write, schema mismatch swallowed somewhere, a peer offline during a window), there is currently no mechanism that detects the resulting divergence — both peers report "synced" while actually holding different data. This was called out as a known gap during PRD generation for the project (2026-08-17).

## Decision

Add an opt-in, schedulable full-reconciliation pass per sync category:

- A new reconciliation routine compares record-id sets (and a cheap content checksum, e.g. hash of `updatedAt` + a stable field subset) between two peers for a given entity type, rather than just advancing cursors.
- Surface divergence as an actionable diff (record IDs present on one side only, or present on both with mismatched checksums) rather than a silent pass/fail.
- Run it as a CoS-schedulable job (consistent with existing `cosSchedule` patterns) so it's opt-in per install, not a forced background cost — default off, user enables per peer/category.
- Surface the last-reconciliation result and any open divergence on the Instances page next to the existing cursor counts.

## Affected files

- `server/services/brainSync.js`, `server/lib/schemaVersions.js` — sync category definitions live here
- `server/services/instances.js`, `server/routes/instances.js` — peer registry and sync trigger routes
- `client/src/pages/Instances.jsx`, `client/src/components/instances/` — surfacing reconciliation state in the UI

## Acceptance criteria

- A reconciliation run against two peers with intentionally-diverged test records (one inserted directly on peer B, bypassing sync) reports the divergence, not a clean bill of health.
- A reconciliation run against two fully-synced peers reports zero divergence.
- The routine does not run automatically on every sync cycle — it is a distinct, schedulable/on-demand operation.

atomantic:
Partially shipped in #4550 (merged).

**Done ✓**
- `GET /api/brain/reconcile/manifest` — one `{ id, updatedAt, deleted }` row per record per entity type, tombstones included, ids and clocks only (no record bodies).
[...]

PR #4550, opened 2026-08-18
Audit brain federation for record-level divergence, not just cursor positions

atomantic:
## Summary

The Instances page reported "synced" from delta-log cursor positions, which only say "we pulled everything you published" — never that the two brains actually hold the same records. Anti-entropy reconcile (`brainReconcile.js`) does heal a whole-brain checksum mismatch by pulling the peer's snapshot, but it is silent and pull-only: it names nothing, and it cannot fix records the *other* side is missing. A peer that stopped syncing this direction stayed divergent while both cards read green.

This adds an on-demand parity audit that makes divergence visible:

- **`GET /api/brain/reconcile/manifest`** — each side serves one `{ id, updatedAt, deleted }` row per record per entity type, tombstones included. Ids and LWW clocks only: no record bodies, no names, so an audit never pulls brain content the peer wouldn't otherwise receive.
- **`server/services/brainParity.js`** — diffs local vs peer manifests *per entity type* (brain ids are only unique within a store) through the existing `computeRecordIntegrity` in `server/lib/syncIntegrity.js`, classifying every record as in-parity / local-only / peer-only / diverged. It also compares the whole-brain checksum, which catches the one case a manifest cannot see: matching ids **and** matching `updatedAt` over different record bodies.
- **`POST /api/brain/reconcile/parity`** — runs the audit for one peer or sweeps every federating peer. **Not wired into the sync cycle by design** — a routine 60s sync must never pay for a full manifest exchange.
- **`GET /api/brain/reconcile/parity`** — the last stored result per peer, keyed by `instanceId`. A local file read, so the page renders parity state on load without contacting anyone.
- **`BrainParityPanel`** on each peer card — verdict pill, per-status counts, and the actual out-of-parity record ids grouped by entity type, with a Check button to re-run.

Backward compatible in both directions: a peer too old to serve the manifest is reported as `peer-too-old` only after falling back to projecting a manifest out of its existing `/reconcile/snapshot`, so parity is auditable against any peer supporting reconcile at all. Older peers simply never call the new route. `peer-unreachable` stays distinct from `peer-too-old` so the UI never tells you to upgrade a peer that is merely offline.

Storage: `data/brain_parity_reports.json`, classified `ephemeral-file` and documented in `docs/STORAGE.md` — a regenerable point-in-time observation, machine-local, never federated, with per-type id lists capped at 25 plus a `truncated` flag so a badly diverged install can't grow the file without bound.

## Test plan

- `cd server && NODE_ENV=test npx vitest run services/brainParity.test.js routes/brainSync.test.js` — 32 tests. Covers the acceptance criteria directly: a peer holding a record we don't reports the divergence with the record id; two fully-synced peers report zero divergence and a matching checksum; a checksum mismatch with zero divergent records is surfaced rather than reported clean. Also covers the snapshot fallback, unreachable-vs-too-old, malformed peer rows, cross-type id collisions, tombstone agreement vs one-sided deletes, the storage cap, the sweep filter, and that an unreachable peer costs no local brain read.
- `cd server && NODE_ENV=test npx vitest run services/brainReconcile.test.js services/brainSync.test.js services/syncOrchestrator.test.js lib/index.test.js lib/syncIntegrity.test.js` — 111 tests, no regressions in the existing sync paths.
- `cd client && npx vitest run src/components/instances src/services src/lib/syncCounts.test.js` — 164 tests, including 9 new `BrainParityPanel` cases (verdict states, expanded record listing, inline error handling, and that rendering the panel never triggers an audit).
- `cd client && npx biome lint --error-on-warnings src/components/instances/BrainParityPanel.jsx src/pages/Instances.jsx src/services/apiBrain.js` — clean.

## Remaining

Refs #4519. The issue also asks for the audit to be runnable as a **CoS-schedulable job** (default off, per peer/category). `runBrainParityCheck()` with no `peerId` is exactly that unit of work — it sweeps every federating peer and persists the results — but it is not yet registered as a schedulable job or exposed as an opt-in toggle. Wiring it into the autonomous-jobs registry (`server/services/autonomousJobs/`) with a default-disabled entry, plus a settings affordance, is the remaining slice.

Refs #4519


atomantic:
Review loop notes for the record.

**codex** (effort=high) raised three findings, all fixed in c49702fb4:
- P1: an unverified checksum fell through to the green `in parity` verdict. Fixed — that case now reads `unverified` and states that record bodies were not compared.
- P2: the content-mismatch note claimed the next sync cycle would pull the peer's copy. It would not — `applyRemoteRecord` skips on an equal `updatedAt`, so equal-clock body drift is a standoff in both directions. The note now says so and gives the user the actual remedy.
- P2: a card keyed by the local peer id kept a stale fresh report when the install behind that address was replaced. Fixed with a reset on `instanceId` change.

**antigravity** (effort=medium) raised three findings in its first round, all fixed in 8657ea281: the report-key fallback the Instances page was missing, defensive access on a stored report with no summary, and a sweep that held one in-memory report map across every peer (now a read-modify-write per peer, so a mid-sweep single-peer check is not clobbered and an interrupted sweep keeps what it got through).

Its confirming round returned one finding — `ReferenceError: checksum is not defined` in `compareChecksums` — which does not reproduce. The line reads `local === peerChecksum`, and `brainParity.test.js` asserts both the `match: true` and `match: false` paths against a reachable peer. Partial-context artifact, no change made.

PR #4587, opened 2026-08-19
fix: make OrcaRouter API key setup discoverable

atomantic:
## Summary
- Add a direct action from OpenCode OrcaRouter wrappers to edit the sibling OrcaRouter API provider.
- Explain that users paste the key in the API Key field and do not configure an environment variable.
- Document the internal `ORCAROUTER_API_KEY` handoff and cover the flow with a UI regression test.

## Test plan
- `cd client && npm test -- --run src/pages/AIProviders.test.jsx`
- `cd client && npm run lint`
- `cd client && npm run build`


atomantic:
Antigravity review was attempted with the required `agy` CLI, but the provider returned: `Individual quota reached` (reset reported in about 40 minutes). The binary is installed; this is a reviewer-service quota block, so no substitute reviewer was used and the PR remains open pending the required review.

Issue #4619, opened 2026-08-20
Show the traceback when a setup-image-video.sh install probe fails, instead of discarding it 14 times


atomantic:
## Problem

Every runtime install in `scripts/setup-image-video.sh` verifies itself with an import probe, and all but two throw the diagnostic away:

```bash
if ! "${LTX2_PY}" -c "import ltx_pipelines_mlx" 2>/dev/null; then
  echo "❌ LTX-2 MLX synced but the import failed." >&2
```

`2>/dev/null` appears at lines 149, 170, 198, 294, 335, 387, 639, 684, 724, 757, 855, 895, 921, 944. When one of these fails the user gets a headline with no cause — a missing system library, a wheel that resolved for the wrong arch, and a genuinely broken pin all read identically, and the traceback that would say which is discarded before anyone sees it.

Two sites already diverge, in two different ways:

- `:807` re-runs the failed probe under `2>&1 | tail -5 >&2` — correct output, but it executes the probe twice, so a non-deterministic failure can print a different cause than the one that failed.
- `:449` (MiniMax H3, added in #4606) captures `2>&1` into a variable up front and tails it on the failure branch — one execution, right output.

So there are now three spellings of one problem, and the next contributor picks from three.

## What to do

Add one helper beside `have()` (around `:39`) and convert the sites to it:

```bash
# Run an install probe, and on failure show WHY. Every probe here fails for more
# than one reason — a missing system library, a wheel resolved for the wrong
# arch, a moved pin — and the traceback is the only thing that tells them apart.
# Captured rather than re-run: a probe that fails non-deterministically must not
# report a different cause than the one that failed.
probe_or_fail() {
  local headline="$1" hint="$2"; shift 2
  local out
  if ! out="$("$@" 2>&1)"; then
    echo "❌ ${headline}" >&2
    printf '%s\n' "$out" | tail -n 5 | sed 's/^/   /' >&2
    echo "   ${hint}" >&2
    exit 1
  fi
}
```

Then each site becomes one line, e.g.:

```bash
probe_or_fail "LTX-2 MLX synced but the import failed." \
  "Use Repair / Upgrade from the Video Gen runtime panel to retry." \
  "${LTX2_PY}" -c "import ltx_pipelines_mlx"
```

Convert all 16 sites (the 14 `2>/dev/null` ones, the `:807` re-run, and the `:449` capture) so one pattern remains. Success-path behaviour must not change: the probe still runs once and prints nothing when it exits 0.

[...]

Issue #4618, opened 2026-08-20
Surface the BYOV runtime probe's stderr on failure instead of discarding it at both call sites


atomantic:
## Problem

Both call sites that run `scripts/minimax_h3_runtime_probe.py` throw its stderr away, so the probe can only ever communicate an exit code:

- `server/services/videoGen/runtimes.js#runVenvProbe()` — `stdio: ['ignore', 'ignore', 'ignore']`
- `scripts/setup-image-video.sh:445` — `2>/dev/null`, then prints a fixed `❌ MiniMax H3 MLX synced but its pipeline import failed.`

The probe writes actionable single-line diagnostics naming exactly which import or seam failed. None of it reaches the user or the logs. Whoever bumps the MLX pin gets "its pipeline import failed" and has to re-run the probe by hand to find out why — which is most of the diagnostic gap #4606 set out to close.

## What to do

Capture the probe's stderr and surface it on failure at both sites:

- `runVenvProbe()`: switch stderr to `'pipe'`, accumulate it with a small cap (a few KB — the probe emits single lines, and an unbounded buffer on a wedged child is its own problem), and on a non-zero exit log it in the repo's single-line emoji style. Keep the resolve contract a boolean; the message goes to the log, not the return value. Do not pipe stdout — nothing reads it, and an unread pipe can fill and block the child.
- `scripts/setup-image-video.sh`: drop the `2>/dev/null` so the probe's own lines land beside the existing `❌` message, or capture and re-echo them under it.

Note that `runVenvProbe()` is shared by every BYOV runtime probe, not just MiniMax H3 — the improvement applies to all of them, and the change should stay generic rather than special-casing one runtime.

## Notes

Deferred from #4606 (which moved the seam assertions into the probe): making the probe fail at install time was the fix; making its *reason* visible is this separate change to the call sites.

Refs #4606

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

atomantic:
Shipped in #4637.

Two notes on scope while working this:

- The `scripts/setup-image-video.sh` half of this issue is already fixed on `main` — the runtime probe there now captures with `2>&1` and tails the last lines under the `❌`. Only `runVenvProbe()` still discarded stderr, so that is all #4637 touches.
[...]

PR #4620, opened 2026-08-20
fail a MiniMax H3 pin bump at Install/Repair, not on a failed keyframe render

### LINKED ISSUE #4606: Check MiniMax H3 pin-drift seams at install time, not on a failed keyframe render
## Problem

`scripts/generate_minimax_h3.py` carries four adapters that patch methods of the pinned MiniMax H3 MLX text encoder (`install_key_prefix_map`, `install_pil_image_processor`, `install_vision_weight_sanitizer`, `install_vision_embed_merge`). Each is only correct against the exact pinned implementation, so each guards its seam: the hook must still exist, `processor` must still be a property, and `encode`'s source must still hash to `PINNED_ENCODE_DIGEST`.

Those guards fire at **render** time. `verify_runtime_checkout()` already byte-verifies the checkout against `MINIMAX_H3_EXPECTED_REVISION` before any of it runs, so the guards are unreachable for a stock user — their only real audience is whoever bumps the pin. That person finds out after a cache resolve and a pipeline load, on a failed keyframe render, rather than at the action that caused it.

## What to do

Move the seam assertions into `scripts/minimax_h3_runtime_probe.py`, which already runs at Install/Repair and already imports `mlx_vlm.models.qwen3_vl.language`. Assert there that:

- `MiniMaxH3TextEncoder.processor` exists and is a `property`
- `MiniMaxH3TextEncoder._load_weights` exists
- `MiniMaxH3TextEncoder.encode` exists, still contains `PINNED_BROADCAST_MERGE`, and still hashes to `PINNED_ENCODE_DIGEST`
- `mlx_vlm.models.qwen3_vl.qwen3_vl.Model` is importable (the merge correction needs it; the probe does not currently import it)
- `mlx_vlm.utils.sanitize_weights` is importable

Keep the render-time guards — they are the last line for an overridden pin — but surface the same failure at install, where it is actionable. The constants should have one home; hoisting them into a small module both the runner and the probe import is preferable to duplicating the digest.

## Why it was deferred

Out of scope for the fix that introduced the adapters (#PR): that change is confined to the runner and its suite, and the guards it added are correct where they are — just later than they could be.

atomantic:
## Summary

PortOS patches four adapters onto PipeNetwork's pinned MiniMax H3 MLX text encoder, and each is only correct against the exact implementation it was written for. Those seam guards only fired at **render** time — so whoever bumps `MINIMAX_H3_EXPECTED_REVISION` found out after a cache resolve and a pipeline load, and only on the image-conditioned path three of the four serve.

- The pin facts (`PINNED_BROADCAST_MERGE`, `PINNED_ENCODE_DIGEST`, the seam table, `pinned_encoder_hook`) move to one module, `scripts/_minimax_h3_pin.py`, which the runner and the install probe both import. The digest now has exactly one home — a second copy is a copy that goes stale.
- `minimax_h3_runtime_probe.py --verify-seams` asserts every seam up front: `_wanted` / `processor` (still a `property`) / `_load_weights` / `encode` (still contains the broadcast merge, still hashes to the digest), plus the three mlx-vlm attributes the corrections call — two of which the probe did not import at all before.
- `scripts/setup-image-video.sh` passes the flag and now prints the probe's message instead of discarding stderr. That message is the only thing distinguishing a moved pin from a broken package sync.
- The seam table also collapses the four call sites: each installer now says `pinned_encoder_hook("encode")` and reads its consequence/remedy/shape from the table, so the install-time check cannot drift from the render-time one.

The render-time guards stay as the last line for anyone overriding the pin.

### The flag is deliberately Install/Repair-only

The same script is the readiness probe behind `isByovRuntimeReady`. A failure there sets `byovGateBlocked` and disables Generate outright — but a moved seam breaks only the keyframe and substituted-conditioner paths, so failing readiness over it would block text-only renders that still work, and offer Repair as the remedy, which re-runs the same probe. That is precisely the trap `minimax_h3_lora_probe.py` is kept a separate probe to avoid:

> Kept separate from minimax_h3_runtime_probe.py so a runtime that is perfectly healthy for plain renders is never marked unready just because it predates the LoRA applicator.

No coverage is lost: a pin bump reaches Install/Repair anyway, because a moved revision already fails `isByovRuntimeCurrent`, and a re-locked mlx-vlm needs a re-sync.

### Two drift guards for gaps this exposed

- **The shipped digest must move with the pinned revision.** Both probe suites overwrite `PINNED_ENCODE_DIGEST` with a freshly computed one, so nothing asserted the *shipped* value — a revision bump with a stale digest would have passed every test and shipped green. Same co-movement shape as `PROMPT_VERSIONS` / `PREVIOUS_DEFAULT_PROMPTS`.
- **Every mlx-vlm symbol the corrections import must be in the probed set**, so a correction reaching for a fourth one can't sail past Install/Repair and fail mid-render.

Closes #4606. Filed #4619 for the related-but-separate sweep: 14 other install probes in `setup-image-video.sh` still discard their traceback.

## Test plan

- `scripts/minimax_h3_runtime_probe.test.js` — new, 18 cases. A checkout with every seam intact passes; each seam individually dropped (and `processor` demoted from a property) fails Install/Repair with its own message; an `encode` body that moved under an intact hook, and one where upstream stopped merging by broadcast, fail with the two opposite remedies; each mlx-vlm module/attribute dropped fails by name. One case asserts the **flagless** probe still exits 0 with a moved seam, which is the readiness-gate regression this guards.
- The fixture builds a real on-disk package (the encode seam is guarded with `inspect.getsource`, so a `sys.modules`-only fake won't do) and splices the merge line in from the shipped constant, so a re-recorded pin can't leave the suite asserting the old one.
- Guards verified non-vacuous by deliberately breaking each: neutering `verify_pinned_seams()` → 12 failures; bumping the revision without re-recording the digest → co-movement test fails; adding an unprobed `from mlx_vlm.newthing import Widget` to the runner → import-drift test fails.
- `scripts/generate_minimax_h3.test.js` (45 cases) retargeted at the constants' new home and still green.
- Full server suite: 1489 files, 30,794 tests passing. `bash -n scripts/setup-image-video.sh` clean.

### DIFF
--- a/scripts/_minimax_h3_pin.py
+++ b/scripts/_minimax_h3_pin.py
@@ -1,0 +1,169 @@
+"""The pinned-MiniMax-H3 seams PortOS patches, stated once for both callers.
+
+`generate_minimax_h3.py` corrects four things PipeNetwork's pinned MLX port
+gets wrong on the keyframe path, and each correction wraps a specific method of
+`MiniMaxH3TextEncoder` — so each is only valid against the exact implementation
+it was written against. Which methods those are, and what "unchanged" means for
+the one whose body is copied rather than wrapped, are facts about the PIN, not
+about the runner, and two callers need them:
+
+  - the runner, which refuses to install a correction onto a hook that moved
+  - `minimax_h3_runtime_probe.py --verify-seams`, which Install / Repair runs and
+    which therefore fails at the action that moved the pin, rather than minutes
+    into the first keyframe render that happens to reach the seam
+
+`PINNED_ENCODE_DIGEST` in particular must have exactly ONE home. A second copy
+is a copy that goes stale, and a stale digest asserts a pin bump is fine when
+it is not — which is the failure this module exists to make impossible.
+
+Stdlib-only at import, like `_runner_common.py` and `_minimax_h3_common.py`:
+the pinned package and mlx-vlm are imported inside the functions that need
+them, so nothing here forces a venv to grow a dependency, and the probe can
+import this module before deciding the checkout is usable at all.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import importlib
+import inspect
+
+# The merge the pinned port performs instead of a scatter. Asserted before the
+# replacement is installed, so the day upstream fixes this the patch says so
+# loudly rather than shadowing a working implementation forever.
+PINNED_BROADCAST_MERGE = "mx.where(image_mask[..., None], hidden.astype(inputs_embeds.dtype)[None], inputs_embeds)"
+
+# sha256 of the whole pinned `encode`, because the replacement re-runs its body
+# with one line changed rather than wrapping it. Matching only the merge line
+# would let every OTHER edit a pin bump makes — a new `_hidden_states` argument,
+# a different dtype, an added step — pass silently into a stale copy.
+# Re-record this when bumping MINIMAX_H3_EXPECTED_REVISION, after re-reading
+# `encode` and folding whatever changed into the replacement.
+PINNED_ENCODE_DIGEST = "8047e407e797cd46cd7538024ca09d97402d369d97b1d825757a1590416bda7d"
+
+# Every method of the pinned encoder a PortOS correction wraps: what breaks when
+# the pin no longer has it, the shape the correction requires it to still be, and
+# the way out that rescues the render. A TABLE rather than arguments at the four
+# call sites, because the runner patches these and the probe asserts them —
+# stating the pair apart is exactly how the install-time check drifts from the
+# render-time one, which is the failure this module exists to prevent.
+#
+# `remedy` defaults to dropping the keyframe, which is what rescues three of
+# them. The key-prefix map is the exception: it is reached by a substituted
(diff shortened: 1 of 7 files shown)


atomantic:
Closing as a duplicate of #4610.

Both PRs implement issue #4606 — hoisting the MiniMax H3 pinned-seam constants into a shared module and asserting them in `minimax_h3_runtime_probe.py` at Install/Repair. They create the same shared module under different names (`_minimax_h3_pin.py` here, `_minimax_h3_mlx_pins.py` there) and edit the same four files, so they cannot both land.

#4610 is the one that carries the issue claim (assignee + `in-progress` on #4606), was opened first, and its probe assertions were validated against the real pinned MLX checkout — the shipped digest matches the live `encode`, and each assertion was confirmed to fire when its attribute was removed at runtime.

The extra scope that is unique to this PR — surfacing the probe's stderr instead of discarding it in `runVenvProbe()` and `setup-image-video.sh` — is not lost: it is filed as #4618, scoped generically across all BYOV runtime probes rather than just MiniMax H3.

PR #4637, opened 2026-08-20
harden BYOV probe logging: a normal LoRA verdict is not a failure

atomantic:
Follow-up to #4622, which landed the core of #4618 (piping the probe's stderr). This fixes four cases its logging still gets wrong.

## What's wrong today

- **A normal verdict is reported as a failure.** `runVenvProbe` serves two callers with opposite meanings for a non-zero exit. For `isByovRuntimeReady` it's a fault; for the LoRA-capability probe it's the *documented normal answer* — `scripts/minimax_h3_lora_probe.py` says any non-zero exit means "this checkout has no applicator", and the result is cached as a legitimate `false`. Every healthy H3 install predating the applicator currently logs `❌ BYOV runtime probe failed: Traceback… ModuleNotFoundError`.
- **A failure with empty stderr logs nothing**, which reads identically to "never probed" — the exact sentinel-vs-empty case `CLAUDE.md` calls out.
- **The spawn-error and timeout paths are still silent**, so "the venv python is missing" and "the import wedged for 30s" stay as opaque as before the fix.
- **Lines are unattributable** — a status request fans out over several runtimes and every line reads the same.

## Changes

- Split the icons: `⚠️` when the probe answered, `❌` when it never did.
- Empty capture logs `(no stderr output)` rather than going quiet.
- Log on spawn error and on timeout.
- Label each line with the runtime and which probe ran. The spawn error is reduced to `err.code` — Node's message interpolates the interpreter path, which embeds the OS username.
- Decode stderr on the stream, so a multi-byte character split across chunks survives instead of becoming U+FFFD.

## Test plan

- 4 new tests in `services/videoGen/runtimes.test.js`, **each verified failing against #4622 as merged** and passing here.
- #4622's own test is kept; only its cap assertion changed, since it pinned the old fixed prefix and the line is now labelled per runtime.
- Full server suite: 31002 passed / 19 skipped. Client lint clean.

Refs #4618

### DIFF
--- a/server/services/videoGen/runtimes.js
+++ b/server/services/videoGen/runtimes.js
@@ -316,32 +316,81 @@ export async function isByovRuntimeReady(runtimeId) {
   // earlier successful probe.
   if (info.expectedRevision && !await isByovRuntimeCurrent(runtimeId)) return false;
   if (readyCache.get(runtimeId) === true) return true;
-  const probeOk = await runVenvProbe(info.venvPython, info.probeArgs || ['-c', info.importProbe]);
+  const probeOk = await runVenvProbe(
+    info.venvPython, info.probeArgs || ['-c', info.importProbe], `${runtimeId} readiness`,
+  );
   if (probeOk) readyCache.set(runtimeId, true);
   return probeOk;
 }
 
+// How much probe stderr to retain. The probes emit single diagnostic lines, so
+// a few KB is generous; an unbounded buffer on a child that wedges mid-spew is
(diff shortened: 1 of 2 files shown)


atomantic:
Closing in favor of #4622, which covers the same helper and is the one being reviewed. Superseded, not abandoned — the branch stays on origin if any of the extras here are worth folding into #4622.

atomantic:
Reopening for a side-by-side comparison with #4622 — not a claim on which should land.

PR #4871, opened 2026-08-23
fix: centralize direct-invocation guards

### LINKED ISSUE #4868: Adopt isDirectlyInvoked in the 9 scripts still hand-rolling the is-main check
`scripts/lib/directInvocation.js` (added in #4866) owns the "was this module run directly, or imported?" check, with the two pieces of care a naive comparison misses: **realpath resolution** and **case folding** on NTFS/APFS. Two call sites adopted it (`checkNodeVersion.js`, `checkNpmVersion.js`); nine still hand-roll a weaker variant.

## Why this matters

Both hand-rolled forms compare `resolve()`d paths without resolving symlinks, and never case-fold. So under a checkout reached through a **symlinked parent** — a repo under a symlinked home, `/tmp` → `/private/tmp` on macOS, a symlinked worktree — `import.meta.url` (which Node's ESM loader resolves) and `process.argv[1]` (whatever the caller typed) disagree, the guard is false, and **the script silently does nothing and exits 0**.

The worst case is `scripts/run-migrations.js`: `npm run migrations` reports success having applied no migrations. That is a silent data-shape failure, not a cosmetic one. `server/scripts/run-db-migrations.js` has the same shape.

## Sites

`pathToFileURL` form:

- `scripts/run-migrations.js:181`
- `server/scripts/run-db-migrations.js:114`

`resolve(...) === resolve(fileURLToPath(...))` form:

- `scripts/ci-base-sha.js:77`
- `scripts/ci-test-plan.js:470`
- `scripts/run-ci-lint.js:73`
- `scripts/run-ci-tests.js:197`
- `scripts/trusted-rebuild-stamp.js:104`
- `scripts/trusted-rebuilds.js:150`
- `scripts/verify-ci-status.js:141`

## What to do

Replace each with `isDirectlyInvoked(import.meta.url)` from `scripts/lib/directInvocation.js` (`../lib/` from `server/scripts/`), dropping the now-unused `resolve` / `fileURLToPath` / `pathToFileURL` imports where nothing else uses them. Several sites `&&` the check with another condition (e.g. an env guard) — keep that half as-is.

**One decision is already made, so this is ready to work:** if `run-migrations.js` turns out to have a real reason to avoid the helper's eval-time `realpathSync` I/O, keep its current form but add a one-line comment saying so — right now the divergence reads as an oversight rather than a choice. Do not leave it silently different.

Add a guard test that fails when a new `scripts/*.js` hand-rolls the comparison instead of importing the helper, so the tenth copy can't appear — the same shape as the existing drift tests in `scripts/node-version-drift.test.js`.

Found during review of #4866.

Freakz2z:
## Summary
- replace the nine hand-rolled entrypoint comparisons from #4868 with the shared `isDirectlyInvoked` helper
- migrate the same weaker guard in `generate-prompt-stage-call-sites.js`, found while adding the drift check
- add a drift test that rejects future direct comparisons in top-level JavaScript scripts

## Test plan
- `cd server && ./node_modules/.bin/vitest run ../scripts/direct-invocation-drift.test.js ../scripts/ci-base-sha.test.js ../scripts/ci-test-plan.test.js ../scripts/generate-prompt-stage-call-sites.test.js ../scripts/run-ci-lint.test.js ../scripts/run-ci-tests.test.js ../scripts/run-migrations.test.js ../scripts/trusted-rebuild-stamp.test.js ../scripts/trusted-rebuilds.test.js ../scripts/verify-ci-status.test.js` (142 passed)
- invoked `scripts/ci-base-sha.js` through a symlinked checkout path; the CLI body ran instead of silently exiting
- full server suite: 33,450 passed, 9 skipped, 1 environment-specific failure in the unchanged `lib/connectivity.test.js` because `192.0.2.1` is reachable from this machine

Closes #4868

### DIFF
--- a/scripts/direct-invocation-drift.test.js
+++ b/scripts/direct-invocation-drift.test.js
@@ -1,0 +1,38 @@
+/**
+ * Drift guard for direct-invocation checks (issue #4868).
+ *
+ * A hand-rolled comparison between process.argv[1] and import.meta.url misses
+ * symlink and case-folding behavior handled by scripts/lib/directInvocation.js.
+ * Keep every import-safe CLI on the shared helper so a direct run cannot
+ * silently no-op under a symlinked checkout.
+ */
+import { describe, expect, it } from 'vitest';
+import { readdirSync, readFileSync } from 'fs';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+
+const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
+const SCRIPT_DIRS = ['scripts', 'server/scripts'];
+
+const scriptFiles = SCRIPT_DIRS.flatMap((relativeDir) => {
+  const directory = join(REPO_ROOT, relativeDir);
+  return readdirSync(directory, { withFileTypes: true })
+    .filter((entry) => (
+      entry.isFile()
+      && entry.name.endsWith('.js')
+      && !entry.name.endsWith('.test.js')
+    ))
+    .map((entry) => `${relativeDir}/${entry.name}`);
+});
+
+describe('direct-invocation checks have one owner (issue #4868)', () => {
+  it('top-level JavaScript scripts do not compare argv[1] with import.meta.url themselves', () => {
+    const offenders = scriptFiles.filter((relativePath) => {
+      const source = readFileSync(join(REPO_ROOT, relativePath), 'utf8');
+      return /process\.argv\[1\][\s\S]{0,200}import\.meta\.url/.test(source)
+        || /import\.meta\.url[\s\S]{0,200}process\.argv\[1\]/.test(source);
+    });
+
+    expect(offenders).toEqual([]);
+  });
+});
--- a/scripts/ci-base-sha.js
+++ b/scripts/ci-base-sha.js
@@ -26,8 +26,7 @@
  */
 
 import { spawnSync } from 'child_process';
-import { resolve } from 'path';
-import { fileURLToPath } from 'url';
+import { isDirectlyInvoked } from './lib/directInvocation.js';
 import { writeStepEnv } from './lib/githubOutput.js';
 
 /** Resolve a revision to a commit sha, or null when git cannot. */
@@ -73,7 +72,4 @@ function main() {
   console.log('No pull-request merge base to resolve — this run tests the complete suite.');
 }
 
-const isMain = process.argv[1]
-  && resolve(process.argv[1]) === resolve(fileURLToPath(import.meta.url));
-
-if (isMain) main();
+if (isDirectlyInvoked(import.meta.url)) main();
--- a/scripts/ci-test-plan.js
+++ b/scripts/ci-test-plan.js
@@ -1,8 +1,7 @@
 #!/usr/bin/env node
 
 import { execFileSync } from 'child_process';
-import { fileURLToPath } from 'url';
-import { resolve } from 'path';
+import { isDirectlyInvoked } from './lib/directInvocation.js';
 import { writeStepOutput } from './lib/githubOutput.js';
 
 const TEST_FILE_RE = /\.(?:test|spec)\.[cm]?[jt]sx?$/i;
@@ -466,7 +465,4 @@ function main() {
   emitGitHubPlan(buildCiTestPlan(changedFiles, { trackedFiles, forceFull, forceFullReason }));
 }
 
-const isMain = process.argv[1]
-  && resolve(process.argv[1]) === resolve(fileURLToPath(import.meta.url));
-
-if (isMain) main();
+if (isDirectlyInvoked(import.meta.url)) main();
--- a/scripts/generate-prompt-stage-call-sites.js
+++ b/scripts/generate-prompt-stage-call-sites.js
@@ -28,6 +28,7 @@ import { execFileSync } from 'node:child_process';
 import { readFileSync, writeFileSync } from 'node:fs';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
+import { isDirectlyInvoked } from './lib/directInvocation.js';
 
 const HERE = dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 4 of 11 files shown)


scripts/direct-invocation-drift.test.js
```diff
@@ -0,0 +1,38 @@
...
+  it('top-level JavaScript scripts do not compare argv[1] with import.meta.url themselves', () => {
+    const offenders = scriptFiles.filter((relativePath) => {
+      const source = readFileSync(join(REPO_ROOT, relativePath), 'utf8');
+      return /process\.argv\[1\][\s\S]{0,200}import\.meta\.url/.test(source)
+        || /import\.meta\.url[\s\S]{0,200}process\.argv\[1\]/.test(source);
+    });
+
+    expect(offenders).toEqual([]);
+  });
+});
```
atomantic, line 38: **[IMPROVEMENT]** The guard can go vacuous without anyone noticing.

`offenders` is only ever non-empty while both regexes still match. If a future edit breaks the detector (or the file list quietly comes back empty), this test passes forever and the tenth hand-rolled copy lands unchallenged — the exact failure mode the test exists to prevent.

The sibling guard in this repo already sets the precedent: `scripts/node-version-drift.test.js` verifies `pinSatisfiesFloor` against literals first, with the note *"The pin comparison decides whether the .nvmrc / CI assertions below pass, so it is verified against literals rather than trusted."* Same reasoning applies here.

Suggested: lift the predicate out, probe it against synthetic offenders and a synthetic clean site, and assert the file list is non-empty. Verified passing (2 tests) against this branch.

```suggestion
/** True when `source` hand-rolls the argv[1] <-> import.meta.url comparison. */
const handRollsCheck = (source) => (
  /process\.argv\[1\][\s\S]{0,200}import\.meta\.url/.test(source)
  || /import\.meta\.url[\s\S]{0,200}process\.argv\[1\]/.test(source)
);

describe('direct-invocation checks have one owner (issue #4868)', () => {
  // The scan below can only fail while the detector still fires, so it is
  // verified against synthetic sources rather than trusted.
  it('handRollsCheck flags hand-rolled guards and clears the shared helper', () => {
    expect(handRollsCheck('const isMain = process.argv[1]\n  && resolve(process.argv[1]) === resolve(fileURLToPath(import.meta.url));')).toBe(true);
    expect(handRollsCheck('if (import.meta.url === pathToFileURL(process.argv[1]).href) main();')).toBe(true);
    expect(handRollsCheck('if (isDirectlyInvoked(import.meta.url)) main();')).toBe(false);
  });

  it('top-level JavaScript scripts do not compare argv[1] with import.meta.url themselves', () => {
    expect(scriptFiles.length).toBeGreaterThan(0);
    expect(scriptFiles.filter((relativePath) => (
      handRollsCheck(readFileSync(join(REPO_ROOT, relativePath), 'utf8'))
    ))).toEqual([]);
  });
});
```

atomantic (commented):
Reviewed by `/do:review` — **0 critical, 1 improvement.** The change is correct, complete, and does what it says.

### Verified empirically

Checked out the PR head and reproduced the real failure mode rather than reasoning about it:

- **Baseline reproduces the bug.** `node <symlinked-checkout>/scripts/ci-base-sha.js` on `main` printed nothing and exited 0 — the silent no-op #4868 describes.
- **This branch fixes it.** The same invocation on `a3da059` printed `No pull-request merge base to resolve...` and exited 0. The CLI body actually ran.
- **The drift guard genuinely fires.** Appended a hand-rolled `resolve(process.argv[1]) === resolve(fileURLToPath(import.meta.url))` to `scripts/ci-base-sha.js`; `direct-invocation-drift.test.js` failed as intended. Reverted.
- **Tests pass.** 10 files / 142 tests green — matching the count in the PR body exactly.
- **No dead imports left behind.** Every retained `pathToFileURL` / `resolve` / `fileURLToPath` still has a live use (`run-migrations.js:139,153`, `run-db-migrations.js:88`, `generate-prompt-stage-call-sites.js:36`).

### Coherence check

Every claim in the PR description checks out: the nine sites from #4868 plus the `generate-prompt-stage-call-sites.js` site found in passing, the drift test, the symlink probe, and the 142-test count. The `connectivity.test.js` note accurately describes a local-environment artifact, not a regression.

On the one decision #4868 flagged: migrating `run-migrations.js` to the helper is the right call. The dropped comment justified avoiding *eval-time filesystem I/O*, and `isDirectlyInvoked` does add two `realpathSync` calls at import — but it stays synchronous, so the load-bearing half of that rationale (not turning the module async) is preserved, and two `realpathSync` calls are not worth keeping a silently-broken guard for.

### Out-of-diff observations

Neither blocks this PR — both are follow-up material.

1. **`server/services/pipeline/migrateSeriesCanon.js:226`** still carries the weakest variant of all: ``import.meta.url === `file://${process.argv[1]}` ``. On top of the symlink and case-folding gaps this PR closes, that form also breaks on a relative `argv[1]` and on any path needing percent-encoding (spaces, non-ASCII). It sits outside `SCRIPT_DIRS`, so the new guard will not catch it.
2. **`SCRIPT_DIRS` scans top level only.** `readdirSync` without `recursive: true`, plus the `entry.isFile()` filter, means `scripts/lib/*.js` and any future `scripts/<subdir>/*.js` are unguarded. Widening the scan (or adding `server/services/` to it) would cover both this and item 1 — reasonable either here or as a follow-up.

Nice catch on `generate-prompt-stage-call-sites.js` — that one was not in the issue and had the weaker single-sided `resolve()`.

_Generated by /do:review_

scripts/direct-invocation-drift.test.js
atomantic: Picked this up in #4877 rather than holding the merge — no action needed here.

That PR adds the probe as suggested, and while in there also widened the scan from the two directory tops to every tracked `.js` file (~65ms for 2507 files, so the narrower scope was buying nothing). Widening it surfaced one real offender the directory-scoped version could never have seen: `server/services/pipeline/migrateSeriesCanon.js`, carrying the weakest spelling of all — ``import.meta.url === `file://${process.argv[1]}` `` — on a documented data-recovery path. Fixed there too.

Thanks for this one — the centralization was clean and the guard test was the right instinct.

PR #5009, opened 2026-08-24
feat: delegate release checks to configured reviewers

atomantic:
## Summary

- Delegate scheduled release-check mechanics to the bundled slashdo `release` workflow.
- Remove the hardcoded Copilot review polling and pin the PortOS-configured reviewer set.
- Add prompt v11 compatibility and an update migration for uncustomized schedules.
- Lock release-check metadata to the bundled workflow so app overrides cannot reintroduce a parallel release path.

## Test plan

- `server/node_modules/.bin/vitest run --root server --config vitest.config.js services/taskPromptDefaults.test.js services/taskSchedule.test.js services/agentPromptBuilder.test.js services/cosTaskGenerator.test.js services/codeReview.test.js lib/slashdoInvocation.test.js ../scripts/migrations/297-release-check-slashdo-review-prompt.test.js`
- `node --check` on all changed JavaScript files
- `git diff --check`


atomantic:
## Review blocker

The required `codex` reviewer was invoked in read-only mode with the pinned `gpt-5.6-luna` model and `model_reasoning_effort=max` against `main`, but it did not emit a verdict within the documented 30-minute bound and was stopped. The worktree remained unchanged during the review. Per the review contract this pass is `cli-error`/inconclusive, so this PR remains open and unmerged; no Copilot fallback was used.

PR #5056, opened 2026-08-25
Centralize the streaming-wrapper error envelope into apiCore.throwApiError

### LINKED ISSUE #5047: Extract the duplicated fetch error-envelope in the streaming API wrappers into apiCore
## Problem

Six client service wrappers bypass `request()` (they need the raw `Response` to read a stream or a blob) and each hand-rolls the same `!response.ok` block that `apiCore.js:47-70` already owns:

- `client/src/services/apiCore.js:47` — the original
- `client/src/services/apiMedia.js:321`
- `client/src/services/apiMedia.js:360`
- `client/src/services/apiImageVideo.js:495`
- `client/src/services/apiLocalLlm.js:188`
- `client/src/services/apiSystem.js:172` (`downloadBackupSnapshot`)

The block is: json-parse-with-fallback → `maybeRedirectToLogin(response, error)` → `new Error(error.error || \`HTTP ${status}\`)` → attach `.code` and `.status` → throw.

They have already drifted. `apiCore.js:69` forwards the server's structured `error.context` onto the thrown error; **none of the five copies do**, so a streaming caller silently loses it. The copies also skip `request()`'s `fetch(...).catch(() => null)` guard, so a dropped connection rejects with a raw `TypeError: Failed to fetch` instead of the app's "Server unreachable — check your connection and try again".

## Approach

Export a helper from `apiCore.js` and route all six through it:

```js
export async function throwApiError(response) {
  const error = await response.json().catch(() => ({ error: `HTTP ${response.status}` }));
  maybeRedirectToLogin(response, error);
  const err = new Error(error.error || `HTTP ${response.status}`);
  err.code = error?.code;
  err.status = response.status;
  if (error?.context) err.context = error.context;
  throw err;
}
```

Then each call site becomes `if (!response.ok) await throwApiError(response);`. Refactor `request()` itself to use it so the definition has exactly one home (note `request()` additionally toasts unless `silent` — keep the toasting in `request()`, not in the helper, since the streaming callers deliberately surface their own errors).

Per the module maintenance rule, `apiCore.js` is already exported via `client/src/services/api.js`; add a row for the new export in `client/src/services/README.md` if that file enumerates exports rather than modules.

## Acceptance criteria

- [ ] `throwApiError` exists in `apiCore.js` and is the only place the error envelope is constructed.
- [ ] All six sites listed above call it; no `err.code = ` / `err.status = ` assignment remains in the service wrappers.
- [ ] `error.context` now survives on errors thrown from the five streaming/blob wrappers (add a test for at least one).
- [ ] Existing client suites stay green — several mock these modules, so check the mock seams.

## Non-goals

- Not converting the streaming wrappers to use `request()` — they need the unconsumed `Response`. A `responseType: 'response'` mode in `request()` is a plausible follow-up but is a larger behavioral change and is out of scope here.

## Context

Surfaced by a `/simplify` reuse+altitude pass on #5034 (snapshot download), which added the sixth copy. Deferred from that PR to keep the feature diff scoped.

Bryandero98:
Fixes #5047.

## What changed

`apiCore.js` gains `throwApiError(response)`: parses the server's JSON error body, calls `maybeRedirectToLogin` for session expiry, and throws an `Error` carrying `.code`/`.status`/`.context`. `request()` now delegates to it for the envelope construction itself, keeping its own toast/`silent` handling wrapped around the call so behavior there is unchanged.

## Update: the sixth site (apiSystem.js) is now covered too

An earlier revision of this description said `apiSystem.js`'s `downloadBackupSnapshot` "no longer exists in the codebase" - that was true against `main` at the time, but PR #5050 landed after this branch was opened and reintroduced it with its own local `!response.ok` envelope. Per review feedback, I rebased onto current `main` and routed `downloadBackupSnapshot` through `throwApiError` as well (keeping the pre-existing file-handle abort-on-failure cleanup, which now runs before the delegated throw).

All sites are now centralized:

- `apiMedia.js` - `cleanImage`, `fetchCleanResult`
- `apiImageVideo.js` - `installLoraFromHuggingfaceStream`
- `apiLocalLlm.js` - `streamLocalLlmTest`
- `apiSystem.js` - `downloadBackupSnapshot`

## A real bug fix, not just dedup

`apiImageVideo.js` and `apiLocalLlm.js`'s copies never forwarded the server's structured `error.context` - exactly the problem the issue opens with. All four now do, for free, since they go through the same helper `apiMedia.js` and `request()` already used.

## Testing

- `apiCore.test.js`: `throwApiError` directly - message/code/status from the body, the `HTTP {status}` fallback when the body isn't JSON *and* when it's valid JSON but not an object (e.g. a bare `null`), `error.context` surviving, plus `request()`'s toast/`silent`/`PLATFORM_UNAVAILABLE` behavior (including the warning toast itself being suppressed under `{ silent: true }`, not just `toast.error`).
- `apiMedia.test.js`'s `./apiCore.js` mock now mirrors production's null-safe body normalization instead of diverging from it, and asserts `error.context` survives through the direct-fetch wrapper and that session expiry (`maybeRedirectToLogin`) is honored the same way `request()` handles it.
- `apiSystem.download.test.js`: added a case asserting `error.context` survives through `downloadBackupSnapshot`'s direct-fetch path the same way.
- **44/44 passing** across `apiCore.test.js`, `apiMedia.test.js`, `apiSystem.test.js`, `apiSystem.download.test.js`; `biome check` clean on all touched files.
- Ran the full `client` suite twice (770 files / 9850 tests, ~11 min each) prior to the rebase. Both runs had a handful of pre-existing failures unrelated to this change (failing set differed between runs, consistent with pre-existing flake), and none of those files import `apiCore`/`apiMedia`/`apiImageVideo`/`apiLocalLlm`/`apiSystem`.

### DIFF
--- a/client/src/services/apiCore.js
+++ b/client/src/services/apiCore.js
@@ -27,7 +27,12 @@ export function maybeRedirectToLogin(response, error) {
 // own (with `silent` support); the streaming/blob callers that bypass request()
 // to read a raw Response deliberately surface their own errors instead.
 export async function throwApiError(response) {
-  const error = await response.json().catch(() => ({ error: `HTTP ${response.status}` }));
+  // A valid JSON body that isn't an object (e.g. a bare `null`) parses
+  // successfully but has no `.error`/`.code`/`.context` to read — fall back
+  // to the same HTTP-status shape used when the body isn't JSON at all.
+  const parsedError = await response.json().catch(() => null);
+  const error =
+    parsedError && typeof parsedError === 'object' ? parsedError : { error: `HTTP ${response.status}` };
   // Auth gate (server: services/authGate.js) returns 401 with code AUTH_REQUIRED
   // for any /api request without a valid session. Bounce to /login so the
   // user can re-authenticate; skip if we're already there.
--- a/client/src/services/apiCore.js
+++ b/client/src/services/apiCore.js
@@ -21,6 +21,28 @@ export function maybeRedirectToLogin(response, error) {
   }
 }
 
+// Shared `!response.ok` envelope: parse the server's JSON error body, bounce to
+// /login on session expiry, and throw an Error carrying `.code`/`.status`/
+// `.context` from the response. Toasting is NOT done here — request() does its
+// own (with `silent` support); the streaming/blob callers that bypass request()
+// to read a raw Response deliberately surface their own errors instead.
+export async function throwApiError(response) {
+  const error = await response.json().catch(() => ({ error: `HTTP ${response.status}` }));
+  // Auth gate (server: services/authGate.js) returns 401 with code AUTH_REQUIRED
+  // for any /api request without a valid session. Bounce to /login so the
+  // user can re-authenticate; skip if we're already there.
+  maybeRedirectToLogin(response, error);
+  const err = new Error(error.error || `HTTP ${response.status}`);
+  err.code = error?.code;
+  err.status = response.status;
+  // Forward structured context the server attached to the error (e.g.
+  // ERR_PARTIAL_COMMIT_ISSUES carries `{ universeId, seriesId,
+  // arcAlreadyPersisted, skipArcOnRetry }` so the Importer client can
+  // shape its retry without re-overwriting persisted state).
+  if (error?.context) err.context = error.context;
+  throw err;
+}
+
 export async function request(endpoint, options = {}) {
   const { silent, responseType, ...fetchOptions } = options;
   const url = `${API_BASE}${endpoint}`;
@@ -45,29 +67,19 @@ export async function request(endpoint, options = {}) {
   }
 
   if (!response.ok) {
-    const error = await response.json().catch(() => ({ error: 'Request failed' }));
-    const errorMessage = error.error || `HTTP ${response.status}`;
-    // Auth gate (server: services/authGate.js) returns 401 with code AUTH_REQUIRED
-    // for any /api request without a valid session. Bounce to /login so the
-    // user can re-authenticate; skip if we're already there.
-    maybeRedirectToLogin(response, error);
-    if (!silent) {
-      // Platform unavailability is a warning, not an error
-      if (error.code === 'PLATFORM_UNAVAILABLE') {
-        toast(errorMessage, { icon: '⚠️' });
-      } else if (error.code !== 'AUTH_REQUIRED') {
-        toast.error(errorMessage);
+    try {
+      await throwApiError(response);
+    } catch (err) {
+      if (!silent) {
+        // Platform unavailability is a warning, not an error
+        if (err.code === 'PLATFORM_UNAVAILABLE') {
+          toast(err.message, { icon: '⚠️' });
+        } else if (err.code !== 'AUTH_REQUIRED') {
+          toast.error(err.message);
+        }
       }
+      throw err;
     }
-    const err = new Error(errorMessage);
-    err.code = error?.code;
-    err.status = response.status;
-    // Forward structured context the server attached to the error (e.g.
-    // ERR_PARTIAL_COMMIT_ISSUES carries `{ universeId, seriesId,
-    // arcAlreadyPersisted, skipArcOnRetry }` so the Importer client can
-    // shape its retry without re-overwriting persisted state).
-    if (error?.context) err.context = error.context;
-    throw err;
   }
 
   // Handle 204 No Content
--- a/client/src/services/apiCore.test.js
+++ b/client/src/services/apiCore.test.js
@@ -50,17 +50,22 @@ describe('throwApiError', () => {
 
   it('omits context when the server did not send any', async () => {
     const response = makeResponse({ status: 500, json: { error: 'boom' } });
-    try {
-      await throwApiError(response);
-      throw new Error('expected throwApiError to throw');
-    } catch (err) {
-      expect(err.context).toBeUndefined();
-    }
+    const err = await throwApiError(response).catch((e) => e);
+    expect(err).toBeInstanceOf(Error);
+    expect(err.context).toBeUndefined();
   });
 
   it('falls back to an HTTP-status message when the body is not JSON', async () => {
     const response = makeResponse({ status: 502 }); // json:null → makeResponse's json() rejects
     await expect(throwApiError(response)).rejects.toMatchObject({ message: 'HTTP 502' });
+  });
+
+  it('falls back to an HTTP-status message when the body is valid JSON but not an object', async () => {
+    // response.json() resolves successfully with `null` here — a real case
+    // (an endpoint that responds 500 with a bare `null` body) distinct from
+    // "body isn't JSON at all" above, which rejects instead of resolving.
+    const response = { status: 500, ok: false, json: async () => null };
+    await expect(throwApiError(response)).rejects.toMatchObject({ message: 'HTTP 500' });
   });
 });
 
@@ -74,11 +79,12 @@ describe('request() error path (now delegating to throwApiError)', () => {
     expect(toast.error).toHaveBeenCalledWith('bad input');
   });
 
-  it('does not toast for PLATFORM_UNAVAILABLE errors (warns instead)', async () => {
+  it('warns instead of toasting an error for PLATFORM_UNAVAILABLE', async () => {
     global.fetch.mockResolvedValue(
       makeResponse({ status: 503, json: { error: 'offline', code: 'PLATFORM_UNAVAILABLE' } }),
     );
     await expect(request('/x')).rejects.toMatchObject({ code: 'PLATFORM_UNAVAILABLE' });
+    expect(toast).toHaveBeenCalledWith('offline', { icon: '⚠️' });
     expect(toast.error).not.toHaveBeenCalled();
   });
 
--- a/client/src/services/apiCore.test.js
+++ b/client/src/services/apiCore.test.js
@@ -1,0 +1,90 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// The real Toast default export is callable (`toast(msg, opts)`) AND carries
+// `.error`/`.success`/etc, mirroring react-hot-toast's API — a plain object
+// mock would make request()'s `toast(err.message, { icon })` call throw.
+vi.mock('../components/ui/Toast', () => {
+  const toastFn = vi.fn();
+  toastFn.error = vi.fn();
+  toastFn.success = vi.fn();
+  return { default: toastFn };
+});
+
+import toast from '../components/ui/Toast';
+import { throwApiError, request } from './apiCore.js';
+
+const makeResponse = ({ status = 400, ok = false, json = null } = {}) => ({
+  status,
+  ok,
+  json: json === null ? async () => { throw new Error('not json'); } : async () => json,
+});
+
+beforeEach(() => {
+  global.fetch = vi.fn();
+  toast.mockReset();
+  toast.error.mockReset();
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe('throwApiError', () => {
+  it('throws with the server message, code, and status', async () => {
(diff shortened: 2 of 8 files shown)


Bryandero98:
Hi! Extracted and centralized the duplicated fetch error-envelope into apiCore's new throwApiError. Cleaned up the 4 active duplication sites across the streaming API wrappers (noting that the backup snapshot wrapper had already been factored out previously). All 16 tests passing, Biome lint clean. Let me know what you think!

client/src/services/apiCore.js
```diff
@@ -21,6 +21,28 @@ export function maybeRedirectToLogin(response, error) {
   }
 }
 
+// Shared `!response.ok` envelope: parse the server's JSON error body, bounce to
+// /login on session expiry, and throw an Error carrying `.code`/`.status`/
+// `.context` from the response. Toasting is NOT done here — request() does its
+// own (with `silent` support); the streaming/blob callers that bypass request()
+// to read a raw Response deliberately surface their own errors instead.
+export async function throwApiError(response) {
+  const error = await response.json().catch(() => ({ error: `HTTP ${response.status}` }));
```
atomantic: **[IMPROVEMENT]** A valid JSON null error body makes response.json() resolve successfully, then error.error below throws a TypeError. The streaming/blob callers now all depend on this helper, so they lose the intended HTTP-status error and status metadata for that response shape.

```suggestion
  const parsedError = await response.json().catch(() => null);
  const error = parsedError && typeof parsedError === 'object'
    ? parsedError
    : { error: 'HTTP ' + response.status };
```

client/src/services/apiCore.test.js
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  it('omits context when the server did not send any', async () => {
+    const response = makeResponse({ status: 500, json: { error: 'boom' } });
+    try {
+      await throwApiError(response);
+      throw new Error('expected throwApiError to throw');
+    } catch (err) {
+      expect(err.context).toBeUndefined();
+    }
```
atomantic: **[IMPROVEMENT]** This try/catch catches the test’s own sentinel error. If throwApiError() accidentally resolves, line 55 throws and line 57 still passes because that sentinel has no context, so this test does not require the rejection contract it claims to test.

```suggestion
    const err = await throwApiError(response).catch((error) => error);
    expect(err).toBeInstanceOf(Error);
    expect(err.context).toBeUndefined();
```

client/src/services/apiCore.test.js
```diff
@@ -0,0 +1,90 @@
...
+    await expect(request('/x')).rejects.toMatchObject({ code: 'VALIDATION', status: 422, context });
+    expect(toast.error).toHaveBeenCalledWith('bad input');
+  });
+
+  it('does not toast for PLATFORM_UNAVAILABLE errors (warns instead)', async () => {
+    global.fetch.mockResolvedValue(
+      makeResponse({ status: 503, json: { error: 'offline', code: 'PLATFORM_UNAVAILABLE' } }),
+    );
+    await expect(request('/x')).rejects.toMatchObject({ code: 'PLATFORM_UNAVAILABLE' });
+    expect(toast.error).not.toHaveBeenCalled();
```
atomantic, line 88: **[IMPROVEMENT]** The test name says PLATFORM_UNAVAILABLE “warns instead,” but this only proves that the error toast is absent. It still passes if the warning toast is removed entirely, so the user-facing warning contract is not pinned.

```suggestion
    expect(toast).toHaveBeenCalledWith('offline', { icon: '⚠️' });
    expect(toast.error).not.toHaveBeenCalled();
```

atomantic (commented):
Reviewed by /do:review — 4 improvements; no critical findings.

## Highlights

- client/src/services/apiCore.js:30 can crash on a valid JSON null error body instead of preserving the HTTP error envelope.
- client/src/services/apiCore.test.js:53 can pass even if throwApiError() stops rejecting.
- client/src/services/apiMedia.test.js:14 replaces the canonical helper with a redirect-free duplicate, so the new auth-redirect contract is not covered through these consumers.
- client/src/services/apiCore.test.js:82 does not assert the warning toast named by the test.

## Out-of-diff observations

PR #5050 currently adds client/src/services/apiSystem.js:downloadBackupSnapshot() with its own raw non-OK response envelope. PR #5056 does not touch that file, so merging both PRs in either order leaves that fifth duplicate in the final tree. If #5050 lands first, please rebase/update this PR so that new function also routes through throwApiError; otherwise the “4 sites (not 6)” claim does not describe the combined result.

## Coherence check

The new helper intentionally changes the non-JSON request() fallback from the previous “Request failed” message to “HTTP <status>”; the added test pins that new wording. If intentional, the PR description should call out that small user-visible change rather than saying request behavior is unchanged.

No automated checks are currently reported for PR #5056, so this review is code-only and should not be treated as CI validation.

_Generated by /do:review_

client/src/services/apiMedia.test.js
```diff
@@ -8,6 +8,17 @@ vi.mock('./apiCore.js', () => ({
   request: (...a) => request(...a),
   API_BASE: '/api',
   maybeRedirectToLogin: vi.fn(),
+  // Mirrors the real apiCore.throwApiError (minus the redirect side effect,
+  // already covered by the maybeRedirectToLogin mock above) so assertions
+  // against `.code`/`.status`/`.context`/message keep working unchanged.
+  throwApiError: vi.fn(async (response) => {
```
atomantic: **[IMPROVEMENT]** This mock reimplements throwApiError() and never invokes the maybeRedirectToLogin mock declared above it. Consequently, cleanImage() and fetchCleanResult() tests can pass while the consumer-side auth redirect contract regresses; prefer importing the real helper and mocking only the request transport, or at minimum exercise the redirect path explicitly.

Evidence: the production callers now do await throwApiError(response), while this test replaces that helper with a redirect-free copy.

Bryandero98:
Thanks for the thorough review! Pushed a follow-up commit addressing all four findings:

- `apiCore.js`: `throwApiError` no longer crashes on a valid-JSON-but-non-object error body (e.g. a bare `null`) — falls back to the same `HTTP {status}` shape as a non-JSON body.
- `apiCore.test.js`: the "omits context" test no longer catches its own sentinel error if `throwApiError` stops rejecting; the `PLATFORM_UNAVAILABLE` test now asserts the warning toast actually fires; added a case for the null-body fallback.
- `apiMedia.test.js`: the mock's `throwApiError` now calls the same `maybeRedirectToLogin` mock the file asserts against, plus a new test exercises a 401 `AUTH_REQUIRED` response through `cleanImage()` to confirm the redirect call happens.

On the out-of-diff note about #5050 reintroducing `apiSystem.js`'s `downloadBackupSnapshot` — makes sense, I'll rebase this on whichever lands first so the combined tree doesn't leave a 5th duplicate. And you're right that the "Request failed" → "HTTP {status}" fallback wording change should've been called out explicitly in the description — noting it here: that's a real, if narrow, user-visible message change for the case where an error response body fails to parse as JSON.

18/18 tests green, Biome clean.

client/src/services/apiMedia.test.js
```diff
@@ -4,20 +4,40 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
...
-  maybeRedirectToLogin: vi.fn(),
-}));
+vi.mock('./apiCore.js', () => {
+  // Mirrors the real apiCore.throwApiError, including the maybeRedirectToLogin
+  // call — routing through the SAME mock the test file asserts against, so a
+  // regression in the consumer-side auth-redirect contract actually fails
+  // these tests instead of a mock that silently drops the side effect.
+  const maybeRedirectToLogin = vi.fn();
+  const throwApiError = vi.fn(async (response) => {
+    const error = await response.json().catch(() => ({ error: `HTTP ${response.status}` }));
```
atomantic: The mock's `throwApiError` implementation diverges from production for valid non-object JSON such as `null`: it dereferences the value and produces a TypeError without the `.status` contract. Mirror `apiCore.js`'s `parsedError && typeof parsedError === 'object'` normalization, or partially mock the real helper so this consumer test cannot drift.

client/src/services/apiMedia.test.js
```diff
@@ -66,6 +86,13 @@ describe('cleanImage', () => {
     global.fetch.mockResolvedValue(makeResponse({ status: 400, ok: false, json: { error: 'No FLUX', code: 'REGEN_BACKEND_UNAVAILABLE' } }));
     await expect(cleanImage(fakeFile, { diffusion: 'gpu' })).rejects.toMatchObject({ code: 'REGEN_BACKEND_UNAVAILABLE' });
```
atomantic, line 92: Please include `context` in this assertion. Preserving structured `error.context` through direct-fetch wrappers is the behavior this refactor is meant to provide; asserting only `.code` would let that propagation regress while the test stays green.

client/src/services/apiCore.test.js
```diff
@@ -0,0 +1,96 @@
...
+  it('warns instead of toasting an error for PLATFORM_UNAVAILABLE', async () => {
+    global.fetch.mockResolvedValue(
+      makeResponse({ status: 503, json: { error: 'offline', code: 'PLATFORM_UNAVAILABLE' } }),
+    );
+    await expect(request('/x')).rejects.toMatchObject({ code: 'PLATFORM_UNAVAILABLE' });
+    expect(toast).toHaveBeenCalledWith('offline', { icon: '⚠️' });
+    expect(toast.error).not.toHaveBeenCalled();
+  });
+
+  it('stays silent when { silent: true } is passed', async () => {
```
atomantic, line 91: This silent-path test covers only the `toast.error` branch. Please add a `PLATFORM_UNAVAILABLE` case and assert that neither `toast` nor `toast.error` is called when `{ silent: true }` is passed.

atomantic (dismissed):
Reviewed by /do:review — REQUEST_CHANGES.

## Blocking: the sixth call site is still present

Current `main` includes `client/src/services/apiSystem.js:169-215` from merged PR #5050. Its non-OK branch at `apiSystem.js:194-203` still constructs the response error envelope locally. The #5056 head (`d4c88bea653f71747caac9b75122c4353c77b9e7`) does not include that function in its changed-file set, so merging this PR leaves the exact sixth duplicate that issue #5047 asks to centralize. The PR description's “4 sites (not 6)” statement is stale against the current base.

Please update/rebase the branch and route `downloadBackupSnapshot` through `throwApiError` as well, including the relevant regression coverage.

## Verified test-contract improvements

- `client/src/services/apiMedia.test.js:14` has a second `throwApiError` implementation that already diverges from production for valid non-object JSON such as `null`; mirror the production normalization or partially mock the real helper.
- `client/src/services/apiMedia.test.js:87` should assert that `error.context` survives through the direct-fetch wrapper, since that is the behavior this refactor is intended to preserve.
- `client/src/services/apiCore.test.js:91` should cover the `PLATFORM_UNAVAILABLE` warning path under `{ silent: true }`, not only the `toast.error` path.

The changed production code had no additional runtime or security findings in the independent review passes. The repository ruleset requires `CI Gate`; PR #5056 currently reports no checks, so that gate must report successfully after the branch is updated.

Bryandero98:
Thanks for catching that - rebased onto current main and routed `apiSystem.js`'s `downloadBackupSnapshot` through `throwApiError` as well (kept the existing abort-on-failure cleanup for the opened file handle, now running before the delegated throw). Updated the PR description accordingly since the earlier "no longer exists" claim was true against main at the time but went stale once #5050 landed.

Also addressed all three test-contract findings:
- `apiMedia.test.js`'s mock now mirrors production's null-safe body normalization instead of diverging from it
- `apiMedia.test.js` and the new `apiSystem.download.test.js` case both assert `error.context` survives through their direct-fetch paths
- `apiCore.test.js` now asserts the PLATFORM_UNAVAILABLE warning toast (not just `toast.error`) is suppressed under `{ silent: true }`

44/44 passing across the four touched test files, Biome clean.

PR #5070, opened 2026-08-25
fix: resolve Jira Cloud assignees before ticket creation (#5046)

### LINKED ISSUE #5046: JIRA Cloud: resolve a configured assignee email to an accountId before createTicket
`createTicket` (`server/services/jira.js`) sends a literal `ticketData.assignee` as `fields.assignee = { name: <value> }` on every instance type. JIRA Cloud retired `name` for GDPR and accepts only `{ accountId }`, so a Cloud instance rejects the assignee outright — the ticket is created unassigned (or the create fails, depending on the project's field config).

The value almost always IS an email: `JiraConfigPanel.jsx` defaults `jira.assignee` to the instance's configured email, and that value flows straight through `promptSections/appContext.js` when the CoS ticket creator files a ticket.

#5042 added a `currentUser` sentinel that resolves the authenticated account and writes `{ accountId }` on Cloud / `{ name }` on Server-DC, which is what the claim flow's epic decomposition uses. It deliberately left the LITERAL path untouched so that change stayed purely additive.

## Scope

- On a Cloud instance, resolve a literal assignee that looks like an email (or a display name) to an `accountId` via `GET /rest/api/2/user/search?query=<value>` before the create, and cache it per instance so every ticket does not pay a second round-trip.
- An unresolvable value is a sentinel case: report it (the ticket still gets created, unassigned) rather than silently sending a field JIRA will reject.
- Server/DC keeps `{ name }` — nothing changes there.

## Acceptance

- A Cloud instance configured with an assignee email files tickets that are actually assigned to that user.
- A Server/DC instance's behavior is unchanged.
- `server/services/jira.test.js` covers both instance types plus the unresolvable case.

Related to #5042

atomantic:
## Summary
- Resolve configured Jira Cloud assignee emails and display names to GDPR-compatible account IDs before ticket creation.
- Cache positive lookups per instance with bounded storage and single-flight coalescing, while safely retrying unresolved or failed lookups.
- Preserve Server/DC `{ name }` assignees and cover the Cloud edge cases with focused tests.

## Test plan
- `npm test --prefix server -- services/jira.test.js` (35 tests)

Closes #5046

### DIFF
--- a/server/services/jira.js
+++ b/server/services/jira.js
@@ -4,13 +4,36 @@
  */
 
 import fs from 'fs/promises';
+import { createBoundedStateMap } from '../lib/boundedStateMap.js';
 import { createHttpClient } from '../lib/httpClient.js';
+import { createSingleFlight } from '../lib/singleFlight.js';
 import path from 'path';
 import { ensureDir, PATHS, readJSONFile } from '../lib/fileUtils.js';
 import { hostFromOriginUrl } from '../lib/workTracker.js';
 import { countConfiguredInstances } from '../lib/instanceFeatureRegistry.js';
 
 const JIRA_CONFIG_FILE = path.join(PATHS.data, 'jira.json');
+
+// Resolved Cloud user-search results are stable until the instance credentials
+// change, and ticket creation can happen repeatedly in one run. Keep only
+// positive results in a bounded per-instance cache so a transient failure or an
+// initially hidden user can recover on the next ticket.
+const cloudAssigneeCache = new Map();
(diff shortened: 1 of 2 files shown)


atomantic:
CI remains blocked after one rerun by an unrelated Server tests failure in server/services/backup.test.js (subdirFilter rsync flags / echoes the subdirFilter back in the result) timing out after 10s. The selected Jira tests passed; no Jira-related failure was reported. Leaving this PR open for follow-up.

PR #5078, opened 2026-08-26
fix: centralize client API fetch errors (#5047)

### LINKED ISSUE #5047: Extract the duplicated fetch error-envelope in the streaming API wrappers into apiCore

atomantic:
## Summary

- Extract the shared fetch error envelope into `throwApiError` in `apiCore`.
- Route raw stream/blob responses through it so code, status, and structured context stay consistent.
- Preserve connection errors and abort semantics for the affected fetch wrappers.

Closes #5047

## Test plan

- `npm test` (client)
- `npm run build` (client)
- `npx biome lint --error-on-warnings` on changed client service files

### DIFF
--- a/client/src/services/README.md
+++ b/client/src/services/README.md
@@ -29,7 +29,7 @@ toasts on throw). **Custom catch ⇒ `silent: true`** — otherwise toasts fire 
 | File | Purpose |
 |---|---|
 | `api.js` | Barrel — re-exports every `apiX.js`. |
-| `apiCore.js` | `request()` helper + stable PortOS-app id. Shared error / toast handling. |
+| `apiCore.js` | `request()` and `throwApiError(response)` helpers + stable PortOS-app id. Shared error / toast handling. |
 | `apiBatch.js` | `fetchByIds(path, ids, options)` — batch-fetch records through a list route's `?ids=a,b,c` filter (dedupe, empty-list short-circuit, `{ items }` envelope unwrap). |
 | `socket.js` | Singleton Socket.IO client over relative path (Tailscale-friendly). |
 | `appUrls.js` | Compute candidate launch URLs for an app from page context. |
--- a/client/src/services/apiCore.js
+++ b/client/src/services/apiCore.js
@@ -21,6 +21,19 @@ export function maybeRedirectToLogin(response, error) {
   }
 }
 
+// Shared by raw-response callers that need to consume a stream or blob instead
+// of handing the response to request(). Keep the envelope construction here so
+// structured server context and status metadata survive every API surface.
+export async function throwApiError(response, parsedError) {
+  const error = parsedError ?? await response.json().catch(() => ({ error: `HTTP ${response.status}` }));
+  maybeRedirectToLogin(response, error);
+  const err = new Error(error.error || `HTTP ${response.status}`);
+  err.code = error?.code;
+  err.status = response.status;
+  if (error?.context) err.context = error.context;
+  throw err;
+}
+
 export async function request(endpoint, options = {}) {
   const { silent, responseType, ...fetchOptions } = options;
   const url = `${API_BASE}${endpoint}`;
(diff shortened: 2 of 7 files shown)


atomantic:
Closing as a duplicate: merged PR #5056 already delivered #5047 on the default branch (merge commit 74d540475).

atomantic:
Redundant — #5047 was already solved on `main` by #5056 (merged), which landed `throwApiError` in `apiCore.js` and routed the same wrappers (`apiImageVideo`, `apiLocalLlm`, `apiMedia`, `apiSystem`) through it. Issue #5047 is closed.

Two independent claims of the same issue. Main's version is the more complete one — it ships a dedicated `apiCore.test.js` covering `throwApiError` (non-object JSON bodies, auth-gate redirect, structured `context` forwarding) that this branch doesn't have.

The one behavioural difference is `fetchCleanResult` in `apiMedia.js`: this branch adds a second `parsedError` parameter so the 409 path can reuse `throwApiError` after the body has already been read. Main keeps that block hand-rolled instead, which is correct as-is — the response body is consumed by the `RESULT_NOT_READY` check, so a body-reading `throwApiError(response)` can't be used there, and the path is a 409 that never carries the auth-gate 401 the helper redirects on.

Closing as superseded.

PR #5262, opened 2026-08-27
fix: classify all audit tasks in workflow graph

### LINKED ISSUE #5244: Six scheduled audit types render under Ambient instead of Audit in the workflow graph
### Problem

`AUDIT_STAGE.taskTypes` in `server/services/workflow.js:66-80` enumerates the audit task types the workflow graph groups under the **Audit** stage. Six audit types registered in `SELF_IMPROVEMENT_TASK_TYPES` are absent from it:

- `data-safety` and `simplify` — shipped earlier and never added
- `api-contract`, `react-lifecycle`, `observability`, `copy` — added in #5237

### Trigger

`buildWorkflowGraph` builds `itemStage` from the stage tables, then at `server/services/workflow.js:157` falls back for any task not found:

```js
const stageId = itemStage.get(`task:${taskType}`) || 'ambient';
```

So any user who enables one of those six scheduled audits sees its node rendered under **Ambient** in the workflow view instead of **Audit**, alongside unrelated always-on work.

### Impact

The workflow graph is the page users read to understand what their install runs and when. Six audits — half the audit catalog — are grouped wrongly, which makes the Audit stage look emptier than it is and the Ambient stage look like a junk drawer.

### Fix

Add all six ids to `AUDIT_STAGE.taskTypes` in `server/services/workflow.js`.

Deliberately **not** folded into #5237: moving `data-safety` and `simplify` is a visible regrouping of two pre-existing types in the user's workflow diagram, which is a behavior change that deserves its own reviewable commit rather than riding along in a labels/prompt PR.

Rejected alternative: deriving the stage list from `isAuditTaskType()` instead of a literal list. It would prevent the drift permanently, but the stage tables are intentionally hand-ordered for presentation, and `AUDIT_STAGE` also carries `jobIds` that no predicate covers — so the literal list stays, with a guard test instead.

### Acceptance criteria

- [ ] `AUDIT_STAGE.taskTypes` contains `data-safety`, `simplify`, `api-contract`, `react-lifecycle`, `observability`, and `copy`
- [ ] A test in `server/services/workflow.test.js` asserts every `isAuditTaskType()` id resolves to stage `audit`, so a future audit type cannot land unstaged
- [ ] No task type appears in more than one stage list

atomantic:
## Summary
- Classify all six scheduled audit task types under the Audit workflow stage.
- Add a regression guard tied to the audit catalog so new audit types cannot silently fall through to Ambient.

## Test plan
- `npx vitest run services/workflow.test.js`
- `npm test -- workflow.test.js ../lib/auditCatalog.test.js`

Closes #5244

### DIFF
--- a/server/services/workflow.js
+++ b/server/services/workflow.js
@@ -76,7 +76,13 @@ export const WORKFLOW_STAGES = [
       'typing',
       'ui-bugs',
       'mobile-responsive',
-      'ux'
+      'ux',
+      'data-safety',
+      'simplify',
+      'api-contract',
+      'react-lifecycle',
+      'observability',
+      'copy'
     ],
     jobIds: ['job-wiki-maintenance', 'job-refresh-local-llm-catalog', 'job-refresh-cli-provider-catalogs']
   },
--- a/server/services/workflow.test.js
+++ b/server/services/workflow.test.js
@@ -18,6 +18,7 @@ const { getScheduleStatus } = await import('./taskSchedule.js');
 const { getAllJobs } = await import('./autonomousJobs.js');
 const { checkJobGate, hasGate, getRegisteredGates } = await import('./jobGates.js');
 const { getWorkflowGraph, projectWorkflowTimeline, WORKFLOW_STAGES } = await import('./workflow.js');
+const { AUDIT_TASK_TYPES } = await import('../lib/auditCatalog.js');
 
 const STAGE_IDS = WORKFLOW_STAGES.map(s => s.id);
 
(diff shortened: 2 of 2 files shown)


atomantic:
Review gate remains unsatisfied: the pinned local Codex reviewer (gpt-5.6-luna, model_reasoning_effort=high) was interrupted before producing a usable verdict. Focused workflow and audit-catalog tests passed, but this PR is intentionally left open with its claim markers for cold pickup; it was not merged.

PR #5334, opened 2026-08-28
fix: derive Spotify redirects from Tailscale host (#5321)

### LINKED ISSUE #5321: Use the active Tailscale hostname for Spotify OAuth redirect URIs
## Problem / Goal

Make Spotify's auto-derived OAuth redirect URI match the endpoint PortOS is actually serving when Tailscale HTTPS is active, without requiring users to configure `PUBLIC_HOST`, while preserving explicit overrides and the existing localhost fallback.

## Context

`getRedirectUri()` in `server/services/spotifyAuth.js:53-58` currently returns `SPOTIFY_REDIRECT_URI` verbatim when set; otherwise it builds `http://${PUBLIC_HOST || 'localhost'}:${PORT || 5555}/api/spotify/oauth/callback`.

PortOS already exposes the required runtime signals:

- `getSelfHost()` in `server/lib/peerSelfHost.js:11-19` returns `PORTOS_HOST` when explicitly configured, or the hostname from Tailscale certificate metadata when `meta.mode === 'tailscale'`; it returns `null` for self-signed or missing metadata.
- `getHttpsEnabledAtBoot()` in `server/lib/httpsState.js:22-24` reports whether the current process booted with HTTPS enabled.
- `server/lib/networkExposure.js:37-47` already combines these helpers to report the active scheme and Tailscale hostname — the pattern to follow, not one invented for this task.

`server/services/spotifyAuth.test.js:60-77` covers the `SPOTIFY_REDIRECT_URI` override and `PUBLIC_HOST`, but not the localhost fallback or the new auto-detected path.

`getAuthStatus()` in `server/services/spotifyAuth.js:225-234` already returns `getRedirectUri()`, so the resulting URI reaches the Settings UI (`client/src/components/settings/SpotifyTab.jsx`) with no client change needed. `.env.example:41-50` currently documents localhost as the default and describes manual Tailscale configuration.

`server/services/googleAuth.js:14` has the identical `PUBLIC_HOST || 'localhost'` pattern for the Google Calendar OAuth redirect — same gap, out of scope here (see below).

## Proposed approach

In `server/services/spotifyAuth.js`:

1. Import `getSelfHost` from `server/lib/peerSelfHost.js` and `getHttpsEnabledAtBoot` from `server/lib/httpsState.js`.
2. Preserve the existing precedence:
   - Return `SPOTIFY_REDIRECT_URI` verbatim when set.
   - If `PUBLIC_HOST` is set, continue building the existing `http://PUBLIC_HOST:PORT/...` form (an explicit user override shouldn't be silently superseded by auto-detection).
3. When neither explicit override is set, call `getSelfHost()`.
   - If it returns a hostname, use that hostname.
   - If it returns `null`, continue using `localhost`.
4. For the auto-detected-host branch, derive the scheme from `getHttpsEnabledAtBoot().value`: `https` when true, `http` when false. (When Tailscale HTTPS is active, PortOS serves `:5555` over TLS per `docs/PORTS.md`, so the redirect URI must match.)
5. Keep the existing `PORT` value and `/api/spotify/oauth/callback` path unchanged.
6. Update `.env.example` to explain that Spotify automatically uses the active Tailscale hostname and HTTPS state when available, while `PUBLIC_HOST` and `SPOTIFY_REDIRECT_URI` remain explicit overrides.

## Acceptance criteria

- [ ] With neither `SPOTIFY_REDIRECT_URI` nor `PUBLIC_HOST` set, `getSelfHost()` returning `host-example.ts.net`, `getHttpsEnabledAtBoot().value === true`, and `PORT=5555`, `getRedirectUri()` returns `https://host-example.ts.net:5555/api/spotify/oauth/callback`.
- [ ] With neither override set, `getSelfHost()` returning a hostname and HTTPS disabled, `getRedirectUri()` returns the same hostname with the `http` scheme.
- [ ] With neither override set and `getSelfHost()` returning `null`, `getRedirectUri()` returns `http://localhost:<port>/api/spotify/oauth/callback` (unchanged default).
- [ ] `SPOTIFY_REDIRECT_URI` continues to be returned verbatim, regardless of detected host or HTTPS state.
- [ ] `PUBLIC_HOST` continues to take precedence over auto-detection and preserves the current `http://PUBLIC_HOST:<port>/...` behavior.
- [ ] `server/services/spotifyAuth.test.js` mocks the host/HTTPS helpers and covers the explicit URI override, `PUBLIC_HOST`, localhost fallback, Tailscale-host/HTTPS path, and Tailscale-host/HTTP path.
- [ ] `getAuthStatus()` continues to expose the URI returned by `getRedirectUri()`.
- [ ] `.env.example` documents the new Tailscale auto-detection behavior and retains both environment variables as explicit overrides.

## Out of scope

- `server/services/googleAuth.js`'s identical `PUBLIC_HOST || 'localhost'` pattern for the Google Calendar redirect URI — a separate follow-up if wanted, not bundled here.
- Changing `getSelfHost()`, `getHttpsEnabledAtBoot()`, certificate detection, `PORTOS_HOST`, or server startup behavior.
- Changing Spotify routes, credential storage, `SpotifyTab.jsx`, or `apiSpotify.js` — the redirect URI is server-derived and already flows to the UI via `getAuthStatus()`.
- Automatically changing the redirect URI registered in the Spotify developer dashboard.

atomantic:
## Summary

- Derive Spotify OAuth callback host and scheme from active Tailscale HTTPS state when no explicit override is configured.
- Preserve `SPOTIFY_REDIRECT_URI`, `PUBLIC_HOST`, and localhost fallback precedence.
- Reject invalid certificate-metadata placeholder hosts and document the derived behavior.

## Tests

- `cd server && npm test -- --run services/spotifyAuth.test.js`

Closes #5321

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -43,15 +43,14 @@ PGPASSWORD=portos
 # HOST=0.0.0.0
 
 # Public-facing hostname baked into OAuth redirect URIs for Google Calendar/Gmail
-# (googleAuth.js) and Spotify (spotifyAuth.js). Defaults to "localhost". Set this to
-# the host the OAuth provider will redirect back to (e.g. your Tailscale MagicDNS name)
-# so the callback URL matches what you registered in the provider dashboard.
+# (googleAuth.js) and Spotify (spotifyAuth.js). Google and Spotify default to
+# "localhost"; Spotify otherwise auto-detects the active Tailscale hostname and
+# HTTPS state. Set this explicit override when the provider must use another host.
 # PUBLIC_HOST=host-example.ts.net
 
 # Explicit override for the Spotify OAuth redirect URI (spotifyAuth.js). When set, it is
-# used verbatim instead of the auto-derived http://PUBLIC_HOST:PORT/api/spotify/oauth/callback.
-# It must match the redirect URI registered in the Spotify dashboard byte-for-byte
-# (use this when your public URL is HTTPS or differs from the auto-derived form).
+# used verbatim instead of PUBLIC_HOST or the active Tailscale hostname/HTTPS state.
+# It must match the redirect URI registered in the Spotify dashboard byte-for-byte.
 # SPOTIFY_REDIRECT_URI=https://host-example.ts.net:5555/api/spotify/oauth/callback
 
 # OpenClaw integration (configured via Settings UI; env overrides take precedence)
--- a/server/services/loras.test.js
+++ b/server/services/loras.test.js
@@ -1091,10 +1091,11 @@ describe('LoRA key layout', () => {
     });
     const list = await lorasService.listLoras();
     expect(list.find((l) => l.filename === 'lora-comfy.safetensors').keyLayout).toBe('comfyui');
-    // Sidecar write is fire-and-forget; let the microtask queue drain.
-    await new Promise((r) => setTimeout(r, 10));
-    const sidecar = JSON.parse(await fs.readFile(join(tmpLoras, 'lora-comfy.safetensors.metadata.json'), 'utf-8'));
-    expect(sidecar.keyLayout).toBe('comfyui');
+    // Sidecar write is fire-and-forget; wait for it to complete.
+    await vi.waitFor(async () => {
+      const sidecar = JSON.parse(await fs.readFile(join(tmpLoras, 'lora-comfy.safetensors.metadata.json'), 'utf-8'));
+      expect(sidecar.keyLayout).toBe('comfyui');
+    });
   });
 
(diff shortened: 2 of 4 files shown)


atomantic:
Serialized swarm merge gate: CI was re-run once. Server and Windows jobs failed again on the pre-existing stale `server/lib/socketEventCatalog.generated.json` assertion (calendar source-line drift); DB and client jobs passed. Leaving this PR open and its claim worktree/branch intact for cold pickup.

PR #5431, opened 2026-08-30
Unfreeze scheduled prompts stuck on the pre-unification generations

atomantic:
## Summary

Nine scheduled improvement-task prompts on this install have been **frozen out of every prompt upgrade since Feb 2026** — including the `feature-ideas` prompt that dispatched this run, which still points at `data/COS-GOALS.md`, a file the same change folded into `GOALS.md`.

**Root cause.** Unifying the self-improvement and app-improvement schedules replaced two prompt generations — `[Self-Improvement] …` and `[App Improvement: {appName}] …` — with a single `[Improvement: {appName}] …` body and preserved **neither** in `PREVIOUS_DEFAULT_PROMPTS`. Installs holding either generation stopped matching any shipped default, were stamped `promptCustomized` by the legacy migration, and became permanently ineligible for auto-upgrade.

**Fix** — the documented distribution contract, per task type:

- Register the **12 outgoing bodies** (11 task types) extracted from the pre-unification prompt maps. Per-type registration is what makes coverage auditable; a blanket string rewrite in the shared predicate was tried first and rejected on review — it widened "is a shipped default" for every consumer and hid which types were covered.
- **Finish the upgrade path.** Clearing `promptCustomized` alone wasn't enough: the legacy migration stamped `promptVersion = PROMPT_VERSIONS[taskType]` alongside the flag, so an install flagged after its type was versioned held the *current* version with a *retired* body — `stored < current` was false and the stale prompt survived, now un-flagged. The self-heal now resets the version to 1 whenever the matched body is a prior default.
- **Close the parity gap that caused it.** `console-errors`, `error-handling` and `typing` had no `PROMPT_VERSIONS` entry at all, so the upgrade block skipped them outright. They (plus `jira-sprint-manager` / `jira-status-report`) are versioned, and a new guard asserts every persisted prompt is versioned against an explicit allowlist of the five unpersisted pipeline-stage bodies.
- **De-duplicate the predicate.** `promptMatchesShippedDefault` moves to `taskPromptDefaults/shippedPrompts.js`; `taskScheduleStore`, `taskPromptService` and `autonomousJobs/portosCatalogRefresh` each carried a byte-identical copy.

Also bundled, because it lands in the same version/previous-default contract files: **`feature-ideas` v11** establishes product-source precedence when brainstorming — root `PRD.md` first, then `GOALS.md`, then README/docs — mirroring what `plan-feature` v2 already does. v10 read `GOALS.md` only, so a repo whose product intent lives in a PRD (PortOS included) had its most specific source unread. v10 is preserved.

## Test plan

- `cd server && npx vitest run` — 35,776 passed, 19 skipped.
- New cases in `taskSchedule.test.js` cover each shape the freeze took (header-only drift, a body that also changed, a type with no version entry, the older self-improvement generation), each under **both** version stamps a frozen install can carry. Verified non-vacuous: all five current-stamp cases fail without the store fix.
- `taskPromptDefaults.test.js` pins the v11 bump + preserved v10, and the new persisted-prompt version parity guard.
- Snapshot reproducibility re-checked across origins (`PORT=5554`, `PORT=3000`, `PORTOS_HOST=…`, `PORTOS_API_URL=http://localhost`) — all green.
- Verified against real stored schedule state: all nine frozen types now self-heal and converge on the current default; none remain stuck.

## Known limitation (pre-existing, widened)

`updateTaskInterval` sets `promptCustomized` purely on "a prompt was supplied", so it cannot distinguish a flag written by the buggy legacy migration from a deliberate user pin to an older shipped body. The self-heal now un-pins 12 more bodies than before. Distinguishing the two needs a user-write marker in the stored shape — out of scope here; filing a follow-up.


atomantic:
Follow-up for the known limitation filed as #5432 (prompt-customization provenance).

Issue #1077, opened 2026-06-09
Brain sync is delta-log-only — diverged peers never re-converge (no anti-entropy)


atomantic:
Discovered while verifying void ↔ null sync on 2026-06-09. **null was missing 21 live brain records that exist on void** (12 links, 7 projects, 2 ideas), with no mechanism to deliver them. Re-seeded manually (no-op PUT on void re-logged each record → null pulled them; verified zero diff afterward), but the underlying protocol gaps will re-orphan records on the next compaction.

## Root cause

Unlike the snapshot categories (`goals`/`character`/`universe`/`pipeline`/etc. in `dataSync.js`, which have a checksum+snapshot reconcile and self-heal), **brain sync is delta-log-only** (`server/services/brainSync.js`, `server/services/brainSyncLog.js`, `syncBrainFromPeer` in `syncOrchestrator.js`). Two compounding bugs:

### Bug 1 — compaction floor uses the wrong cursor
`syncAllPeers()` (`server/services/syncOrchestrator.js:777-789`) computes the compaction floor from `cursors[peerId].brainSeq` — that is **void’s *outbound pull* cursor into the peer** (how much of the peer’s log *we* consumed), NOT "how much of *our* log the peer has consumed." So void drops sync-log entries before peers have actually pulled them. The peer-consumption value already exists — peers report it as `cursorForYou` in the `/api/sync/status?forPeer=<id>` payload (confirmed in null’s probe of void). Compaction should floor on `min(peers’ cursorForYou into us)`, not our pull cursors.

### Bug 2 — no anti-entropy / snapshot reconcile for brain
Once a record’s create/update entry is compacted out of the log, a peer that never pulled it (was offline, cursor was stale, or sync was disabled during the window) can **never** learn its current state. Brain is therefore eventually-*divergent*, not eventually-consistent. There is no brain checksum/snapshot endpoint (confirmed: no `/api/brain/snapshot`, no brain entry in `dataSync.getSupportedCategories()`).

## Fix

1. **Anti-entropy reconcile for brain** — add per-type checksum + snapshot endpoints mirroring the `dataSync` pattern (`/api/sync/brain/checksum` + `/api/sync/brain/snapshot`, `forPeer`-scoped). On checksum mismatch, pull the snapshot and LWW-merge by `updatedAt` (apply is already idempotent via `applyRemoteRecord`). This makes a compacted-past peer re-converge regardless of log state.
[...]

PR #5462, opened 2026-08-30
fix(brain): compact sync log on installs with no brain-sync peers

### LINKED ISSUE #5439: Brain sync log never compacts on an install with no brain-sync peers
## Problem

`syncAllPeers()` (`server/services/syncOrchestrator.js:829-877`) compacts the brain sync log only inside `if (brainPeers.length > 0)` (:870-876). On an install with **no enabled brain-sync peer** — the default, since federation is opt-in — `compactLog` is never called, so every `create` / `update` / `upsertWithId` / `remove` on all ten brain types (`brainStorage.js:458, 491, 540, 592, 638, 684, 725`) appends a full-record entry to `data/brain/sync_log.jsonl` that nothing will ever read.

The guard was added in 5217deb16 (#1077) to avoid `Math.min()` over an empty array (`Infinity`), not as a decision to keep an unread log forever — the comment beside it is about a *reporting-less peer* floor of 0, which is a different case.

## Trigger

Any single-machine install writing brain records. The heaviest producer is the Daily Log: `DailyLogTab.jsx:20-21` autosaves on a 1.5 s debounce / 10 s max-wait, and each PUT → `setJournalContent` → `upsertWithId` appends the day's **entire** `content` *and* `segments` (which duplicate the text, `brainJournal.js:361-366`) as one log line. An hour of journaling adds hundreds of multi-KB entries; a ChatGPT import adds ~1,400 entries of up to 9.8 KB each. None of it is ever pulled.

## Impact

- `sync_log.jsonl` grows without bound and lives under `data/`, so it rides every backup.
- `initSyncLog()` (`server/services/brainSyncLog.js:27-46`) `readFile`s the whole file into one string and `split`s it just to read the last line — boot latency and a 2–3× file-size memory spike scale with the user's entire brain edit history.
- The first peer ever added pulls that whole history 100 entries at a time (`syncOrchestrator.js:172`), each page re-parsing the full file (see the companion read-path issue), even though the #1077 anti-entropy snapshot would converge it in one exchange.

## Fix

In `syncAllPeers`, add an `else` branch to the `brainPeers.length > 0` block:

```js
} else if (brainSyncLog.getCurrentSeq() > 0) {
  // No brain-sync peer will ever pull these entries; a peer added later
  // converges through the reconcile snapshot (#1077). Keep the last entry so
  // initSyncLog still recovers the sequence counter across restarts.
  await brainSyncLog.compactLog(brainSyncLog.getCurrentSeq());
}
```

`compactLog(minSeq)` keeps entries with `seq >= minSeq` (`brainSyncLog.js:147`), so the newest entry survives and the counter never resets to 0 — which matters because a peer that was removed and later re-added still holds a `cursorForYou.brainSeq > 0` and would otherwise skip new entries until reconcile caught it. The `> 0` guard keeps a fresh install from rewriting an empty file every minute. `compactLog` already logs only when it runs; it is a no-op rewrite when nothing is below the floor, so add a cheap `dropped === 0` early return inside `compactLog` (or check `lines.length > 1` before rewriting) so an idle install doesn't rewrite the file every 60 s.

**Rejected:** skipping the `appendChange` entirely when there are no peers — changes the write contract at seven call sites and leaves a peer added later with no delta log at all. **Rejected:** `compactLog(Infinity)` / truncating to empty — resets the sequence counter on the next boot (see above).

**Files:** `server/services/syncOrchestrator.js` (~5 lines), `server/services/brainSyncLog.js` (no-op guard), `server/services/syncOrchestrator.test.js`, `server/services/brainSyncLog.test.js`.

**Tests:** in `syncOrchestrator.test.js` next to "floors at 0 when a brain-enabled peer has not reported its cursor into us" (~:838): (a) with peers configured but none brain-enabled, `compactLog` is called with `getCurrentSeq()`; (b) with `getCurrentSeq() === 0`, `compactLog` is not called; (c) the existing two floor tests stay green. In `brainSyncLog.test.js`: `compactLog(currentSeq)` keeps exactly the last line and a subsequent `initSyncLog` recovers that seq.

## Acceptance criteria

- [ ] On an install with no enabled brain-sync peer, `sync_log.jsonl` never holds more than the newest entry after a sync cycle.
- [ ] After that compaction and a restart, `getCurrentSeq()` equals the pre-restart value (no reset to 0).
- [ ] An idle install (no new entries since the last cycle) does not rewrite the file every cycle.
- [ ] Existing peer-floor behavior (min over `cursorForYou.brainSeq`, 0 for an unreported peer) is unchanged; `cd server && npm test` green.

prateeekbuilds:
## Description

Fixes #5439.

On installs without enabled brain-sync peers (the default single-machine posture), `compactLog` was previously never called because compaction only ran within `if (brainPeers.length > 0)`. As a result, mutations across brain types continuously appended entries to `data/brain/sync_log.jsonl` without bound, increasing startup latency and memory spikes in `initSyncLog()`.

### Key Changes

1. **Compatibility-Preserving Terminal LWW Replay Compaction (`server/services/brainSyncLog.js`)**:
   - Replaced destructive single-newest-delta truncation with terminal LWW replay per `(type, id)` (following `applyRemoteRecord` tie-break rules and Migration 080 replay).
   - Prunes redundant intermediate update churn while preserving the active state for every entity so inbound, asymmetric, or pre-#1077 delta-only consumers pulling `GET /api/brain/sync?since=0` receive all records.
   - When compacting with a positive floor (`minSeq > 0`), retains pre-floor winning state if tail operations are stale (`tailTs <= olderWinner.updatedAt`), preventing record resurrection on fresh/delta-only peers.
   - Always preserves the durable maximum sequence number (`maxDurableSeq`) on the survivor to anchor peer cursors and sequence recovery.

2. **Durable Sequence Resolution Under Log Mutex (`server/services/brainSyncLog.js` & `server/services/syncOrchestrator.js`)**:
   - Resolves `maxDurableSeq` strictly from on-disk entries under `withLock` inside `compactLog()`, guarding against index/sequence skew if an append previously failed.
   - In `syncOrchestrator.js`, `syncAllPeers()` passes `minSeq` (minimum reported peer cursor, or `0` when no brain peers exist) rather than an unmutexed `getCurrentSeq()`.

3. **Avoid Idle Rewrites (`server/services/brainSyncLog.js`)**:
   - Added early exit `if (dropped <= 0) return 0;` before `atomicWrite()`, avoiding disk rewrites and log noise on idle sync cycles.

4. **Tests & Verification (`server/services/brainSyncLog.test.js` & `server/services/syncOrchestrator.test.js`)**:
   - Added delegating spy on `atomicWrite` asserting absence of writes when `dropped === 0`.
   - Added test fixtures verifying terminal LWW replay, positive-floor stale-tail LWW retention, and durable sequence recovery after failed appends.
   - Updated `syncOrchestrator.test.js` to assert `compactLog(0)` is invoked when no brain peers exist.

## Testing & Verification

- `npx vitest run server/services/brainSyncLog.test.js server/services/syncOrchestrator.test.js` passed (73 tests).
- 114 brain storage, reconcile, sync, parity, and migration tests passed.
- Full server test suite passed (36,082 tests).

### DIFF
--- a/server/services/brainSyncLog.js
+++ b/server/services/brainSyncLog.js
@@ -237,9 +237,38 @@ export async function getChangesSince(sinceSeq, limit = 100) {
 }
 
 /**
- * Compact the log by dropping entries below minSeq
+ * Replay entries for a single (type, id) according to runtime LWW rules,
+ * returning the surviving terminal entry.
+ *
+ * Incumbent wins ties (same timestamp) matching applyRemoteRecord.
+ * Entries missing updatedAt are skipped.
  */
-export async function compactLog(minSeq) {
+function replayTerminal(entries) {
+  let accepted = null;
+  for (const e of [...entries].sort((a, b) => a.seq - b.seq)) {
+    const ts = e?.record?.updatedAt;
+    if (ts == null) continue;
+    if (accepted == null || ts > accepted.record.updatedAt) accepted = e;
+  }
+  return accepted;
+}
+
+/**
+ * Compact the sync log using a compatibility-preserving compaction representation.
+ *
+ * Deltas at or above `minSeq` (unconsumed by at least one active peer) are
+ * preserved verbatim in sequence order.
+ *
+ * For history below `minSeq` (or all entries when minSeq is 0, e.g. on installs
+ * with no brain-sync peers), redundant intermediate updates are pruned by
+ * retaining only the surviving terminal LWW entry per (type, id).
+ *
+ * The durable sequence counter (maxSeq) is always determined from disk state
+ * under the log mutex, preventing index skew after a failed append, and is
+ * preserved so initSyncLog recovers the monotonic sequence counter across
+ * restarts.
+ */
+export async function compactLog(minSeq = 0) {
   return withLock(async () => {
     await ensureBrainDir();
     // Load first: the rebuild below marks the index loaded, so skipping this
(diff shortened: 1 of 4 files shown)


server/services/syncOrchestrator.js
```diff
@@ -874,6 +874,11 @@ export async function syncAllPeers() {
     });
     const minSeq = Math.min(...consumedSeqs);
     await brainSyncLog.compactLog(minSeq);
+  } else if (brainSyncLog.getCurrentSeq() > 0) {
+    // No brain-sync peer will ever pull these entries; a peer added later
+    // converges through the reconcile snapshot (#1077). Keep the last entry so
+    // initSyncLog still recovers the sequence counter across restarts.
+    await brainSyncLog.compactLog(brainSyncLog.getCurrentSeq());
```
atomantic: **[CRITICAL]** Choose the compaction floor from durable state under the log mutex

`appendChange` and `appendChanges` reserve `currentSeq` before `appendFile`; a failed append leaves `indexLoaded = false` but the exported sequence ahead of disk. This call can pass N+1, then `compactLog()` reloads durable N and still filters with N+1, dropping every line; a restart resets the sequence to 0 and can reuse numbers that retained peer cursors already consumed. Resolve and check the durable current sequence inside `brainSyncLog` while holding its mutex, such as an atomic compact-to-current operation, rather than reading it here.

server/services/syncOrchestrator.js
```diff
@@ -874,6 +874,11 @@ export async function syncAllPeers() {
     });
     const minSeq = Math.min(...consumedSeqs);
     await brainSyncLog.compactLog(minSeq);
+  } else if (brainSyncLog.getCurrentSeq() > 0) {
+    // No brain-sync peer will ever pull these entries; a peer added later
```
atomantic: **[CRITICAL]** Local pull settings do not prove the log has no remote consumers

`getEffectiveCategories(peer).brain` controls what this install pulls, but inbound or asymmetric peers can still call `GET /api/brain/sync`; announced peers begin sync-disabled and older peers may remain one-directional. Pre-#1077 peers also have no reconcile endpoint and explicitly fall back to delta-only, so keeping only the newest delta permanently strands all earlier records. Use a compatibility-preserving bootstrap or compaction representation, or a proven remote-consumer capability gate; the local category map cannot justify destructive truncation.

server/services/brainSyncLog.test.js
```diff
@@ -434,5 +434,46 @@ describe('brainSyncLog', () => {
...
+        await appendChange('create', 'people', `p${i}`, { name: `Person ${i}` }, 'inst-1');
+      }
+      const beforeContent = readFileSync(syncLogPath(), 'utf8');
+
+      // Calling compactLog with minSeq <= 1 drops nothing
+      const dropped = await compactLog(1);
+      expect(dropped).toBe(0);
+
+      const afterContent = readFileSync(syncLogPath(), 'utf8');
+      expect(afterContent).toBe(beforeContent);
```
atomantic: **[IMPROVEMENT]** Assert the absence of a write, not only identical bytes

Removing the new `if (dropped === 0) return 0` would still reconstruct and atomically write the same bytes, so this before/after content comparison passes while the idle-rewrite regression returns. Wrap `atomicWrite` with a delegating spy in the `fileUtils` proxy and assert it was not called after `compactLog(1)`.

atomantic (dismissed):
Reviewed by /do:review — 2 critical, 1 improvement, 0 nits.

### Highlights
- server/services/syncOrchestrator.js:881 — the compaction floor can exceed the durable sequence after a failed append, emptying the log and resetting sequence recovery.
- server/services/syncOrchestrator.js:878 — local pull categories do not account for inbound/asymmetric or pre-reconcile delta consumers.

### Coherence check
The patch matches issue #5439, but its later-peer reconciliation claim is false for supported pre-#1077 delta-only peers.

_Generated by /do:review_

prateeekbuilds:
Thanks for the review! I have addressed all the feedback in the latest commit:

Compatibility-preserving compaction: Switched from single newest delta truncation to terminal LWW replay compaction per (type, id), preserving active entity state for delta-only/pre-#1077 consumers and retaining max sequence numbers.

Mutex-guarded durable floor: Compaction floor and sequence recovery are now determined under the log mutex from disk state, preventing skew after failed appends.

Idle-write test assertion: Added a spy on atomicWrite to verify no writes occur when dropped === 0.

server/services/brainSyncLog.js
```diff
@@ -248,33 +277,114 @@ export async function compactLog(minSeq) {
...
+          unindexedOrUntypedOlder.push(item);
+        }
+      }
+    }
+
+    // Replay terminal winning state for older keys not already superseded in tail
+    const keptOlder = [];
+    const olderWinnersByKey = new Map();
+    for (const [key, items] of olderEntriesByKey) {
+      if (tailKeys.has(key)) {
```
atomantic: **[CRITICAL] Preserve the pre-floor LWW winner when the tail is stale**

`tailKeys` proves only that a later sequence exists, not that its `updatedAt` wins. The new test already supplies the counterexample: seq 2 is a Jan-02 delete and seq 99 is a stale Jan-01 create; with a positive floor such as `compactLog(50)`, this skip drops the delete and serves only the stale create. A fresh, asymmetric, or pre-#1077 peer has no incumbent, accepts that create, and can resurrect the record without a reconcile endpoint to heal it. Retain the replayed older winner before the verbatim tail (or compare the actual LWW winner across older history and tail), and add the same fixture with a positive floor.

atomantic (dismissed):
Reviewed by /do:review — 1 critical, 1 improvement, 0 nits.

### Highlights
- `server/services/brainSyncLog.js:332` — tail membership proves only a later sequence, not that the tail operation wins the LWW replay.

### Original thread resolution
- Durable floor after a failed append: resolved.
- Idle no-rewrite side-effect assertion: resolved.
- Legacy/asymmetric convergence: partially resolved; floor-0 replay works, but a positive floor can still discard the true LWW winner.

### Coherence check
The live PR description still documents the superseded `else if` / `compactLog(getCurrentSeq())` newest-only design and says 71 tests. The current patch retains one terminal entry per `(type, id)` and the two focused files contain 72 tests. Please update the description and acceptance wording to match the chosen compatibility-preserving bound.

_Generated by /do:review_

prateeekbuilds:
Addressed the latest review feedback:

1. **Pre-floor LWW retention when tail is stale (`server/services/brainSyncLog.js`)**: Updated `compactLog` to compare timestamps against tail operations; if the tail contains only stale/losing entries for a key (e.g. older winner is a Jan-02 delete and tail has an echoed Jan-01 create), the pre-floor winning entry is retained before the verbatim tail so fresh/delta-only peers do not accept the stale create and resurrect the record.
2. **Positive-floor test fixture (`server/services/brainSyncLog.test.js`)**: Added test verifying `compactLog(50)` preserves the pre-floor winning delete before the verbatim tail and serves both on `getChangesSince(0)`.
3. **PR description & test counts**: Updated the PR description to accurately reflect the compatibility-preserving terminal LWW replay design, mutex-guarded durable floor, idle-write assertion, and 73 focused tests.

atomantic (commented):
Reviewed the updated head 76775fe with /do:review — no blocking findings.

The latest commit resolves the positive-floor LWW issue: a pre-floor winner is retained for equal or stale tail timestamps, while any strictly newer tail entry correctly makes that baseline redundant. The earlier durable-floor, idle-write, and legacy/asymmetric compatibility findings also remain resolved, so I resolved their threads and dismissed the superseded change-request reviews.

Non-blocking follow-up: index preserved tail entries by record key during compaction instead of filtering the whole tail once per overlapping key. That will make large positive-floor peer compactions linear and avoid unnecessarily long log-mutex holds; it does not affect the no-peer path fixed here.

The PR description now matches the compatibility-preserving terminal-per-key design and current focused coverage. Required current-head CI is still pending authorization.

_Generated by /do:review_

PR #5561, opened 2026-08-31
fix: mitigate MLX video GPU watchdogs

### LINKED ISSUE #5553: Wire M5 Max GPU-watchdog mitigations into FastVideo/video-gen MLX renders (silent solid-color output, no crash)
## Symptom

Two `fastmetal_14b_qad` renders on this box (Apple **M5 Max**, `Mac17,7`/`T6050`) at `1024x576`, 121 frames both completed with exit code 0, valid mp4/thumbnail, ~870s render time — but every sampled frame across the whole clip is byte-identical solid `RGB(54,45,43)`, independent of prompt/seed. That color is consistent with a VAE decode of a degenerate (zero/garbage) latent — the denoising loop produced no real signal.

A quick, tiny repro (480x272, 9 frames, ~55s total, ~30s of denoise) on the same box with the same cached model weights rendered correctly (2000+ distinct colors, matched the prompt). So the runtime, MLX/Metal GPU detection, and cached weights are all fine — this is not a missing-install problem.

## Root cause (high confidence, not yet proven with telemetry)

This machine already has a documented history of Apple M5 ("brand-new silicon") GPU/Metal driver instability under **sustained** heavy compute — see `docs/research/2026-06-13-mflux-training-watchdog-panic.md` and issue #1329. That incident traced repeated hard-reboot kernel panics during mflux LoRA training to upstream `ml-explore/mlx#3267` ("Metal GPU watchdog kills training when display is active" — IOGPU command-buffer kill under active-display WindowServer GPU contention) and `ml-explore/mlx#3186` (a related IOGPU `completeMemory()` panic, corroborated on M5). The validated mitigations for LoRA training on this exact box were:

1. `AGX_RELAX_CDM_CTXSTORE_TIMEOUT=1` env var before spawning the Metal-heavy child (`scripts/train_mflux_lora.py`)
2. Sleeping the display for the duration of sustained GPU compute (`server/services/loraTraining/displayPower.js`)
3. Segmenting long sustained-GPU work into shorter chunks with full process teardown between segments (`--segment-steps`/cooldown in `train_mflux_lora.py`)

None of those three mitigations are wired into the FastVideo path (`scripts/generate_fastvideo.py` → `~/.portos/fastvideo/examples/inference/basic/mlx_wan_prompt_to_video.py`), nor into the other sustained-Metal MLX video runtimes (`generate_wan22.py`, `generate_minimax_h3.py`, `generate_ltx2.py`). A ~14.5-minute continuous FastMetal-14B denoise+decode is exactly the sustained-load shape that triggered the watchdog contention during training — the working theory is that under lighter/partial contention (vs. the full watchdogd timeout that caused a reboot during training) a GPU command buffer can silently fail and return a corrupted/zeroed buffer that the pipeline decodes and writes out anyway, rather than crashing the process or the box. This would explain why the corruption is deterministic (same constant color regardless of prompt/seed — a driver-level failure, not a numerically-diverging one) and why a different (likely older, more mature-driver) machine with 128GB RAM does not reproduce it — RAM headroom is not the variable, sustained GPU duration + active display is.

## Proposed fix

Port the training path's mitigations to the FastVideo runner, and audit the other sustained MLX video runtimes for the same gap:

1. Set `AGX_RELAX_CDM_CTXSTORE_TIMEOUT=1` in `scripts/generate_fastvideo.py`'s child env before spawning `mlx_wan_prompt_to_video.py` (mirror the pattern already in `train_mflux_lora.py`).
2. Extract/generalize `server/services/loraTraining/displayPower.js`'s display-sleep-for-duration-of-run helper so `server/services/videoGen/generateVideo.js` can call it around any BYOV MLX render (fastvideo, wan22, minimax_h3, ltx2), gated behind a settings toggle mirroring `settings.loraTraining.displaySleep` (default on).
3. Investigate whether the FastVideo entry script can be segmented (checkpoint/resume mid-denoise) the way `train_mflux_lora.py` is; if not feasible upstream, at minimum document the display-off requirement for long FastVideo renders in `docs/TROUBLESHOOTING.md` next to the existing "GPU watchdog kernel panic during LoRA training" entry.
4. Consider a cheap post-render sanity check (e.g. flag when decoded frame variance is ~0 across the whole clip) so a corrupted render surfaces as a visible warning/retry prompt in the Video Gen UI instead of silently looking like a successful render.

## Evidence / repro

- Corrupted renders: `data/video-history.json` entries `8e166f5a-e39b-421d-8e1f-c3029ea849e9` and `aa625837-3d40-4a8d-91c9-53e0ebcafa51` (both `fastmetal_14b_qad`, 1024x576, 121 frames, ~870s render time, both decode to solid `RGB(54,45,43)` at every sampled frame).
- Working tiny repro: same cached snapshot, `--width 480 --height 272 --num-frames 9 --num-inference-steps 3`, ~55s wall time, correct varied-color output.
- Direct microbenchmarks of `mx.quantized_matmul` (affine int8, group_size=64) and `mx.fast.scaled_dot_product_attention`/`mx.fast.rms_norm` at various shapes/dtypes on this box all returned numerically correct results — rules out a blanket M5 kernel-correctness bug for these ops in isolation; the failure needs sustained wall-clock duration to reproduce, matching the watchdog-contention theory rather than a per-op bug.
- Related: `docs/research/2026-06-13-mflux-training-watchdog-panic.md`, issue #1329, upstream `ml-explore/mlx#3267`, `ml-explore/mlx#3186`.

Not filed upstream against FastVideo/mlx — this is a proposed PortOS-side mitigation (matching what was already necessary for the training path on this same hardware class), not a claim that FastVideo/mlx itself is broken.

atomantic:
## Summary
- apply the M5 Metal watchdog environment mitigation to FastVideo
- sleep the display for local MLX video runtimes while keeping the system awake
- add an opt-out in Settings and document the mitigation

## Test plan
- `cd server && npm test -- --run services/loraTraining/displayPower.test.js services/videoGen/runtimes.test.js`
- `cd client && npm test -- --run src/components/settings/ImageGenTab.test.jsx`
- `npx vitest run scripts/generate_fastvideo.test.js`
- `cd client && npm run lint`

Closes #5553

### DIFF
--- a/client/src/components/settings/ImageGenTab.jsx
+++ b/client/src/components/settings/ImageGenTab.jsx
@@ -110,6 +110,7 @@ export function ImageGenTab() {
   // never rendered, only round-tripped at save time so sibling keys
   // (defaultModelId) survive the settings PUT's wholesale slice replace.
   const [videoGenMode, setVideoGenMode] = useState('');
+  const [videoGenDisplaySleep, setVideoGenDisplaySleep] = useState(true);
   const videoGenSliceRef = useRef({});
   const [sdapiUrl, setSdapiUrl] = useState('');
   const [pythonPath, setPythonPath] = useState('');
@@ -174,6 +175,7 @@ export function ImageGenTab() {
     denoiseByMode: { external: false, local: false, codex: false, grok: false, agy: false },
     renderDefaultsJson: '{}',
     videoGenMode: '',
+    videoGenDisplaySleep: true,
   });
 
   const [status, setStatus] = useState(null);
@@ -245,6 +247,7 @@ export function ImageGenTab() {
         // ('auto'/blank → '', i.e. no pin) for the select.
         const vg = (s?.videoGen && typeof s.videoGen === 'object') ? s.videoGen : {};
         const vgMode = normalizeRenderPinValue(vg.mode) || '';
+        const vgDisplaySleep = vg.displaySleep !== false;
         const m = ig.mode || IMAGE_GEN_MODE.EXTERNAL;
         const url = normalizeUrl(ig.external?.sdapiUrl || ig.sdapiUrl);
         const py = ig.local?.pythonPath || '';
@@ -279,6 +282,7 @@ export function ImageGenTab() {
         setMode(m);
         setRenderDefaults(rd);
         setVideoGenMode(vgMode);
+        setVideoGenDisplaySleep(vgDisplaySleep);
         videoGenSliceRef.current = vg;
         setSdapiUrl(url);
         setPythonPath(py);
@@ -306,6 +310,7 @@ export function ImageGenTab() {
           cleanC2PAByMode: c2, denoiseByMode: dn,
           renderDefaultsJson: JSON.stringify(rd),
           videoGenMode: vgMode,
+          videoGenDisplaySleep: vgDisplaySleep,
         });
         setToolRegistered(tools.some((t) => t.id === SDAPI_TOOL_ID));
         setCodexToolRegistered(tools.some((t) => t.id === CODEX_TOOL_ID));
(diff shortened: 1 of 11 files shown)


atomantic:

PR #5626, opened 2026-09-01
fix: dedupe insightsService's transport onto the shared aiProvider one

### LINKED ISSUE #5617: [module-hygiene-insights-api-transport] Reuse aiProvider for Insights LLM calls
**Slice/date:** Server-side AI provider/runner boundary, specifically `server/services/insightsService.js` and its service/test boundary; audited 2026-09-01.

## Problem

`server/services/insightsService.js:166-226` defines and exports a second `callProviderAISimple`; its own JSDoc says it “Replicate[s] callProviderAISimple pattern from taste-questionnaire.js.” The existing owner is `server/services/aiProvider.js:136-195` (the HTTP completion) and `server/services/aiProvider.js:221-336` (provider readiness, status, timeout, model recovery, response classification, and the public helper). Insights invokes its copy at `server/services/insightsService.js:430-433` and `:537-540`.

This is already behaviorally divergent, not just duplicate naming. The shared service rejects a missing/whitespace completion at `server/services/aiProvider.js:172-185`, while the Insights copy accepts a valid JSON response with empty content and returns `{ text: '' }` at `server/services/insightsService.js:212-222`. The cross-domain path then writes that empty result to `narrative.json` at `server/services/insightsService.js:546-556`. The two test suites consequently maintain separate transport contracts: `server/services/insightsService.test.js:21-98` tests the copy, while `server/services/aiProvider.test.js:86-151` tests the shared implementation.

History shows the maintenance cost: `d0a8ebf40` promoted `aiProvider` into `server/services` while this copy remained, and `015e617be` plus `f712182c2` required an Insights-specific non-JSON fix and test. Prior issue #2037 verified that Insights generation is user-triggered; this finding is about the duplicated transport after that trigger contract, not about adding an unsolicited call.

## Impact

A provider response can be classified as a successful empty completion in Insights while the same response is a provider error everywhere using `aiProvider`. Shared fixes for status reporting, provider readiness/model recovery, timeout behavior, or malformed/empty response handling do not automatically reach the two user-triggered Insights generation workflows. The result is different runtime/data behavior and repeated manual transport patches; the cross-domain workflow can persist an empty narrative instead of surfacing the provider failure.

## Reuse search

Searched `server/lib/README.md`, `server/lib/index.js`, `server/lib/aiToolkit/index.js`, and repository imports of `callProviderAISimple`. The existing `server/services/aiProvider.js` is the appropriate owner; no new helper or catalog entry is needed. Prior layering work in #4901/PR #5029 deliberately moved provider orchestration into `server/services` and left only pure text helpers in `server/lib`.

## Fix

Make `server/services/aiProvider.js` the sole owner of the simple API completion transport. Delete the implementation and transport-only imports/constants from `server/services/insightsService.js`, and import `callProviderAISimple` from `./aiProvider.js`. Keep Insights-specific context assembly, JSON/text parsing, cached reads, and persistence in `insightsService.js`; only the provider call crosses the shared seam. Move the non-JSON/empty/HTTP transport matrix into `server/services/aiProvider.test.js`, and keep `server/services/insightsService.test.js` focused on the public Insights workflows and disk-only read paths.

Compatibility obligations: preserve the exported Insights functions, their return shapes, prompt/model/temperature choices, `themes.json` and `narrative.json` paths, and the no-cold-bootstrap read behavior. Update the expected test for the intentional canonical classification of an empty completion; do not change the Insights trigger contract.

## Acceptance criteria

- [ ] `server/services/insightsService.js` contains no second `callProviderAISimple` implementation and no unused transport imports/constants.
- [ ] Both `generateThemeAnalysis` and `refreshCrossDomainNarrative` use the shared `server/services/aiProvider.js` helper.
- [ ] A malformed, non-JSON, non-2xx, or whitespace-only provider response reaches the shared error classification and does not write a successful empty narrative/theme result.
- [ ] Insights cached-read entry points remain disk-only and make no provider call.
- [ ] `server/services/aiProvider.test.js` covers the transport classification, and `server/services/insightsService.test.js` covers the public workflow contract without re-testing the deleted transport copy.

**Scope:** medium

Bryandero98:
## Summary
- `insightsService.js` defined a second `callProviderAISimple`, behaviorally divergent from the canonical `server/services/aiProvider.js` one: it classified a non-JSON/blank 200 body as an error (correct) but a *valid* 200 body with **empty content** as a successful `{ text: '' }` — the opposite of `aiProvider`'s classification, which treats a whitespace-only completion as a provider error.
- Both `generateThemeAnalysis` and `refreshCrossDomainNarrative` persist their result unconditionally once `result.error` is unset, so the divergent copy could overwrite a real cached narrative/theme with nothing on an otherwise-valid-but-content-empty response.
- Deletes the duplicate and routes both entry points through the shared `aiProvider.callProviderAISimple` — same call shape, drop-in compatible (also incidentally fixes a smaller divergence: the deleted copy called `ensureOllamaProviderReady` for *any* API-type provider, not just Ollama-backed ones, unlike the shared implementation's `isOllamaProvider` gate).
- Moves the transport/malformed-body/SSRF-guard test matrix onto `aiProvider.test.js` (where the shared implementation now lives, and where its whitespace-only-completion test already lived). `insightsService.test.js` keeps only the disk-only-read-path contract, matching its own docstring's stated scope.

## Test plan
- [x] `npx vitest run services/insightsService.test.js services/aiProvider.test.js` — 24 tests pass
- [x] Confirmed no other file destructure-imports the deleted `callProviderAISimple` export from `insightsService.js` (the route file uses `import * as insightsService`, unaffected)
- [x] Repo-wide grep for any other file importing from `insightsService.js` — only the route file and its own test

Fixes #5617

### DIFF
--- a/server/services/aiProvider.test.js
+++ b/server/services/aiProvider.test.js
@@ -363,3 +363,94 @@ describe('callProviderAISimple through the ChatGPT subscription', () => {
     expect(result).toMatchObject({ text: 'fallback answer' });
   });
 });
+
+// The regression these two guard: a non-JSON/blank 200 body used to return
+// { text: '' } from the (now-deleted) insightsService copy of this transport,
+// which refreshCrossDomainNarrative / generateThemeAnalysis then persisted over
+// narrative.json / themes.json — overwriting the cached result with nothing.
+// The shared transport must surface an error instead, so the `if (result.error)
+// return` guard at both call sites bails before any write.
+describe('callProviderAISimple — malformed / non-2xx responses', () => {
+  const provider = { id: 'provider-1', name: 'Example Provider', type: 'api', endpoint: 'https://api.example.com/v1' };
+
+  const respondWithText = (text, { ok = true, status = 200 } = {}) =>
+    vi.stubGlobal('fetch', vi.fn(async () => ({ ok, status, text: async () => text })));
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it('returns an error, not an empty success, on a non-JSON 200 body', async () => {
+    respondWithText('<html><body>502 Bad Gateway</body></html>');
+
+    const result = await callProviderAISimple(provider, 'model-1', 'prompt');
+
+    expect(result.text).toBeUndefined();
+    expect(result.error).toMatch(/malformed/i);
+  });
+
+  it('returns an error on a blank 200 body', async () => {
+    respondWithText('');
+
+    const result = await callProviderAISimple(provider, 'model-1', 'prompt');
+
+    expect(result.error).toMatch(/malformed/i);
+  });
+
+  it('returns an error with the status and body on a non-2xx response', async () => {
+    respondWithText('boom', { ok: false, status: 500 });
+
+    const result = await callProviderAISimple(provider, 'model-1', 'prompt');
+
+    expect(result.error).toMatch(/Provider returned 500: boom/);
+  });
+});
+
+describe('callProviderAISimple — endpoint guard (SSRF / key-exfiltration)', () => {
+  const provider = { id: 'provider-1', name: 'Example Provider', type: 'api', endpoint: 'https://api.example.com/v1' };
+
+  const respondWithHello = () => vi.stubGlobal('fetch', vi.fn(async () => ({
+    ok: true,
+    status: 200,
+    text: async () => JSON.stringify({ choices: [{ message: { content: 'hello' } }] }),
+  })));
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it('blocks a keyed provider pointed at a non-allowlisted endpoint and never calls fetch', async () => {
+    const fetchSpy = vi.fn();
+    vi.stubGlobal('fetch', fetchSpy);
+
+    const result = await callProviderAISimple(
+      { ...provider, apiKey: 'secret-key', endpoint: 'https://not-an-allowlisted-host.example' },
+      'model-1', 'prompt',
+    );
+
+    expect(result.error).toContain('Provider endpoint blocked');
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
+
+  it('allows a keyed provider on a non-allowlisted host through when allowCustomEndpoint is true', async () => {
+    respondWithHello();
+
+    const result = await callProviderAISimple(
+      { ...provider, apiKey: 'secret-key', endpoint: 'https://not-an-allowlisted-host.example', allowCustomEndpoint: true },
+      'model-1', 'prompt',
+    );
+
+    expect(result).toMatchObject({ text: 'hello' });
+  });
+
+  it('allows a keyless provider on a non-allowlisted host through (guard only applies when apiKey is set)', async () => {
+    respondWithHello();
+
+    const result = await callProviderAISimple(
+      { ...provider, endpoint: 'https://not-an-allowlisted-host.example' },
+      'model-1', 'prompt',
+    );
+
+    expect(result).toMatchObject({ text: 'hello' });
+  });
+});
--- a/server/services/insightsService.test.js
+++ b/server/services/insightsService.test.js
@@ -1,101 +1,21 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
-// Mock the provider-readiness check and the HTTP client so the call runs offline.
+// Mock the provider-readiness check so any code path that touches it runs offline.
 vi.mock('./ollamaManager.js', () => ({
   ensureProviderReady: vi.fn().mockResolvedValue({ success: true }),
 }));
-vi.mock('../lib/fetchWithTimeout.js', () => ({
-  fetchWithTimeout: vi.fn(),
-}));
-
-import { fetchWithTimeout } from '../lib/fetchWithTimeout.js';
-import { mockJsonResponse, mockTextResponse } from '../lib/testHelper.js';
-import {
-  callProviderAISimple,
-  getThemeAnalysis,
-  getCrossDomainNarrative,
-} from './insightsService.js';
-
-const PROVIDER = { type: 'api', endpoint: 'http://localhost:1234/v1' };
-
-describe('insightsService.callProviderAISimple — non-JSON-body guard', () => {
-  beforeEach(() => {
-    vi.clearAllMocks();
-  });
-
-  // The regression this guards: a non-JSON 200 body used to return { text: '' },
-  // which refreshCrossDomainNarrative / generateThemeAnalysis then persisted over
-  // narrative.json / themes.json — overwriting the cached result with nothing.
-  // It must surface as { error } so the `if (result.error) return` guard bails
-  // before any write.
-  it('returns { error } (not empty success) on a non-JSON 200 body', async () => {
-    fetchWithTimeout.mockResolvedValue(mockTextResponse('<html><body>502 Bad Gateway</body></html>'));
-    const result = await callProviderAISimple(PROVIDER, 'm', 'prompt');
-    expect(result.text).toBeUndefined();
-    expect(result.error).toMatch(/non-JSON response/);
-  });
-
-  it('returns { error } on a blank 200 body', async () => {
-    fetchWithTimeout.mockResolvedValue(mockTextResponse(''));
-    const result = await callProviderAISimple(PROVIDER, 'm', 'prompt');
-    expect(result.error).toMatch(/non-JSON response/);
-  });
-
-  // A valid body with empty content is a legitimate (if unusual) result and must
-  // still flow through as { text: '' } — the guard must not conflate valid-empty
-  // with a parse failure.
-  it('returns { text: "" } for a valid body with empty content', async () => {
-    fetchWithTimeout.mockResolvedValue(mockJsonResponse({ choices: [{ message: { content: '' } }] }));
-    const result = await callProviderAISimple(PROVIDER, 'm', 'prompt');
-    expect(result).toEqual({ text: '' });
-  });
-
-  it('returns the content for a valid populated body', async () => {
-    fetchWithTimeout.mockResolvedValue(mockJsonResponse({ choices: [{ message: { content: 'hello' } }] }));
-    const result = await callProviderAISimple(PROVIDER, 'm', 'prompt');
-    expect(result).toEqual({ text: 'hello' });
-  });
-
-  it('returns { error } with the status code on a non-2xx response', async () => {
-    fetchWithTimeout.mockResolvedValue(mockTextResponse('boom', { ok: false, status: 500 }));
-    const result = await callProviderAISimple(PROVIDER, 'm', 'prompt');
-    expect(result.error).toMatch(/Provider returned 500: boom/);
-  });
+// generateThemeAnalysis/refreshCrossDomainNarrative now call the shared
+// aiProvider.callProviderAISimple transport (see aiProvider.test.js for its own
+// contract coverage) — stub only that export so the disk-only read paths below
+// can assert it's never reached, while stripCodeFences/parseLLMJSON (also
+// imported from this module) stay real.
+vi.mock('./aiProvider.js', async (importOriginal) => {
+  const actual = await importOriginal();
+  return { ...actual, callProviderAISimple: vi.fn() };
 });
 
-describe('insightsService.callProviderAISimple — endpoint guard (SSRF / key-exfiltration)', () => {
-  beforeEach(() => {
-    vi.clearAllMocks();
-  });
-
-  it('blocks a keyed provider pointed at a non-allowlisted endpoint and never calls fetch', async () => {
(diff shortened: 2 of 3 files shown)


Bryandero98:
Appreciate the detailed writeup on this one — pinning the exact behavioral divergence (line-referenced, both sides) made it straightforward to confirm which classification was canonical instead of guessing. Let me know if you'd rather see the moved test coverage organized differently.

server/services/aiProvider.test.js
```diff
@@ -150,3 +150,94 @@ describe('callProviderAISimple completion classification', () => {
     expect(statusOp.error).not.toHaveBeenCalled();
   });
 });
+
+// The regression these two guard: a non-JSON/blank 200 body used to return
+// { text: '' } from the (now-deleted) insightsService copy of this transport,
+// which refreshCrossDomainNarrative / generateThemeAnalysis then persisted over
+// narrative.json / themes.json — overwriting the cached result with nothing.
+// The shared transport must surface an error instead, so the `if (result.error)
+// return` guard at both call sites bails before any write.
```
atomantic, line 372: **[IMPROVEMENT]** This comment is accurate about *why* the tests exist, but the consolidation also **flips a contract in the opposite direction**, and neither the comment nor the PR description says so.

The deleted `insightsService` copy ended with:

```js
return { text: data.choices?.[0]?.message?.content || '' };
```

so a **valid** 200 body whose `content` was empty returned `{ text: '' }` — a *success*. The deleted suite pinned that on purpose: *"A valid body with empty content is a legitimate (if unusual) result and must still flow through as `{ text: '' }` — the guard must not conflate valid-empty with a parse failure."* That was written against AGENTS.md's sentinel rule (absent vs. intentionally-empty).

The shared transport decides the opposite:

```js
if (!content.trim()) {
  return { error: 'Provider returned an empty completion' };
}
```

**I think the new behavior is the right one**, and it actually finishes the bug your comment describes: under the old code an empty-content 200 sailed through as `{ text: '' }` and *did* get persisted over `narrative.json` / `themes.json`; now `if (result.error) return` at both call sites bails and the cache survives. I also confirmed the contract is already pinned on `main` (`aiProvider.test.js` — "classifies a whitespace-only completion as a provider error"), so no coverage was lost by dropping the old case.

The ask is just honesty in the record: the PR title and body read as a pure dedupe, but this is a **deliberate semantics change**. Please say so in the description — a reader hitting `git log` later needs to know the empty-completion verdict moved. Same for two smaller behavior deltas worth a line: the non-API rejection message changes (`Insights analysis requires an API-based provider` → `This operation requires an API-based provider`), and Insights can now dispatch to a Codex subscription-text provider, which it previously refused.

server/services/insightsService.test.js
```diff
@@ -109,13 +29,13 @@ describe('insightsService read paths — disk-only, no provider call', () => {
 
   it('getThemeAnalysis performs no provider call (returns not_generated when uncached)', async () => {
     const result = await getThemeAnalysis();
-    expect(fetchWithTimeout).not.toHaveBeenCalled();
+    expect(callProviderAISimple).not.toHaveBeenCalled();
```
atomantic, line 32: **[IMPROVEMENT]** This weakens the assertion these two tests exist for. The block comment above them calls out the **no-cold-bootstrap contract** (AGENTS.md: *"PortOS must never queue up AI provider calls a user hasn't knowingly triggered"*), and `fetchWithTimeout` was the right altitude to pin it: it caught *any* outbound provider call from the disk-only read paths, whatever transport got there.

`callProviderAISimple` is one export. A future read path that reaches a provider by some other route — a direct `fetch`, a streaming helper, a second transport — now slips past both tests while they stay green. That is exactly the regression class this file is guarding.

Keeping a transport-level assertion alongside costs two lines and restores the original strength (applies to line 38 as well):

```suggestion
    expect(callProviderAISimple).not.toHaveBeenCalled();
    expect(fetchSpy).not.toHaveBeenCalled();
```

with `const fetchSpy = vi.fn(); vi.stubGlobal('fetch', fetchSpy);` in the `beforeEach` and `vi.unstubAllGlobals()` in an `afterEach` — mirroring the `stubGlobal` pattern you already use in `aiProvider.test.js`.

atomantic (commented):
Reviewed by `/do:review` (security-focused pass, external contributor) — **0 critical, 2 improvements**. This one is good; **rebase and it's mergeable**. Both findings are non-blocking and we're happy to take them in a follow-up if you'd rather not spin another round.

**Rebase required.** The branch is **188 commits behind `main`** and currently `CONFLICTING`. In this repo a conflicted PR gets **no CI at all** ("no checks reported"), so nothing here has been pipeline-verified. The single conflict is an append-at-EOF collision in `aiProvider.test.js` (`main` added a Codex-subscription block at the same spot) — keep both sides, no semantic overlap.

**Verified locally** (isolated worktree, rebased onto current `main`, conflict resolved by keeping both blocks):
- Full `cd server && npm test` → **1809 files / 37248 tests pass**, exit 0.
- **Transport parity checked field by field**, since a dedupe is only safe if the survivor is equivalent:
  - Timeout: identical — both `provider.timeout || 300000`.
  - Endpoint guard: preserved exactly, and now enforced inside `postChatCompletion`, so the SSRF / key-exfiltration policy, the `allowCustomEndpoint` opt-in, and keyless local-LLM passthrough all behave the same. Your three new guard tests pin it at the new home.
  - Non-2xx: identical message shape (`Provider returned {status}: {body}`).
  - Ollama readiness: **improved** — the old copy called `ensureProviderReady` for *every* `type === 'api'` provider; the shared one scopes it to `isOllamaProvider`, which is what it always meant.
  - Empty-content 200: **changed** — see the inline comment; net improvement, but it needs to be in the description.
- Security scan: pure ASCII apart from em-dashes, no network calls, no `process.env` reads, no `eval`/`exec`/`child_process`, no dependency or CI changes, no new secrets path. Single commit, single author. Net **-59 lines** and one fewer copy of a keyed-request transport, which is a genuine security win on its own — one place to audit instead of two.

**The two notes:**
1. The empty-completion verdict flips from `{ text: '' }` success to `{ error }`. Good change (it's what finally stops an empty narrative overwriting the cache) and already pinned on `main`, but the PR reads as a pure dedupe and doesn't mention it.
2. The read-path no-cold-bootstrap tests drop from a `fetchWithTimeout`-level assertion to a `callProviderAISimple`-level one, which a future non-`callProviderAISimple` path would slip past.

**Coherence check:** description matches the diff otherwise, and the test plan holds up — I re-ran it. Deleting the duplicated transport rather than wrapping it is the right call, and moving its contract tests to live beside the surviving implementation is exactly where they belong.

_Generated by /do:review_

Bryandero98:
Rebased onto current \`main\` (single conflict, exactly as flagged — an append-at-EOF collision in \`aiProvider.test.js\` where \`main\`'s new Codex-subscription test block and this PR's malformed-response test block both landed at the same spot). Kept both in full: worth noting for anyone re-resolving a similar case, git's merge deduplicated the closing \`});\`/\`});\` lines it judged identical between the two sides and left them out of the conflict markers entirely — had to manually restore them after the auto-resolve to close \`main\`'s own \`describe\` block, or the file would have silently mis-nested the two blocks into one.

Re: the empty-completion-verdict note — already called out explicitly in this PR's own description (first bullet: "a *valid* 200 body with **empty content**..."), so no separate edit needed there.

Verified: \`npx vitest run services/aiProvider.test.js services/insightsService.test.js\` — 36/36 pass (up from the 24 in the original test plan, since \`main\` picked up new Codex-transport tests in the meantime).

server/services/aiProvider.test.js
atomantic: You're right and I was wrong to ask — the description's first bullet already states it explicitly ("a *valid* 200 body with **empty content** as a successful `{ text: '' }` — the opposite of `aiProvider`'s classification"), and the commit body carries the same sentence. That's the record I wanted. Resolving.

The two smaller deltas I mentioned (the non-API rejection message text, and Insights now being able to dispatch to a Codex subscription-text provider where it previously refused) are both fallout of adopting the canonical transport rather than anything this PR decided, and both are improvements. I'll note them on the merge rather than ask you to spin another round for a description edit.

Also confirmed your rebase resolution independently: `aiProvider.test.js` at `4cc7525` is byte-identical to `main`'s copy plus your two appended `describe` blocks, so the `});` restoration you described landed correctly and nothing is mis-nested. Good catch on that one — a silently-swallowed closing brace there would have folded `main`'s Codex block into yours and quietly changed which hooks applied to it.

server/services/insightsService.test.js
atomantic: Leaving this one to us — it's a test-strength point on our side of the fence, not a defect in your change, and it isn't worth another round-trip on a PR that's otherwise ready. Merging as-is; we'll open our own follow-up that restores a transport-level assertion here.

For the record of why it still matters: `callProviderAISimple` is one export, and the contract this block guards is "no AI provider call from a read path, by *any* route." A future read path that reached a provider through a different transport would slip past both assertions while they stay green. The follow-up will re-add the `fetch`-level spy alongside the export-level one, and probably also add the generation-side test neither file has today — that an errored transport result leaves `themes.json` / `narrative.json` untouched, which is the actual bug #5617 describes and is currently only proven transitively through the `if (result.error) return` guard.

atomantic (approved):
Second `/do:review` pass (security-focused, external contributor) — **0 findings. Approving and merging.** Both threads from the first round are answered and resolved.

**What changed since round one:** rebased onto current `main`, conflict resolved, and CI is now actually running — `CI Gate` **pass**, `Server tests` **pass**, `Plan test impact` **pass**. `mergeStateStatus: CLEAN`.

**Re-verified independently on `4cc7525`:**
- **Rebase resolution is correct.** `aiProvider.test.js` at HEAD is byte-identical to `main`'s copy plus a pure append of the two new `describe` blocks — no mis-nesting, no swallowed brace, `main`'s Codex-subscription block intact. This was the one thing worth checking twice, since a folded `describe` would have silently changed hook scoping while staying green.
- **Production diff is a pure deletion.** `insightsService.js` loses only the duplicate function and its now-unused imports/constant; grepped the file for every removed symbol (`fetchWithTimeout`, `readResponseJson`, `ensureOllamaProviderReady`, `evaluateSecretEndpoint`, `DEFAULT_AI_TIMEOUT_MS`) — zero residual references. Only `server/routes/insights.js` consumes this module, via `import * as`, so the dropped named export breaks nothing.
- **Signature parity holds.** Shared `callProviderAISimple(provider, model, prompt, options)` carries the same `temperature = 0.3` / `max_tokens = 1000` defaults and the same `provider.timeout || 300000`; both call sites pass their own overrides and gate on `result.error` before `atomicWrite`, so the cache-overwrite path #5617 describes is closed.
- **Security scan, adversarial pass:** no non-ASCII beyond em-dashes, no bidi or zero-width characters, no new `process.env` / `child_process` / `eval` / network calls, no dependency, lockfile, or CI changes. Net **−59 lines** and one fewer copy of a keyed-request transport — one place to audit instead of two, which is the real security win here.
- **The SSRF / key-exfiltration guard is preserved and now better covered.** It lives in `postChatCompletion`, and the three appended guard tests pin all three arms at the surviving implementation: blocked-when-keyed, allowed under `allowCustomEndpoint`, and keyless-passthrough. None of them weakens a guard — they pin behavior that was previously asserted only against the deleted copy, so the consolidation would otherwise have *lost* that coverage.
- **New transport tests are complementary, not duplicates** — `main` already covered whitespace-only and the happy path; this adds non-JSON body, blank body, and non-2xx.

**Against the ticket (#5617):** all five acceptance criteria met. The one soft spot is AC 3 — that an errored transport leaves `themes.json` / `narrative.json` untouched is proven transitively (transport errors + `if (result.error) return`) rather than by a test at the Insights boundary. That's pre-existing, not a regression, and we're taking it in our own follow-up along with the read-path assertion altitude.

**Behavior deltas worth recording**, all fallout of adopting the canonical transport rather than choices this PR made:
1. An empty/whitespace-only completion flips from `{ text: '' }` success to `{ error }` — stated in the description and commit body, and it is what finally stops an empty narrative overwriting the cache.
2. Non-API rejection message text changes (`Insights analysis requires an API-based provider` → `This operation requires an API-based provider`).
3. Ollama readiness is now scoped by `isOllamaProvider` instead of firing for every `type === 'api'` provider — the old copy's behavior was a bug.
4. Insights can now dispatch to a Codex subscription-text provider, which it previously refused outright.

Thanks — deleting the duplicate rather than wrapping it, and relocating its contract tests to sit beside the surviving implementation, is exactly the right shape for this.

atomantic:
Merged as `7bc373f` — thanks, this was a clean one. Issue #5617 closed automatically.

The read-path assertion altitude I raised in the second thread is now our follow-up: #5845. It restores the wire-level `fetch` spy alongside the export-level one, and adds the test neither file had — that an errored transport leaves `themes.json` / `narrative.json` untouched, which is the actual bug #5617 describes and was previously only proven transitively through the `if (result.error) return` guard.

Issue #5596, opened 2026-09-01
Grow the user-action allowlist and detect leftover branches after idle agent runs


atomantic:
Part of #5593.
Depends on: Phase 1 and Phase 2 merged, plus a live install that has recorded operator events.

## Problem / Goal

Phases 1–2 log a **starter** allowlist and let the mind/review task read it. They will not discover every tedious operator loop — that is the point of evaluating on a live install first. This phase (1) extends the closed vocabulary from **observed** gaps, and (2) ships the first concrete detector the motivating example needs: leftover local branches after CoS agents go idle, so the user stops noticing-and-clicking `branch-reconcile` by hand.

## Context

Blocked until:

- Phase 1's `user_action_events` is in `main` and a live install has at least several days of `cos.task.*` / `cos.schedule.trigger` / `settings.update` rows.
- Phase 2's query tool + `user-action-review` exist so new event types are immediately visible to the mind.

Do **not** start from a guessed 40-type taxonomy. Read the live log (`GET /api/user-actions?from=<14d>` and `GET /api/user-actions/types`) plus the user's remaining manual loops. The starter gaps already identified in planning, to implement **unless the live log shows they are unused**:

1. **Event-only creative/Brain pointers** (no prompt/body text): `media.image.enqueue`, `media.video.enqueue`, `brain.capture` — `target` = record/job id, summary = "enqueued image job" / "captured thought". Pointers only, matching the epic's "not every image/video prompt text or brain entry body."
2. **Instance feature toggles** — `PUT /api/settings/features/:featureId` (`server/routes/settings.js:201`) is a distinct write path that phase 1 deliberately skipped, and it does not funnel through the settings `save()` diff the way `PUT /` does. Log `instance-feature.toggle` with `{ id, enabled }`, resolved against `server/lib/instanceFeatureRegistry.js` / `server/services/instanceFeatures.js`.
3. **CoS config writes** that are not `PUT /api/settings` — Improve master toggle and per-app task overrides on `server/routes/cosScheduleRoutes.js`. Add `cos.schedule.update` with changed keys only.
4. **Leftover-branch idle detector** (the motivating automation).

[...]

Issue #5594, opened 2026-09-01
Add a machine-local user_action_events store and instrument core CoS, schedule, and settings operator actions


atomantic:
Part of #5593.

## Problem / Goal

Create the durable, filterable operator-action ledger and start writing the high-signal events the rest of the epic depends on. After this issue, we can query "user created this CoS task / rated this run / hit Run Now on branch-reconcile / changed these settings" without involving Persistent Mind yet.

This phase exists so we can **evaluate the log on a live install** before teaching the mind to read it.

## Context

Parent epic records why existing logs (`history.jsonl`, `human_activity_events`, CoS JSONL ledgers, `record_audit`) are the wrong home.

Where each piece goes (all paths verified against `main`):

- **DDL module** — new `server/lib/db/schema/userActions.js` exporting `userActionsDdl`. Add it to the `export { … }` block in `server/lib/db/schema/index.js` and **append it at the END of `buildUpgradeDdl()`** (after `commissionsDdl`). That file's header says the composed order is load-bearing; appending at the tail adds a table with no FK dependencies and disturbs nothing. Add a row to `server/lib/db/schema/README.md`.
- **Fresh-install mirror** — the same `CREATE TABLE` + indexes in `server/scripts/init-db.sql` (`server/lib/db.catalogDdlParity.test.js` has an "every table in init-db.sql is also created in db.js ensureSchema" assertion at L430).
- **Migration stub** — `scripts/migrations/328-user-action-events-table.js` (current max on disk is 327). DDL is boot-time `IF NOT EXISTS`, same shape as `scripts/migrations/161-human-activity-events-table.js`.
- **Parity test** — add a dedicated `it('user_action_events has the same columns and indexes in both files')` to `server/lib/db.catalogDdlParity.test.js`, using `extractCreateTable` + `extractIndexNames(…, 'idx_user_action_')`. Do not skip this: `human_activity_events` has no dedicated `it()` today and that is a gap to not repeat. (`DB_JS` in that test globs `server/lib/db/schema/*.js`, so the new module is picked up automatically.)
[...]

Issue #5593, opened 2026-09-01
Add a queryable user-action log so Persistent Mind can observe operator work and propose automations


atomantic:
## Problem / Goal

PortOS has no LLM-friendly ledger of **what the human operator did in the app**. Persistent Mind can see its own trajectory (`data/cos/mind-events.jsonl`), CoS agent process events (`data/cos/run-events.jsonl`), and (if granted) Brain/goals/calendar tools — but it cannot see that the user just created a CoS task, rated a run, hit Run Now on `branch-reconcile`, or changed settings.

That gap is why leftover-branch cleanup stays manual: after coding agents wrap, the user notices stray branches and triggers `branch-reconcile` by hand. The mind cannot detect the pattern, so it cannot become proactive. The same class of tedium exists across CoS, schedule, and settings work.

Goal: a **phased, queryable, machine-local user-action log** (not PM2 stdout, not the life-timeline, not every keystroke) that Persistent Mind can review, then an opt-in scheduled/MCP path that turns observed repetition into proposed automations.

## Context

**This is not any of the existing logs.** Do not extend them in place:

| Existing store | What it actually is | Why it is the wrong home |
|---|---|---|
| `data/history.jsonl` via `server/services/history.js` (`logAction` at L47, `MAX_ENTRIES = 500` at L8) | App/ops audit: PM2 start/stop/build (`server/routes/apps/lifecycle.js`), allowlisted shell (`server/services/commands.js`), platform-account / personality / automation-schedule CRUD | Wrong shape, 500-entry cap, in-memory filters. `docs/STORAGE.md` L91 already lists it as a remaining Postgres-first target, not a mind feed. |
| `human_activity_events` (`server/lib/db/schema/humanActivity.js`, `server/services/humanActivity.js`) | Life timeline (iMessage, Spotify, YouTube, calendar) for Digital Twin / Tribe / daily-log | Different domain. Closed epic #2149. |
| `data/cos/run-events.jsonl` / `mind-events.jsonl` (`server/services/agentRunEventLog.js`) | Agent/mind **process** telemetry (ephemeral-file, redacted) | Answers "how did this run evolve?", not "what did the user just do in the UI". |
| `record_audit` (`server/lib/db/schema/audit.js`) | Hard-delete / tombstone forensics | Not user actions. |
| `data/cos/decisions.json` | Why CoS skipped/selected a scheduled task | Scheduler transparency, not operator actions. |
[...]

PR #5627, opened 2026-09-01
test: guard user-action-events file-backend writes against leaking to real data/

### LINKED ISSUE #5605: Guard against user_action_events file-backend writes leaking into the live data/ tree during tests
## Problem / Goal

`server/services/userActions.js` (shipped in #5594) writes through `createPgFileFacade`, which selects the **file backend** under `NODE_ENV=test`. Any suite that exercises an instrumented route without redirecting `PATHS.data` to a temp root therefore writes `data/user-action-events.json` into the developer's live `data/` tree — the same bug class as #3683 / #3687.

#5594 patched the three suites that tripped on it (`server/routes/cos.test.js`, `server/routes/cosTaskRoutes.test.js`, `server/services/cosAgentFeedback.test.js`) by adding the temp-root redirect. That is a per-suite fix, not a guard: **the next hook added to a route an untethered suite exercises silently re-opens the hole**, and epic #5593 phase 3 (#5596) is explicitly about growing the hook allowlist.

## Proposed approach

Add a structural guard so a leaked write fails the suite instead of landing in `data/`. Options, in rough order of preference:

1. A global test-setup assertion that fails when `PATHS.data` still resolves to the repo's real `data/` directory while a persisted-store write is attempted.
2. A repo-hygiene test that fails when `data/user-action-events.json` (or any other file-backend artifact) exists after a suite run.
3. A shared test helper that every suite touching an instrumented route must use, enforced by a source-grep guard.

Pick one and apply it; do not ship all three.

## Acceptance criteria

- [ ] Adding a `recordUserAction` hook to a route covered by a suite that has not redirected `PATHS.data` fails that suite, rather than writing into the repo's `data/` tree.
- [ ] A full `cd server && npm test` run leaves no new files under `data/`.
- [ ] The guard names the offending suite and the fix (redirect `PATHS.data` to a temp root) in its failure message.

## Context

- Store: `server/services/userActions.js`, `server/services/userActionsDb.js` (#5594).
- Backend selection: `server/lib/pgFileFacade.js` (`resolvePgBackend` / `isFileBackend`).
- Prior art for repo-hygiene style guards: the tracked-file-scoped hygiene tests already in the server suite.

Related to #5594. Relevant to #5596, which grows the hook allowlist.

Bryandero98:
## Summary
- `server/services/userActions.js` writes through `createPgFileFacade`, which selects the file backend under `NODE_ENV=test`. Any suite that exercises an instrumented route (`recordUserAction`) without redirecting `PATHS.data` to a temp root writes `data/user-action-events.json` into the developer's live `data/` tree — the same bug class as #3683/#3687.
- #5594 patched the three suites that tripped on it, but that's a per-suite fix: the next hook added to a route an untethered suite exercises silently re-opens the hole (epic #5593 phase 3/#5596 is explicitly about growing that allowlist).
- Adds a structural guard directly on the write path (option 1 from the issue): right before the file-backend's `atomicWrite`, compares the live `PATHS.data` against the real repo `data/` dir — computed independently via the same `fileURLToPath`/`resolveInstallRoot` technique `lib/paths.js` itself uses, so a suite's `PATHS` mock can't spoof the comparison too. A write attempted against the unredirected real path now throws a clear, actionable error (naming the fix and pointing at `lib/mockPathsDataRoot.js`) instead of landing on disk.
- Scoped to the `recordUserAction` write chokepoint specifically (not a blanket wrap of `atomicWrite`, which is used by dozens of unrelated stores repo-wide) — every route that ever attaches a `recordUserAction` hook funnels through this one function regardless of which suite exercises it, so growing the hook allowlist (#5596) stays covered automatically.

## Test plan
- [x] New `userActionsDataRootGuard.test.js` proves the guard fires — deliberately does NOT mock `fileUtils.js` (the one test meant to run unredirected), asserts `recordUserAction` throws before any write, and asserts no file lands in the real `data/` tree
- [x] `userActions.test.js` (own suite, already correctly redirected) — 13/13 pass, guard stays silent
- [x] Every known `recordUserAction` caller + its test suite — `routes/cos.test.js`, `routes/cosTaskRoutes.test.js`, `services/cos.test.js`, `services/cosAgentFeedback.test.js`, `routes/settings.test.js`, `services/settings.test.js`, `services/taskSchedule.test.js` — 655/655 pass
- [x] Manually confirmed no `data/user-action-events.json` existed before or after the guard test run

Fixes #5605

### DIFF
--- a/server/services/userActions.js
+++ b/server/services/userActions.js
@@ -40,15 +40,14 @@
  * fire-and-forget layer.
  */
 
-import { dirname, join } from 'path';
-import { fileURLToPath } from 'url';
+import { join } from 'path';
 import { v4 as uuidv4 } from '../lib/uuid.js';
 import { atomicWrite, ensureDir, PATHS, readJSONFile } from '../lib/fileUtils.js';
 import { createFileWriteQueue } from '../lib/fileWriteQueue.js';
 import { isPlainObject } from '../lib/objects.js';
 import { createPgFileFacade, resolvePgBackend } from '../lib/pgFileFacade.js';
 import { isTestRunner } from '../lib/db.js';
-import { resolveInstallRoot } from '../lib/dataRoot.js';
+import { resolveCodeRootForModule, resolveInstallRoot } from '../lib/dataRoot.js';
 import { isUserActionActor, isUserActionType } from '../lib/userActionTypes.js';
 import { insertUserActionEvent, listUserActionEvents, pruneUserActionEvents } from './userActionsDb.js';
 
@@ -58,14 +57,12 @@ import { insertUserActionEvent, listUserActionEvents, pruneUserActionEvents } fr
 const eventsFile = () => join(PATHS.data, 'user-action-events.json');
 
 // The REAL repo data/ dir, computed independently of the (possibly test-mocked)
-// `PATHS` import above via the same fileURLToPath/resolveInstallRoot technique
-// lib/paths.js itself uses — so a suite that redirects PATHS.data to a temp
-// root can't accidentally spoof this comparison too. `dataRoot.js` reads its
-// own env var directly rather than through anything a PATHS mock would touch.
-const REAL_REPO_DATA_DIR = join(
-  resolveInstallRoot(join(dirname(fileURLToPath(import.meta.url)), '../..')),
-  'data',
-);
+// `PATHS` import above via the same resolveCodeRootForModule/resolveInstallRoot
+// technique lib/paths.js itself uses (both go through the shared helper, so
+// they can't silently drift apart) — so a suite that redirects PATHS.data to a
+// temp root can't accidentally spoof this comparison too. `dataRoot.js` reads
+// its own env var directly rather than through anything a PATHS mock would touch.
+const REAL_REPO_DATA_DIR = join(resolveInstallRoot(resolveCodeRootForModule(import.meta.url)), 'data');
 
 /**
  * Structural guard against the bug class in #3683/#3687/#5605: a suite that
@@ -317,9 +314,15 @@ function makeFileBackend() {
   return {
     name: 'file',
     record: (event) => queueWrite(async () => {
+      // Hoisted above loadFileEvents()/the dedupe check on purpose (#5627
+      // review): an un-redirected suite replaying an existing
+      // (type, dedupeKey) used to hit the dedupe short-circuit's `return null`
+      // BEFORE this guard ever ran, silently no-op'ing past it with no throw —
+      // and by then loadFileEvents() had already read the real ledger into the
+      // test process regardless. Running the guard first closes both holes.
+      assertTestDataRootRedirected();
       const events = await loadFileEvents();
       if (events.some((row) => row.type === event.type && row.dedupeKey === event.dedupeKey)) return null;
-      assertTestDataRootRedirected();
       await ensureDir(PATHS.data);
       await atomicWrite(eventsFile(), { events: pruneEvents([...events, event]) });
       return event;
--- a/server/services/userActions.js
+++ b/server/services/userActions.js
@@ -40,13 +40,15 @@
  * fire-and-forget layer.
  */
 
-import { join } from 'path';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
 import { v4 as uuidv4 } from '../lib/uuid.js';
 import { atomicWrite, ensureDir, PATHS, readJSONFile } from '../lib/fileUtils.js';
 import { createFileWriteQueue } from '../lib/fileWriteQueue.js';
 import { isPlainObject } from '../lib/objects.js';
 import { createPgFileFacade, resolvePgBackend } from '../lib/pgFileFacade.js';
 import { isTestRunner } from '../lib/db.js';
+import { resolveInstallRoot } from '../lib/dataRoot.js';
 import { isUserActionActor, isUserActionType } from '../lib/userActionTypes.js';
 import { insertUserActionEvent, listUserActionEvents, pruneUserActionEvents } from './userActionsDb.js';
 
@@ -54,6 +56,40 @@ import { insertUserActionEvent, listUserActionEvents, pruneUserActionEvents } fr
 // test proxy (lib/mockPathsDataRoot.js), and a load-time join would both bind the
 // pre-mock value and crash any suite whose fileUtils stub omits PATHS.data.
 const eventsFile = () => join(PATHS.data, 'user-action-events.json');
+
+// The REAL repo data/ dir, computed independently of the (possibly test-mocked)
+// `PATHS` import above via the same fileURLToPath/resolveInstallRoot technique
+// lib/paths.js itself uses — so a suite that redirects PATHS.data to a temp
+// root can't accidentally spoof this comparison too. `dataRoot.js` reads its
+// own env var directly rather than through anything a PATHS mock would touch.
+const REAL_REPO_DATA_DIR = join(
+  resolveInstallRoot(join(dirname(fileURLToPath(import.meta.url)), '../..')),
+  'data',
+);
+
+/**
+ * Structural guard against the bug class in #3683/#3687/#5605: a suite that
+ * exercises a route wired to `recordUserAction` without redirecting
+ * PATHS.data to a temp root would otherwise silently write
+ * `user-action-events.json` into the developer's live `data/` tree the next
+ * time such a route gets exercised — #5594 patched three known offenders
+ * one at a time, which is a per-suite fix, not a guard against the next one.
+ * Fires only under the test runner, and only at the moment a write is
+ * actually attempted, so a suite that merely reads (or never triggers a
+ * `recordUserAction` call) is unaffected either way.
+ */
+function assertTestDataRootRedirected() {
+  if (!isTestRunner() || PATHS.data !== REAL_REPO_DATA_DIR) return;
+  throw new Error(
+    'recordUserAction attempted a file-backend write of user-action-events.json ' +
+      "into the repo's real data/ tree. This suite exercises a route wired to " +
+      'recordUserAction but never redirected PATHS.data to a temp root - mock ' +
+      "`../lib/fileUtils.js` with lib/mockPathsDataRoot.js's makePathsProxy/" +
+      'createTempDataRoot (the same fix #5594 applied to cos.test.js / ' +
+      'cosTaskRoutes.test.js / cosAgentFeedback.test.js) rather than letting the ' +
+      'write land here.',
+  );
+}
 
 /** LLM-readable one-liner cap — the ledger is read by a model, not paged through. */
 export const SUMMARY_MAX_CHARS = 240;
@@ -283,6 +319,7 @@ function makeFileBackend() {
     record: (event) => queueWrite(async () => {
       const events = await loadFileEvents();
       if (events.some((row) => row.type === event.type && row.dedupeKey === event.dedupeKey)) return null;
+      assertTestDataRootRedirected();
       await ensureDir(PATHS.data);
       await atomicWrite(eventsFile(), { events: pruneEvents([...events, event]) });
       return event;
--- a/server/services/userActionsDataRootGuard.test.js
+++ b/server/services/userActionsDataRootGuard.test.js
@@ -11,28 +11,46 @@
  * would otherwise be the test that writes a real file into the repo's data/
  * tree, so it also asserts that never happens.
  */
-import { describe, it, expect, afterEach } from 'vitest';
-import { existsSync, rmSync } from 'node:fs';
+import { describe, it, expect } from 'vitest';
+import { existsSync, readFileSync, rmSync } from 'node:fs';
 import { join } from 'node:path';
 import { PATHS } from '../lib/fileUtils.js';
 import { recordUserAction } from './userActions.js';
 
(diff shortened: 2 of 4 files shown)


Bryandero98:
Thanks for leaving the guard mechanism up to the implementer — went with option 1 since it protects the next route/hook automatically rather than needing another per-suite patch. Glad to switch to one of the other two approaches if you have a preference I'm not seeing.

server/services/userActionsDataRootGuard.test.js
```diff
@@ -0,0 +1,38 @@
...
+import { PATHS } from '../lib/fileUtils.js';
+import { recordUserAction } from './userActions.js';
+
+const REAL_EVENTS_FILE = join(PATHS.data, 'user-action-events.json');
+
+afterEach(() => {
+  // Belt-and-suspenders: if the guard ever regresses, don't leave the leaked
+  // file behind for the next run to trip over.
+  rmSync(REAL_EVENTS_FILE, { force: true });
+});
```
atomantic: **[CRITICAL]** This `afterEach` deletes the developer's real `data/user-action-events.json`, and the test also fails on any machine where that file legitimately exists.

`MEMORY_BACKEND=file` is a documented (dev/test) escape hatch, so this path can hold a real operator ledger. This test is the one test that deliberately runs *unredirected*, so `REAL_EVENTS_FILE` points at the live tree. I reproduced it — with a pre-existing file, `cd server && npm test` gives:

```
AssertionError: expected true to be false
  ❯ services/userActionsDataRootGuard.test.js:36
AFTER: ls: data/user-action-events.json: No such file or directory
```

Two failures in one: the guard worked correctly yet the test reports red (non-hermetic), and the cleanup then permanently removes a file the test never created. A test added to *stop* leaks into `data/` must not itself destroy `data/`, per AGENTS.md's "never let tests touch real data" rule (the #3683/#3687 and 2026-06-13 Postgres scars).

Snapshot the pre-existing bytes instead, and only clean up a file this test actually leaked:

```suggestion
import { describe, it, expect, afterEach } from 'vitest';
import { existsSync, readFileSync, rmSync } from 'node:fs';
import { join } from 'node:path';
import { PATHS } from '../lib/fileUtils.js';
import { recordUserAction } from './userActions.js';

const REAL_EVENTS_FILE = join(PATHS.data, 'user-action-events.json');

// A file that is ALREADY here belongs to the developer — anyone running the
// documented `MEMORY_BACKEND=file` escape hatch has a real operator ledger at
// this path. Snapshot it so the test neither fails on it nor deletes it.
const BEFORE = existsSync(REAL_EVENTS_FILE) ? readFileSync(REAL_EVENTS_FILE) : null;

afterEach(() => {
  // Belt-and-suspenders: if the guard ever regresses, don't leave a LEAKED
  // file behind — but never remove one that was already there.
  if (BEFORE === null) rmSync(REAL_EVENTS_FILE, { force: true });
});
```

server/services/userActionsDataRootGuard.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+describe('recordUserAction — data-root guard (#5605)', () => {
+  it('throws instead of writing user-action-events.json into the real data/ tree', async () => {
+    await expect(recordUserAction({
+      type: 'cos.task.create',
+      summary: 'Guard regression probe',
+      dedupeKey: `guard-probe-${Math.random().toString(36).slice(2)}`,
+    })).rejects.toThrow(/real data\/ tree/);
+
+    expect(existsSync(REAL_EVENTS_FILE)).toBe(false);
```
atomantic: **[CRITICAL]** Paired with the snapshot above — comparing bytes proves "the guard wrote nothing" on a machine that already has the file, where `toBe(false)` cannot.

```suggestion
    const after = existsSync(REAL_EVENTS_FILE) ? readFileSync(REAL_EVENTS_FILE) : null;
    expect(after).toEqual(BEFORE);
```

server/services/userActions.js
```diff
@@ -283,6 +319,7 @@ function makeFileBackend() {
     record: (event) => queueWrite(async () => {
       const events = await loadFileEvents();
       if (events.some((row) => row.type === event.type && row.dedupeKey === event.dedupeKey)) return null;
+      assertTestDataRootRedirected();
```
atomantic: **[CRITICAL]** The guard runs *after* the read and *after* the dedupe short-circuit, so it has two holes.

1. **Dedupe bypasses it entirely.** Line 321 returns `null` before line 322 ever runs. An un-redirected suite replaying a `(type, dedupeKey)` that already exists in the real ledger silently no-ops — exactly the leak this PR exists to catch, undetected. Reproduced against the PR branch: seeding `data/user-action-events.json` with `{type:'cos.task.create', dedupeKey:'preexisting-key'}` and calling `recordUserAction` with the same pair returns `null` with **no throw**.

2. **`loadFileEvents()` has already read the real ledger** by then. That file carries operator payloads (redacted for credentials, but still machine-local, never-federated records per the privacy ADR). An untethered suite shouldn't get to read it either — the guard should fire before any I/O against the unredirected path.

Both close by hoisting the assertion to the first statement of the queued callback:

```suggestion
      assertTestDataRootRedirected();
      const events = await loadFileEvents();
      if (events.some((row) => row.type === event.type && row.dedupeKey === event.dedupeKey)) return null;
```

Worth considering the stronger placement too: putting the call at the top of `loadFileEvents()` (line 293) covers `list` as well, which is currently unguarded and lets an un-redirected suite read the live ledger through `listUserActions`.

server/services/userActions.js
```diff
@@ -55,6 +57,40 @@ import { insertUserActionEvent, listUserActionEvents, pruneUserActionEvents } fr
...
 
+// The REAL repo data/ dir, computed independently of the (possibly test-mocked)
+// `PATHS` import above via the same fileURLToPath/resolveInstallRoot technique
+// lib/paths.js itself uses — so a suite that redirects PATHS.data to a temp
+// root can't accidentally spoof this comparison too. `dataRoot.js` reads its
+// own env var directly rather than through anything a PATHS mock would touch.
+const REAL_REPO_DATA_DIR = join(
+  resolveInstallRoot(join(dirname(fileURLToPath(import.meta.url)), '../..')),
+  'data',
+);
```
atomantic: **[IMPROVEMENT]** Nice call deriving this independently of the (mockable) `PATHS` — that part is right, and I verified it matches `lib/paths.js` in all three cases (plain install, `PORTOS_DATA_ROOT` pin, CoS worktree checkout), since both anchor two levels above `server/*`.

The risk is that it **fails open**: this duplicates `paths.js`'s `INSTALL_ROOT` derivation in a second module, and if `paths.js` ever changes how it resolves the root, this constant silently stops matching and the guard degrades to a permanent no-op — a guard that is quietly off is worse than no guard, because the suite still looks green.

Cheapest durable fix: export the derivation once from `lib/dataRoot.js` (which no suite mocks) as e.g. `realRepoDataDir(fallbackRoot)` and have both `paths.js` and this call it, so there is one canonical formula. Non-blocking, but worth doing while the context is fresh.

atomantic (changes requested):
Reviewed by `/do:review` (security-focused pass, external contributor) — **2 critical, 1 improvement**.

**The approach is right and the ticket is satisfied in spirit.** Option 1 from #5605, correctly narrowed to the `recordUserAction` write chokepoint instead of blanket-wrapping `atomicWrite` — that scoping call is good, and it means #5596's growing hook allowlist stays covered for free. Deriving `REAL_REPO_DATA_DIR` independently of the mockable `PATHS` is the right technique; I verified it agrees with `lib/paths.js` for a plain install, a `PORTOS_DATA_ROOT` pin, and a CoS worktree checkout.

**Verified against the branch (worktree, isolated):**
- Full `cd server && npm test` — **1809 files / 36801 tests pass**, exit 0. No `data/user-action-events.json` after the run. AC #2 ✔
- The guard does fire on an unredirected write. AC #1 ✔
- Security scan: diff is pure ASCII apart from em-dashes (no zero-width/bidi), no network calls, no `process.env` reads, no `eval`/`exec`/`child_process`, no dependency or CI changes, production behavior unchanged (`isTestRunner()`-gated). Single commit, single author. Nothing injection-shaped in comments or prose.

**Blocking (please address):**

1. **The new test deletes real user data** (`userActionsDataRootGuard.test.js:22-26`). The `afterEach` `rmSync` is unconditional, and this is the one test that deliberately runs with `PATHS.data` unredirected — so it targets the live tree. On an install using the documented `MEMORY_BACKEND=file` escape hatch, `npm test` both **fails spuriously** (line 36 asserts the file is absent) and then **permanently removes the developer's ledger**. Reproduced end to end. A guard against leaking into `data/` must not itself destroy `data/`.

2. **The guard is bypassed by the dedupe short-circuit** (`userActions.js:321-322`). It sits after `loadFileEvents()` and after the `return null`, so an un-redirected suite replaying an existing `(type, dedupeKey)` silently no-ops with no throw — reproduced. The real ledger has also already been read into the test process by that point. Hoisting the assertion to the top of the queued callback closes both.

**Non-blocking:** the `REAL_REPO_DATA_DIR` derivation is a second copy of `paths.js`'s `INSTALL_ROOT` formula and **fails open** if that formula ever changes — worth collapsing into one exported helper in `lib/dataRoot.js`.

**Coherence check:** the PR description matches the diff; the test-plan claims hold up (I re-ran them). One AC gap worth noting: #5605 asks that the guard "names the offending suite" — the message names the *fix* but not the suite. Vitest's stack trace covers it in practice, so I'd call that satisfied, but a one-line mention would close it cleanly.

Thanks for taking this on — the structural instinct here is exactly right, and with the two fixes above it's a clear improvement over the per-suite patching in #5594.

_Generated by /do:review_

Bryandero98:
Addressed both blocking findings, plus the non-blocking suggestion:

1. **The destructive `afterEach`** — rewrote the guard test to snapshot the real file's presence/content *before* exercising the guard, assert it comes back unchanged (not merely absent, which was the source of the spurious failure on a `MEMORY_BACKEND=file` install), and only clean up a file this run's own regression would have created.
2. **The dedupe bypass** — hoisted `assertTestDataRootRedirected()` above both `loadFileEvents()` and the dedupe short-circuit's `return null`, so neither path can slip past it anymore.
3. **`REAL_REPO_DATA_DIR` drift risk** — extracted the `fileURLToPath`/`dirname`/`../..` depth formula into a new `resolveCodeRootForModule()` in `lib/dataRoot.js`. Both `paths.js`'s `CODE_ROOT` and `userActions.js`'s `REAL_REPO_DATA_DIR` now derive through it — one source of truth, `paths.js`'s own exported values unchanged.

Verified: full `cd server && npm test` (1793/1810 files pass), and confirmed via `git stash` that the 16 files / 58 tests that do fail are identical with or without this change — all pre-existing Windows-only noise (symlink creation needs Developer Mode privileges, a couple of `EBUSY` temp-file races, one missing local Python module), none touching `paths.js`/`dataRoot.js`/`userActions.js`. Also re-ran the guard test and `userActions.test.js` in isolation.

atomantic (approved):
Re-reviewed after the fix commit — **both blocking findings are resolved, and I verified each one independently rather than taking the description's word for it.** Approving.

**Verified against the branch (isolated worktree):**

1. **Destructive cleanup — fixed.** Seeded a real `data/user-action-events.json` (the `MEMORY_BACKEND=file` install shape that broke the previous version) and ran the guard test: it now **passes** and leaves the file byte-identical. The snapshot-then-compare rewrite plus the `!existedBefore` cleanup gate is exactly right, and the `try/finally` is a better shape than the `afterEach` I suggested.
2. **Dedupe bypass — fixed.** Reproduced the original hole's setup: seeded `{type:'cos.task.create', dedupeKey:'preexisting-key'}` and replayed the same pair. Before the hoist this returned `null` silently; now it throws, and `loadFileEvents()` never reads the real ledger.
3. **Bypass probe.** Removed `assertTestDataRootRedirected()` and re-ran — the guard test goes red with the leaked event visible in the diff. The test genuinely pins the contract rather than passing vacuously.
4. **Shared helper (non-blocking) — addressed.** `resolveCodeRootForModule` collapses the duplicated derivation, and `paths.js`'s `CODE_ROOT` is expression-identical to before, so the fail-open drift risk is gone.
5. **Full `cd server && npm test`** — no `data/user-action-events.json` after the run, and zero guard firings across all 36.5k tests (so no existing suite is falsely tripped). CI's Server tests job is green on this head.
6. **Security pass:** diff is pure ASCII apart from em-dashes (no zero-width/bidi), no `child_process`/`eval`/network/`process.env` additions, no dependency or workflow changes, production behavior unchanged (`isTestRunner()`-gated early return). The only destructive call is the now-correctly-gated `rmSync`.

**Ticket alignment (#5605):** option 1, scoped to the `recordUserAction` chokepoint. AC #1 ✔ (bypass probe), AC #2 ✔ (clean tree after a full run), AC #3 — the message names the fix but not the offending suite; Vitest's stack covers that in practice, so I'm calling it satisfied.

**Follow-ups I'll take on our side** (all minor, none worth another round-trip here):
- `list` is still unguarded — an un-redirected suite can read the live ledger through `listUserActions`. Read-only, so outside this ticket's AC.
- If the guard ever regresses on a machine that already has the ledger, the test leaves the probe event appended (it snapshots `contentBefore` but doesn't restore it).
- `lib/README.md`'s `dataRoot.js` row doesn't mention the new export.

Thanks for the thorough turnaround on this — the structural fix is a real improvement over the per-suite patching in #5594, and it keeps #5596's growing hook allowlist covered for free.

Issue #3266, opened 2026-08-01
Retire the hand-rolled mounted-ref pattern: 13 sites still inline useMounted's body


atomantic:
## Problem

#3264 fixed the fourteen sites whose mounted-guard ref was *broken* (never re-armed, so StrictMode left it permanently false). It did not touch the sites that hand-roll the same guard **correctly**.

Thirteen sites still inline the exact body of `client/src/hooks/useMounted.js`:

```js
const mountedRef = useRef(true);
useEffect(() => {
  mountedRef.current = true;
  return () => { mountedRef.current = false; };
}, []);
```

| File | Line |
|---|---|
| `client/src/components/songs/PitchTuner.jsx` | 63 |
| `client/src/components/pipeline/manuscript/ManuscriptReadAloud.jsx` | 49 |
| `client/src/components/loraTraining/LiveSampleGallery.jsx` | 55 |
| `client/src/components/writers-room/ExercisePanel.jsx` | 29 |
| `client/src/components/media/MediaJobsQueue.jsx` | 243 |
| `client/src/components/meatspace/post/PostCognitiveDrillRunner.jsx` | 433, 547, 1071 |
| `client/src/hooks/useAutoscroll.js` | 29 |
| `client/src/hooks/useMediaJobProgress.js` | 56 |
| `client/src/pages/WritersRoom.jsx` | 38 |
| `client/src/pages/Ask.jsx` | 405 |
| `client/src/pages/MoodBoardDetail.jsx` | 65 |

These are all **correct today** — they do set `true` on setup, so the #3264 guard (`client/src/hooks/mountedRefConventions.test.js`) passes them. This is cleanup, not a correctness fix.

Why it is still worth doing: several of these carry multi-line comments *explaining the StrictMode hazard from first principles*, which is direct evidence the hazard is being re-derived per site instead of read off the hook name. Copy-pasting any one of them and dropping the setup line reproduces #3264 exactly — and #3264's own history is that `useMounted` already existed, was used at ~40 sites, and fourteen places hand-rolled the broken form anyway.

## Chosen approach

Mechanical swap to `useMounted()`, one commit, no behavior change:

- Replace the ref + effect pair with `const mountedRef = useMounted();` and import from the appropriate relative path (`hooks/useMounted`).
- Drop the per-site StrictMode explanation comments — `useMounted.js` owns that rationale. Keep any comment describing *what this particular ref guards* (the deferred work), which stays true.
[...]

PR #5906, opened 2026-09-02
docs: correct two stale rules in the portos-socket-ui skill

tzioup:
## Summary

`.claude/skills/portos-socket-ui/SKILL.md` is required reading before wiring a socket-driven view (root `AGENTS.md`), and two of its rules no longer match the code.

**Rule on deferred work** prescribes a `useRef(true)` plus cleanup-only mounted guard and calls it dev-mode-safe. That snippet is the same shape as the `broken` fixture in `client/src/hooks/mountedRefConventions.test.js:200-203`, which `:228` asserts the detector flags. Following it verbatim fails `it('has no ref that is only ever set to false')` (`:173`, assertion at `:185-194`).

The parenthetical is inverted. `client/src/hooks/useMounted.js:5-12` re-arms `.current = true` on every mount *because* StrictMode's mount-cleanup-remount would otherwise leave it permanently false — the two user-visible freezes recorded in commit `69e1c7367`. A second rule (#3266) then banned the corrected hand-roll too, leaving `useMounted()` the only permitted spelling, so a reader who fixes only the obvious half fails a second test.

**Rule on pending-request tracking** cites `client/src/pages/Shell.jsx` for `pendingAttachRef`. That symbol now lives in `client/src/hooks/useShellSession.js:126` (22 occurrences); `Shell.jsx` has none. The socket lifecycle was extracted to the hook and the skill was never repointed.

Why it drifted: SKILL.md was last touched 2026-07-24 (`643ac1826`); the guard test landed 2026-07-31 (`69e1c7367`, issue 3264) and the follow-up ban after that (#3266). Nothing tests SKILL.md itself, which is why it went unnoticed.

This matters past one file: root `AGENTS.md` names `portos-socket-ui` as required reading before wiring or debugging a socket-driven view, so the stale rule is what a reader reaches for at exactly the moment they are about to write the guard.

## Test plan

- `cd client && npx vitest run src/hooks/mountedRefConventions.test.js` — 8 passed before and after; the guard is green today, so a skill-following change is what turns it red.
- `git grep pendingAttachRef -- client/` confirms the corrected path.
- Docs-only change; no code touched.


atomantic (approved):
Docs-only change: 2 lines in .claude/skills/portos-socket-ui/SKILL.md, no code touched. Both corrections verified against the tree at base commit 55cb9715a.

(1) Deferred-work bullet. The removed inline snippet (a useRef(true) whose only assignment is false in a cleanup-only effect) is the same shape as the 'broken' fixture at client/src/hooks/mountedRefConventions.test.js:200-203, which :228 asserts findOneWayRefs flags; the old parenthetical calling that shape dev-mode-safe is inverted relative to client/src/hooks/useMounted.js:5-12, which re-arms .current = true on mount precisely because StrictMode's mount-cleanup-remount cycle would otherwise leave it permanently false. Both claims in the replacement text hold: useMounted() re-arms on mount (rule 1, test at :173), and the guard separately rejects any same-effect re-implementation of the hook (rule 2, test at :237). The new prose is very slightly stricter than the grep -- the detector's own documented blind spots at :47-72 include a raise and lower split across two effects -- which is the correct direction for guidance, since those are detector gaps rather than sanctioned patterns.

(2) Pending-request bullet. pendingAttachRef is declared at client/src/hooks/useShellSession.js:126 (22 occurrences in that file) and has 0 occurrences in client/src/pages/Shell.jsx, so the old pointer was stale. cancelPendingAttach (:136) and clearActiveSession (:398), the two other symbols that same bullet names, also live in the hook -- the whole pointer was stale, not just one symbol, so repointing the line is the complete fix. client/src/hooks/README.md:82 independently documents useShellSession as owning pendingAttachRef. The drift account checks out: SKILL.md's only commit is 643ac1826, and the guard test's commits post-date it. AGENTS.md:222 does name this skill as required reading before wiring a socket-driven view, so the stale rule was reachable at the moment of use.

Test evidence. Narrowest relevant test: from client/, npx vitest run src/hooks/mountedRefConventions.test.js -> 1 file / 8 tests passed, matching the PR's stated baseline. Broader: full client suite -> 820 files passed + 1 skipped, 10073 tests passed + 2 skipped, exit code 0. Nothing in the suite reads .claude/skills/**, so a markdown edit there cannot change suite outcome; the runs establish a green baseline rather than a before/after delta.

How the patch was applied, and two caveats. The supplied .portos-public-review/PR-5906.patch is missing its final newline, so git apply --check on it verbatim fails with 'corrupt patch at line 17'. That is a serialization artifact of the read-only review copy, not a property of the PR: the target file ends with a newline and the patch carries no 'No newline at end of file' marker. Restoring only that single trailing byte -- content otherwise byte-identical -- makes git apply --check pass. Separately, this stage's sandbox denies writes under .claude/, so the change was applied with git apply --cached (1 file changed, 2 insertions, 2 deletions) and the resulting file content read back from the index and verified line by line, rather than written into the working tree. The worktree was returned to its clean base afterward. Flagging both so the coordinator can downgrade to defer if it requires a verbatim working-tree apply; the reviewed content is identical either way.

No blocking or non-blocking defects found. Every factual claim in both replacement lines is verified against the current tree.

PR #5929, opened 2026-09-02
docs: document the fork customization workflow in SELF_UPDATE

tzioup:
## Summary

The branch-and-rebase workflow for fork users lives in exactly one place today: the `UpdateTab`
fork panel (`client/src/components/apps/tabs/UpdateTab.jsx:443-444`). That panel renders only when
`isFork` is true, and only if the user opens that tab. `docs/SELF_UPDATE.md` documents the update
*mechanism* thoroughly — the pull, the forced switch to `main`, pre-checkout stashing and recovery
— but never the workflow that mechanism implies.

This adds one section, `## Running a customized fork`, after "Fork UI: three distinct buttons".
No code change.

It writes out the loop the tip states in short form, plus the half that is currently undocumented
anywhere: after rebasing your branch onto `main` you have to reinstall, rebuild and restart, since
`update.sh` always finishes on `main` and cannot do it for you. It also notes that install order
matters, because `npm install` rewrites `client/package.json` and re-stales an earlier build.

Finally it documents a side effect that reads as a fault and is not one: a clean rebase rewinds the
working tree to the branch's older base and rolls it forward, giving every touched file a new
mtime, so `getInstallState()` reports a stale build and stale deps in all four workspaces with no
dependency changed and no build input edited.

## Test plan

No code paths change, so there is nothing to unit test. Every claim was reproduced on a fork
install running 2.56.0:

- Rebased a private branch onto a freshly updated `main`. `getInstallState()` immediately returned
  `staleBuild: true` and `staleDeps.stale: true` for `root`, `client`, `server` and `autofixer`,
  each with `reason: "manifest-newer"`, though the branch touches no manifest.
- Ran `npm run build` alone: `staleBuild` cleared, `staleDeps.stale` stayed true.
- Ran `npm install` across the four workspaces: `staleDeps.stale` cleared and `staleBuild` returned
  to true, because the install rewrote `client/package.json`. Rebuilding after the install cleared
  both, which is the order the new section prescribes.
- Confirmed the running server executes the checked-out branch rather than `main`: with the same
  tree on disk, `main`'s copy of `installState.js` computed `staleBuild: true` while the branch's
  computed `false`, and the live status endpoint agreed with the branch. `getInstallState()` has no
  cache and `updateChecker.js:284` calls it per request, so the reading is live.
- No docs linter exists in `package.json`; this is a prose-only change to one file.

### DIFF
--- a/docs/SELF_UPDATE.md
+++ b/docs/SELF_UPDATE.md
@@ -105,6 +105,39 @@ When `isFork` is true, `UpdateTab` replaces the single "Update Now" button with 
 
 Keep these three behaviors distinct. Collapsing them strips the user's agency over what touches their GitHub fork.
 
+## Running a customized fork
+
+The `UpdateTab` fork panel states this in short form. The full loop is here, because that panel is
+the only other place it exists and it renders only when `isFork` is true.
+
+Keep `main` a clean mirror of upstream and never commit to it — that is what keeps `gh repo sync`
+fast-forward and avoids 409 `FORK_DIVERGED`. Private changes live on their own branch, rebased
+onto `main` after each sync. Anything shareable goes upstream as a PR instead, so you carry less
+forward each time.
+
+**PM2 boots whatever is checked out**, so the branch you are on is the code that runs. Staying on
+your private branch is what makes your customizations live; there is no separate step.
+
+### After rebasing, reinstall and rebuild
+
+`update.sh` always finishes on `main`, so it cannot do this half for you:
+
+```bash
+git checkout main && ./update.sh
+git checkout <your-branch> && git rebase main
+for d in . client server autofixer; do (cd "$d" && npm install); done
+npm run build && npm run pm2:restart
+```
+
+Order matters. `npm install` rewrites `client/package.json`, so building before installing leaves
+the build stale again.
+
+**A clean rebase makes the install look broken.** Checking out a branch based on an older `main`
+rewinds the working tree, and the rebase rolls it forward, so every touched file gets a new mtime.
+`getInstallState()` (`server/services/installState.js`) compares mtimes, so it reports a stale
+build and stale deps in all four workspaces even though no dependency changed and no build input
+was edited. The reinstall and rebuild above clear it.
+
 ## Image-bearing Persistent Mind work must drain before source transitions
 
 The managed update route refuses to restart into a different source revision while a queued Persistent Mind message or active turn carries image references. `GET /api/update/status` reports the privacy-safe `persistentMindImages` preflight (`safe`, queued count, and active-turn boolean), and `POST /api/update/execute` re-checks it before and after acquiring the update lock.
--- a/docs/SELF_UPDATE.md
+++ b/docs/SELF_UPDATE.md
@@ -120,17 +120,27 @@ your private branch is what makes your customizations live; there is no separate
 
 ### After rebasing, reinstall and rebuild
 
-`update.sh` always finishes on `main`, so it cannot do this half for you:
+`update.sh` / `update.ps1` always finish on `main`, so they cannot do this half for you:
 
 ```bash
-git checkout main && ./update.sh
+git checkout main && ./update.sh            # .\update.ps1 on Windows
 git checkout <your-branch> && git rebase main
-for d in . client server autofixer; do (cd "$d" && npm install); done
+for d in . client server autofixer; do (cd "$d" && npm install --no-save); done
+node scripts/trusted-rebuilds.js server
 npm run build && npm run pm2:restart
 ```
 
-Order matters. `npm install` rewrites `client/package.json`, so building before installing leaves
-the build stale again.
+**Use `--no-save`, not a bare `npm install`.** `--no-save` is what `safe_install` in `update.sh`
+and `scripts/ensure-deps.js` run, for two reasons that both bite a fork: a bare install rewrites
+`client/package.json` and the lockfiles, which dirties your private branch and re-stales a build
+you just made (`client/package.json` is a `staleBuild` input — see below); and an older npm can
+strip newer lockfile metadata it does not understand. `--no-save` still honors `package-lock.json`.
+
+`scripts/trusted-rebuilds.js` is not optional for the server. Every workspace `.npmrc` sets
+`ignore-scripts=true`, so the server's native addons (`node-pty` and friends) are never built by
+the install itself — skip the rebuild and the shell and TUI features crash on a missing binding.
+
+Order matters: install before you build, so the build sees the deps it compiles against.
 
 **A clean rebase makes the install look broken.** Checking out a branch based on an older `main`
 rewinds the working tree, and the rebase rolls it forward, so every touched file gets a new mtime.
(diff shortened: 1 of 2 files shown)


docs/SELF_UPDATE.md
```diff
@@ -85,6 +85,39 @@ When `isFork` is true, `UpdateTab` replaces the single "Update Now" button with
...
+your private branch is what makes your customizations live; there is no separate step.
+
+### After rebasing, reinstall and rebuild
+
+`update.sh` always finishes on `main`, so it cannot do this half for you:
+
+```bash
+git checkout main && ./update.sh
+git checkout <your-branch> && git rebase main
+for d in . client server autofixer; do (cd "$d" && npm install); done
```
atomantic: ⛔ **Blocking** — Bare `npm install` contradicts the repo's own install path

PortOS's install path deliberately uses `npm install --no-save` in all four of these workspaces — `update.sh:96` and `:102`, `update.ps1:137` and `:144` — with an in-tree comment stating why: "This is an installation/reconciliation path, not dependency authoring. --no-save still honors package-lock.json but prevents older npm versions from rewriting newer lockfile metadata (for example `libc` fields)."

All four lockfiles are tracked in git (`package-lock.json`, `client/`, `server/`, `autofixer/`), so a fork user who copy-pastes this line can have npm rewrite tracked lockfiles on their private branch. The wrong outcome is exactly the friction this section exists to prevent: on the next cycle `git checkout main && ./update.sh` finds a dirty tree, stashes it (`update.sh:132-143`), and hands the user back a `git checkout <branch> && git stash pop` recovery step — and the rebase now carries lockfile churn the branch never intended.

```suggestion
for d in . client server autofixer; do (cd "$d" && npm install --no-save); done
```

docs/SELF_UPDATE.md
```diff
@@ -85,6 +85,39 @@ When `isFork` is true, `UpdateTab` replaces the single "Update Now" button with
...
+npm run build && npm run pm2:restart
+```
+
+Order matters. `npm install` rewrites `client/package.json`, so building before installing leaves
```
atomantic: 💡 **Non-blocking** — Ordering rationale is downstream of the line 108 fix

`npm install --no-save` does not write `client/package.json`, so once the command above uses `--no-save` the stated cause ("`npm install` rewrites `client/package.json`") no longer holds and this paragraph documents a mechanism the prescribed command no longer triggers.

The install-before-build order is still worth keeping — a dependency change must land before the bundle is built — so restate the reason rather than deleting it. Tying it to the reinstall touching build inputs under `client/` generally, which `isClientSourceNewer()` (`server/services/installState.js:118-135`) stats wholesale, keeps the paragraph true either way.

docs/SELF_UPDATE.md
```diff
@@ -85,6 +85,39 @@ When `isFork` is true, `UpdateTab` replaces the single "Update Now" button with
...
+
+**PM2 boots whatever is checked out**, so the branch you are on is the code that runs. Staying on
+your private branch is what makes your customizations live; there is no separate step.
+
+### After rebasing, reinstall and rebuild
+
+`update.sh` always finishes on `main`, so it cannot do this half for you:
+
+```bash
+git checkout main && ./update.sh
```
atomantic: 💡 **Non-blocking** — `./update.sh` drops Windows; every other mention in this file pairs the two

Line 5 (`update.sh` / `update.ps1`), line 19, and line 47 all name both, and `update.ps1` implements the identical flow — so a Windows fork user copy-pasting `./update.sh` gets nothing.

Prefer the cross-platform entry point the root manifest already provides: `npm run update` (`package.json:44`) dispatches to `update.ps1` on win32 and `update.sh` elsewhere. Otherwise mirror this file's existing convention and write `./update.sh` / `update.ps1`. The prose on line 103 has the same omission.

```suggestion
git checkout main && npm run update
```

atomantic (changes requested):
🔴 **Changes requested**

The new section is accurate, well scoped, and correctly placed after "Fork UI: three distinct buttons". One blocking problem: the prescribed install command uses bare `npm install`, which contradicts the repo's own install path and can dirty tracked lockfiles.

**Scope:** docs-only — a `docs/README.md` index row plus a new "Running a customized fork" section in `docs/SELF_UPDATE.md`

#### ⛔ Blocking (1)
- `docs/SELF_UPDATE.md:108` — Bare `npm install` contradicts the repo's own install path

#### 💡 Non-blocking (2)
- `docs/SELF_UPDATE.md:112` — Ordering rationale is downstream of the line 108 fix
- `docs/SELF_UPDATE.md:106` — `./update.sh` drops Windows; every other mention in this file pairs the two

#### Test evidence
- ❌ `git apply --check (full patch)` — the `docs/README.md` hunk no longer applies against base c88fb247f — a `MANAGED_APP_UPDATES.md` row landed between the SELF_UPDATE.md and DEPS.md rows after this branch was cut. Pure context drift, so a rebase is required
- ✅ `git apply --check --include=docs/SELF_UPDATE.md` — applies cleanly; `--include=docs/README.md` is the only failing half
- ⏭️ `npx vitest` — the disposable review worktree has no node_modules in root/client/server and the stage forbids package downloads, so the attempt was refused at the registry

#### Notes
- No docs-index or markdown-link guard test exists under `scripts/` or `server/` that this change would exercise, so the untested surface is prose only — verification here was by reading the applied section against the code it describes.
- The `docs/README.md` hunk needs a rebase before this can merge. That is context drift from a change that landed after the branch was cut, not an authoring error.

- `update.sh` always ends on `main` — it force-checks-out main and then installs, builds, and restarts there (`update.sh:132-143`)
- `npm run build` and `npm run pm2:restart` exist (`package.json:18`, `package.json:22`)
- `getInstallState()` lives in `server/services/installState.js` and is mtime-based
- `DEP_WORKSPACES = ['.', 'client', 'server', 'autofixer']` renders `.` as `root`, matching the "all four workspaces" claim, and `manifest-newer` is a real reason string (`server/services/installState.js:46,174`)
- `client/package.json` genuinely is a staleBuild input — `isClientSourceNewer()` stats every file directly under `client/` (`server/services/installState.js:118-135`), so the install-then-build ordering claim is mechanically correct
- `gh repo sync` fast-forward and the 409 `FORK_DIVERGED` response match `server/routes/update.js:179` and the existing `docs/SELF_UPDATE.md:74`
- the `UpdateTab` fork tip the section references exists (`UpdateTab.jsx:444`)
</details>

atomantic (approved):
✅ **Approved** — rebased onto `main` and the blocking finding is fixed in `264d2340a`.

- **Conflict resolved.** The `docs/README.md` hunk was pure context drift (a `MANAGED_APP_UPDATES.md` row landed between the `SELF_UPDATE.md` and `DEPS.md` rows after this branch was cut). Rebase keeps both rows; the index row now reads `…, fork sync, running a customized fork`.
- **Blocking (`SELF_UPDATE.md:108`) fixed.** The loop now prescribes `npm install --no-save`, matching `safe_install` in `update.sh` and `scripts/ensure-deps.js`, with the two reasons spelled out: a bare install rewrites `client/package.json` and the lockfiles (dirtying a fork's private branch and re-staling the build the next line asks for), and an older npm strips newer lockfile metadata.
- **Ordering rationale reworded** — it no longer hangs off the `client/package.json` rewrite, which `--no-save` prevents. It now just states the real invariant: install before you build.
- **`update.ps1` paired with `update.sh`**, matching every other mention in the file.
- **One addition beyond the review:** the loop was missing `node scripts/trusted-rebuilds.js server`. Every workspace `.npmrc` sets `ignore-scripts=true`, so following the loop as written left `node-pty` unbuilt and the shell/TUI features crashing on a missing binding.

Thanks for writing this up — the `getInstallState()` mtime side effect in particular is a real support question that had no answer anywhere.

Issue #5960, opened 2026-09-03
A non-thinking ollama model makes every local reviewer AND the claim-comment gate fail, silently shrinking the claimable queue


atomantic:
**Problem** — `runToolFreeLocalCompletion` in `server/services/codeReview.js` attaches `reasoning_effort` to the `/v1/chat/completions` body whenever a reviewer effort is configured. Ollama translates `reasoning_effort` into its `thinking` parameter, and a model that does not implement thinking rejects the whole request:

```
ollama API error 400: {"error":{"message":"\"<model>\" does not support thinking","type":"invalid_request_error",...}}
```

There is no retry and no capability probe, so a reviewer configured as `ollama[<non-thinking model>]~effort=low` is **100% unavailable** — not degraded, not slower, just dead for every call. The effort suffix is what kills it; the same model answers fine with the field omitted.

**Two user-visible consequences**, both silent:

1. **Local code review never runs.** `/do:next` and `/claim` record an inconclusive verdict every time. For an `~opt` reviewer that is merely a lost review; for a required one it forces `review-blocked`, so the PR opens and then sits unmerged waiting on a reviewer that can never answer.
2. **The tool-free public-comment gate fails closed, and skips claimable work.** `runLocalClaimCommentReview` goes through the same helper. The claim flow treats a failed gate as "skip this candidate for this run", so every issue that has *any* comment becomes unclaimable while the misconfiguration stands. Observed live on 2026-09-03: a `--swarm=3` run had to skip #5704 and #5726 — both otherwise eligible, both with exactly one comment — and fall through to younger issues. Nothing in the run output said "your reviewer model is misconfigured"; it just looked like a shorter queue.

**Evidence** — `server/services/codeReview.js:283-289`:

```js
  const body = {
    model,
    messages,
    temperature: 0.2,
    stream: false,
    ...(resolvedEffort ? { reasoning_effort: resolvedEffort } : {}),
  }
  const response = await fetchWithTimeout(`${baseUrl}/v1/chat/completions`, { ... })
```

Reproduce against any non-thinking ollama model:

```bash
printf 'diff --git a/x.js b/x.js\n+const a = 1;\n' \
  | jq -Rs --arg backend ollama --arg model '<non-thinking-model>' --arg effort low \
      '{ backend: $backend, model: $model, effort: $effort, diff: . }' \
  | node server/scripts/run-local-code-review.mjs
# -> {"ok":false,...,"error":"ollama API error 400: ... does not support thinking"}
```

[...]

PR #5966, opened 2026-09-03
refactor: drop 21 exported-but-unwired Zod schemas and guard against new ones (#5730)

### LINKED ISSUE #5730: Wire or drop the exported-but-unused Zod sub-schemas in server/lib
**Problem** — `server/AGENTS.md` states the rule directly: "the PortOS convention is *all* inputs validated — an exported-but-unwired sub-schema creates false confidence." Twenty exported Zod schemas across five validation modules have no consumer anywhere in the repo — not a route, not a sibling schema, not a test. Each one reads at review time as "this payload is validated" when nothing validates it, and several were clearly written speculatively for a route that was never added.

**Evidence** — the same zero-reference scan as above, run over `server/lib/*Validation.js` and `server/lib/validation.js`:

- `server/lib/pipelineValidation.js:599-604` — a block of six kind-neutral aliases created so "the Pipeline routes share the same validation surface". Only ONE of the six (`placeBibleCreateSchema`) ever acquired a caller; the other five are dead: `characterBibleCreateSchema`, `characterBibleUpdateSchema`, `placeBibleUpdateSchema`, `objectBibleCreateSchema`, `objectBibleUpdateSchema`.
- `server/lib/brainValidation.js:144` — `memoryRecordSchema`, `digestRecordSchema`, `reviewRecordSchema`, `bucketRecordSchema`.
- `server/lib/digitalTwinValidation.js:28` — `valuesTestResultEnum`, `adversarialTestResultEnum`, `testResultSchema`, `goodreadsBookSchema`, `spotifyEntrySchema`, `letterboxdFilmSchema`, `calendarEventSchema`, `importAnalysisResultSchema`.
- `server/lib/agentValidation.js:100` — `platformAccountUpdateSchema = partialWithoutDefaults(platformAccountSchema)`. `server/routes/platformAccounts.js` has no PUT/PATCH at all (`router.post('/')`, `/:id/test`, `/:id/claim` only), so this is a partial-update schema for an update route that does not exist.
- `server/lib/validation.js:136` — `referenceRepoSchema`.

Verified individually, e.g. `grep -rn '\bcharacterBibleCreateSchema\b' server client/src scripts` returns one hit — the alias definition itself.

**Plan**
1. For each of the twenty, decide by checking whether a route accepts the payload it describes:
   - If a route DOES accept it unvalidated, wire it (`validateRequest(schema, req.body)` for POST, `partialWithoutDefaults(schema)` for PUT/PATCH) and keep the export.
   - Otherwise delete the export.
2. Apply the decision now for the ones already traced: delete the five dead `pipelineValidation.js` aliases (lines 599, 600, 602, 603, 604) and keep `placeBibleCreateSchema`, adjusting the block comment to describe the single alias that remains and why it is the un-refined object form. Delete `platformAccountUpdateSchema` (`agentValidation.js:100`) — there is no update route, and re-adding it alongside a real PUT is one line.
3. Do NOT delete `soulMetaSchema` (`digitalTwinValidation.js:330`) even though it is unreferenced: it is labelled "Alias for backwards compatibility" and compat surface is not removed on unused-ness grounds.
4. Keep the schemas that describe a PERSISTED record shape rather than a request body (`memoryRecordSchema`, `digestRecordSchema`, `reviewRecordSchema`, `bucketRecordSchema`, `goodreadsBookSchema`, `spotifyEntrySchema`, `letterboxdFilmSchema`, `calendarEventSchema`) only if they are wired into the corresponding importer/sanitizer in the same PR; otherwise delete them. Decision rule, stated once: a schema either runs on some input or it does not exist.
5. Extend `server/lib/validationSplit.test.js` with a scan that walks `server/lib/*Validation.js` + `server/lib/validation.js`, collects every `export const <name>Schema|<name>Enum`, and fails on any name with no reference outside its own module and the barrel — with an `INTENTIONALLY_UNWIRED` allowlist seeded with `soulMetaSchema` so a compat alias is an explicit, reviewed entry.

**Tests** — the new guard in `server/lib/validationSplit.test.js` is the deliverable; it uniquely catches "a sub-schema is exported but never wired", which is exactly the false-confidence failure `server/AGENTS.md` names. For each schema that gets WIRED instead of deleted, add one route-level rejection case to that route's existing suite (a payload violating the newly-enforced constraint returns 400) — the boundary test, not a unit test of the schema.

**Acceptance criteria**
- [ ] Every exported schema in `server/lib/*Validation.js` and `server/lib/validation.js` is either referenced by a route/sibling schema or listed in `INTENTIONALLY_UNWIRED`.
- [ ] The five dead pipeline bible aliases and `platformAccountUpdateSchema` are gone.
- [ ] `soulMetaSchema` is retained and allowlisted.
- [ ] `cd server && npm test` passes.

**Out of scope** — `server/lib/cosValidation.js` and `server/lib/postValidation.js` restructuring; adding new routes to make a schema reachable; the client-side mirrors of any of these enums.

---
Filed by a `/do:better --scan-only --issues` audit (2026-09-01). Category: `dry` · Severity: `low` · Files: `server/lib/pipelineValidation.js:599, server/lib/brainValidation.js:144, server/lib/digitalTwinValidation.js:28, server/lib/agentValidation.js:100, server/lib/validation.js:136`

All labels already exist in the repo; do NOT create labels. Never add `planner:*` labels.

atomantic:
## Summary

`server/AGENTS.md` requires that *all* inputs be validated and warns that "an exported-but-unwired sub-schema creates false confidence." A Zod schema exported from a validation module but referenced nowhere reads at review time as "this payload is validated" when nothing validates it. Twenty-one such schemas had accumulated across five modules, and several had already drifted from the shape they claimed to describe — `referenceRepoSchema` was missing the `lastKnownGoodSnapshot` field the service has been writing for months, which is exactly the failure mode the rule names.

Applied one decision rule — **a schema either runs on some input or it does not exist** — after checking each against the routes that accept the payload it describes. No route accepted any of the twenty-one unvalidated, so all twenty-one are deleted rather than wired:

- **`pipelineValidation.js`** — the whole kind-neutral `*BibleCreateSchema` / `*BibleUpdateSchema` alias block. Every writers-room and Pipeline bible route validates through the `writersRoom*` schemas directly.
- **`agentValidation.js`** — `platformAccountUpdateSchema` (`server/routes/platformAccounts.js` has no PUT/PATCH) and `featureAgentStatusSchema` (status is server-owned, stamped by the start/pause/complete routes, never accepted from a body).
- **`brainValidation.js`** — `memoryRecordSchema`, `digestRecordSchema`, `reviewRecordSchema`, `bucketRecordSchema`: persisted-record shapes with no importer or sanitizer behind them. Their wired input counterparts (`bucketInputSchema` etc.) are untouched.
- **`digitalTwinValidation.js`** — `testResultEnum`, `valuesTestResultEnum`, `adversarialTestResultEnum`, `testResultSchema`, the four import-entry shapes (`goodreadsBookSchema`, `spotifyEntrySchema`, `letterboxdFilmSchema`, `calendarEventSchema`) and `importAnalysisResultSchema`. The import route validates its request through `importDataInputSchema`; these described parser/LLM output that never reached a validator.
- **`validation.js`** — `referenceRepoSchema`. Its domain documentation moves onto `referenceRepoCreateSchema`, which is the wired one, with a note that the server-owned fields are stamped by `services/referenceRepos.js`.

`soulMetaSchema` is **kept**: it is a labelled backwards-compatibility alias, and compat surface is not removed on unused-ness grounds (distribution model — other installs and forks import the pre-rename name).

**New guard.** `server/lib/validationSplit.test.js` gains a scan over every `server/lib/*Validation.js` plus `validation.js`: it collects each exported `*Schema` / `*Enum` and fails on any name with no reference outside its own declaration and the barrel. `INTENTIONALLY_UNWIRED` is seeded with the one compat alias, and a second test fails when an allowlist entry stops naming a real export, so the allowlist can't rot. The scan builds a single identifier index over the git-tracked tree rather than re-grepping per name, so it runs in well under a second.

### Deviations from the filed plan
- The plan expected `placeBibleCreateSchema` to have retained a caller. It no longer does (verified by direct grep across `server`, `client/src`, `scripts`), so the alias block is removed in full rather than keeping one orphan.
- The plan listed twenty schemas; the scan found twenty-one — `featureAgentStatusSchema` was not in the filed list. Removing `testResultSchema` then orphaned `testResultEnum`, which is removed in the same pass.
- No schema was wired, because none had a route accepting its payload unvalidated. Per the plan's stated fallback ("otherwise delete them"), there is consequently no new route-level 400 test to add.

## Test plan
- `cd server && npm test` — **1879 files passed / 1 skipped, 37881 tests passed / 14 skipped**. No DB-backed suites run (`npm run test:db` not invoked).
- The new guard was verified to actually fail, not merely pass: appending a throwaway `bogusUnwiredProbeSchema` export to `notesValidation.js` made it fail naming `notesValidation.js:bogusUnwiredProbeSchema`, and adding a bogus `INTENTIONALLY_UNWIRED` entry made the allowlist-freshness test fail naming it. Both probes were reverted.
- Barrel/README maintenance rule is a no-op here: no module was added or removed, only exports within existing modules, so `server/lib/index.js` and `server/lib/README.md` are unchanged and `server/lib/index.test.js` stays green.
- Per `.changelog/README.md`, no per-branch changelog entry is written.

> Note: the configured local reviewer (`ollama` / `hf.co/yuxinlu1/gemma-4-12B-coder-fable5-composer2.5-v1-GGUF:Q4_K_M`) returned `400 … does not support thinking` — the known failure tracked in #5960 — so the required review pass is **inconclusive**, not clean.

Closes #5730

https://claude.ai/code/session_01SQQHNCXHxrXNaJ4FEk8N23

### DIFF
--- a/server/lib/agentValidation.js
+++ b/server/lib/agentValidation.js
@@ -97,8 +97,6 @@ export const platformAccountSchema = z.object({
   platformData: z.record(z.unknown()).optional().default({})
 });
 
-export const platformAccountUpdateSchema = partialWithoutDefaults(platformAccountSchema);
-
 // Account registration (when creating new Moltbook account)
 export const accountRegistrationSchema = z.object({
   agentId: z.string().min(1),
@@ -317,7 +315,9 @@ export const moltworldQueueAddSchema = z.object({
 // FEATURE AGENT SCHEMAS
 // =============================================================================
 
-export const featureAgentStatusSchema = z.enum(['draft', 'active', 'paused', 'completed', 'error']);
+// A feature agent's `status` is server-owned — stamped by the start/pause/
+// complete routes, never accepted from a request body — so it has no request
+// schema (#5730).
 export const featureAgentScheduleModeSchema = z.enum(['continuous', 'interval']);
 export const featureAgentAutonomySchema = z.enum(['standby', 'assistant', 'manager', 'yolo']);
 export const featureAgentPrioritySchema = z.enum(['LOW', 'MEDIUM', 'HIGH', 'CRITICAL']);
--- a/server/lib/brainValidation.js
+++ b/server/lib/brainValidation.js
@@ -140,17 +140,6 @@ export const adminRecordSchema = z.object({
   updatedAt: z.string().datetime()
 });
 
-// Memory Record schema (journal entries, daily notes, personal memories)
-export const memoryRecordSchema = z.object({
-  id: z.string().guid(),
-  title: z.string().min(1).max(200),
-  content: z.string().max(10000).optional().default(''),
-  mood: z.string().max(50).optional(),
-  tags: z.array(z.string().max(50)).optional().default([]),
-  createdAt: z.string().datetime(),
-  updatedAt: z.string().datetime()
-});
-
 // Meta/Settings schema
 export const brainSettingsSchema = z.object({
   version: z.number().int().positive().default(1),
@@ -162,29 +151,6 @@ export const brainSettingsSchema = z.object({
   defaultModel: z.string().default('gptoss-20b'),
   lastDailyDigest: z.string().datetime().optional(),
   lastWeeklyReview: z.string().datetime().optional()
-});
-
-// Digest Record schema
-export const digestRecordSchema = z.object({
-  id: z.string().guid(),
-  generatedAt: z.string().datetime(),
-  digestText: z.string().max(2000),
-  topActions: z.array(z.string().max(200)).max(3),
-  stuckThing: z.string().max(200),
-  smallWin: z.string().max(200),
-  ai: aiConfigSchema.optional()
-});
-
-// Weekly Review Record schema
-export const reviewRecordSchema = z.object({
-  id: z.string().guid(),
-  generatedAt: z.string().datetime(),
-  reviewText: z.string().max(3000),
-  whatHappened: z.array(z.string().max(200)).max(5),
-  biggestOpenLoops: z.array(z.string().max(200)).max(3),
-  suggestedActionsNextWeek: z.array(z.string().max(200)).max(3),
-  recurringTheme: z.string().max(500),
-  ai: aiConfigSchema.optional()
 });
 
 // --- Input schemas for API endpoints ---
@@ -556,17 +522,6 @@ export const linksQuerySchema = z.object({
 export const bucketColorEnum = z.enum([
   'accent', 'success', 'warning', 'error', 'purple', 'pink', 'cyan', 'slate'
 ]);
-
-// Bucket Record schema
-export const bucketRecordSchema = z.object({
-  id: z.string().guid(),
-  name: z.string().min(1).max(100),
-  color: bucketColorEnum.default('accent'),
-  icon: z.string().max(50).optional().default(''),
-  order: z.number().int().default(0),
-  createdAt: z.string().datetime(),
-  updatedAt: z.string().datetime()
-});
 
 // Create Bucket input schema
 export const bucketInputSchema = z.object({
--- a/server/lib/digitalTwinValidation.js
+++ b/server/lib/digitalTwinValidation.js
@@ -21,15 +21,6 @@ export const documentCategoryEnum = z.enum([
   'creative'        // Aesthetic preferences, creative interests
 ]);
 
-// Test result enum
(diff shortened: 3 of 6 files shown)


atomantic:

Issue #5984, opened 2026-09-03
App Management's PortOS update skips the preflight guards /api/update/execute enforces


atomantic:
PortOS is a managed app, so **Update** on its App Management card runs `appUpdater.updateApp()`, while the Update tab runs `POST /api/update/execute`. Both now launch `update.sh` through `updateExecutor.executeUpdate()` and both take the atomic `setUpdateInProgress(true)` lock (#5976), but only the route runs the preflight refusals around it.

## Missing on the socket path

`server/routes/update.js` (~:240-285), after acquiring the lock, refuses the update when:

1. **A CoS agent is live** — `countActiveCosAgents() > 0` throws `agentsActiveError`. Restarting severs a running agent. No user acknowledgement exists for this one; it is a flat refusal.
2. **Persistent Mind has image-bearing work** — `getPersistentMindImageWorkGuard()`; refused unless the caller passes `acknowledgePersistentMindImageBackup: true`.
3. **The install is running from a fork that is behind upstream** — refused unless the caller passes `acknowledgeFork: true`.

`server/sockets/apps.js` `app:update` runs none of these. Its `appUpdateSchema` (`server/lib/validation.js`) carries only `appId` and `syncFork`, so there is no channel for the two acknowledgements even if the checks were added.

Consequence: a user who updates from App Management rather than the Update tab gets none of the protections — most sharply, a running CoS agent is severed by the restart with no warning.

## Decision

Extract the route body preflight into one reusable guard in `server/services/updateChecker.js` (or a new `server/services/updatePreflight.js`) returning a structured refusal, and call it from **both** entry points. Extend `appUpdateSchema` with the two optional acknowledgement booleans, and surface the refusal in the App Management update UI the same way the Update tab surfaces it — the refusal reason must reach the user, not just abort the socket operation.

Scope this to the PortOS record only (`app.id === PORTOS_APP_ID`); non-PortOS managed apps have no such preconditions.

## Acceptance criteria

- One shared preflight, called by `POST /api/update/execute` and by the `app:update` socket handler for the PortOS app.
- A live CoS agent blocks an App Management PortOS update, with the same message the route produces.
- The persistent-mind-image and fork acknowledgements can be supplied over the socket path and are honored.
- The App Management UI renders the refusal reason and any available acknowledgement.
- Route behavior is unchanged; a regression test asserts both entry points refuse identically.

## Affected files

- `server/routes/update.js` — the preflight block to extract
[...]

Issue #4124, opened 2026-08-14
Gate CoS spawn engines on updateInProgress to close self-update race


atomantic:
## Problem

The self-update flow blocks `POST /api/update/execute` when a CoS agent is live or mid-spawn via `countActiveCosAgents()` in `server/routes/update.js`:

- A fast-fail pre-check at `server/routes/update.js:159-160`.
- After acquiring the update-in-progress lock (`updateChecker.setUpdateInProgress(true)`, `server/routes/update.js:213`), a post-lock re-check at `server/routes/update.js:223-227` — because an agent (e.g. a scheduled/autopilot spawn) may have started live during the git/fork awaits between the fast-fail pre-check and the lock acquisition.

The Update tab also suppresses restart buttons while agents run.

**Residual gap:** an agent that begins spawning *after* the post-lock re-check but *before* `update.sh` actually reaches `pm2 delete` (a multi-second window covering `git pull`, submodule update, and `npm install`) is still severed when the restart happens. The code itself flags this as unresolved — the comment at `server/routes/update.js:221-222` reads:

> "A spawn that begins AFTER this, during update.sh itself, is the residual the PLAN.md spawn-engine gate will close."

Confirmed still open: no spawn engine currently consults `updateChecker` or any `updateInProgress` state (`grep -rln "updateChecker\|updateInProgress" server/services/*.js` only matches `bootstrap.js`, `github.js`, `socket.js`, `updateChecker.js`, and `updateExecutor.js` — none of the spawn engines).

## Fix shape

Close the window from the spawn side instead of the update side:

1. Add a synchronous in-memory `isUpdateInProgress()` mirror to `server/services/updateChecker.js`, kept in lockstep with the persisted flag set by `setUpdateInProgress()` (`server/services/updateChecker.js:293`).
2. Have every CoS spawn engine check it and skip-and-requeue when set — return the existing `SPAWN_DEDUP_SKIP`-style no-op (`server/services/agentGuards.js`) so the task stays queued for after the restart, rather than spawning into a process about to be killed.

## Scope: all spawn engines, not just one chokepoint

Per prior precedent on CoS autonomous-spawn work, gating a single chokepoint gives false confidence — this must cover every path that can start an agent. Audit and gate:

- `spawnAgentForTask` and related entry points in `server/services/agentLifecycle.js`
- `server/services/subAgentSpawner.js`
[...]

PR #5985, opened 2026-09-03
Run PortOS's App Management update through the detached launcher

### LINKED ISSUE #5976: PortOS update from App Management leaves the server dead: update.sh is tree-killed at its own pm2 delete
## What happens

Updating PortOS from the App Management page stops `portos-server` and never brings it back. Reproduced 2026-09-02.

Server log at the moment of the update:

```
⬇️ Socket update started for PortOS
🛑 Received SIGINT - shutting down gracefully (pm_id=…)
✅ DB pool closed
   ← nothing after this; portos-server and portos-cos are gone from `pm2 list`
```

After the failure `pm2 list` still shows `portos-ui`, `portos-autofixer`, `portos-autofixer-ui` and `portos-browser` online with their pre-update uptime — only the **first two** ecosystem entries (`portos-server`, `portos-cos`) are missing. Manual recovery is `pm2 start ecosystem.config.cjs --only portos-server,portos-cos`.

## Root cause

This is the exact "shuts down but never comes back" failure that `spawnDetached` was introduced to fix — but the fix only ever landed on the self-update path, not the managed-app path.

- `server/services/updateExecutor.js` (self-update) launches `update.sh` through `spawnDetached()`'s double-fork specifically so the script reparents to init and survives PM2's TreeKill. Its header comment spells the failure out.
- `server/services/appUpdater.js` `_doUpdate()` (the managed-app socket path, reached from the `app:update` handler in `server/sockets/apps.js`) runs the same `update.sh` through `runCommand()` → `bufferedSpawnOrThrow()` → a plain `spawn()`. That child keeps `portos-server` as its PPID.

PortOS is itself a managed app, so an update triggered from App Management takes the **appUpdater** path. `update.sh:181` runs `pm2 delete ecosystem.config.cjs`; PM2 walks PPID, so it tree-kills `portos-server` → which kills the still-attached `update.sh` → which kills the in-flight `pm2 delete` itself. The script never reaches `update.sh:353` (`pm2 start ecosystem.config.cjs`).

The partial deletion is the fingerprint. `pm2 delete` works through the ecosystem file in declaration order — `portos-server` (`ecosystem.config.cjs:137`) then `portos-cos` (`:213`) — and dies mid-list the moment it removes the server that is its own ancestor. Everything declared after `portos-cos` survives untouched. That is exactly the observed state.

`data/update-detached/` (spawnDetached's control dir) was last written hours earlier, confirming the failing run did not go through `updateExecutor`.

**A no-op update is enough to trigger it.** `data/update.log` for the failing run reads only:

```
🌐 Pulling from origin: <origin>
Already up to date.
Synchronizing submodule url for 'lib/slashdo'
```

The install was already current and still ended up headless.

## Decision — how to fix

**Delegate the PortOS-self case to the existing self-update path rather than duplicating detached-spawn logic in `appUpdater`.**

In `_doUpdate()`, when `app.id === PORTOS_APP_ID` (already imported there from `server/lib/appIdentity.js`), run the `app-update` step via `updateExecutor.executeUpdate()` instead of `runCommand()`. That path already carries the double-fork survival, the `STEP:` progress parsing that maps onto appUpdater's `emit(step, status, message)` contract, the still-running-script guard, and `recordUpdateResult()`.

Also **skip appUpdater's trailing `restart` step for that case** — `update.sh` runs its own `pm2 start` at `:353`, so a `pm2Service.restartApp()` on top of it is redundant and races the script.

Rejected alternative: calling `spawnDetached` directly from `appUpdater` for any app whose `pm2ProcessNames` include the current server's process. It fixes the symptom but leaves two near-identical detached-update implementations to keep in sync, and `updateExecutor` is where the hard-won lifecycle knowledge already lives.

Watch for an import cycle — `updateExecutor` imports `updateChecker`, and `appUpdater` currently imports neither. If a cycle appears, extract the shared launcher rather than inlining a second copy.

## Defense in depth

Independent of the root-cause fix, nothing today notices "I stopped the server and it did not come back," which is what turned a no-op update into hours of downtime.

Add a post-update verification to the update path: after the restart step, poll `/api/system/health` (the loopback HTTP mirror on `:5553` when HTTPS is on) until it returns `ok` or a timeout elapses; on timeout, run `pm2 start ecosystem.config.cjs` and log loudly. A recovery that only fires when the health probe fails cannot make a healthy update worse.

## Acceptance criteria

- Updating PortOS from App Management leaves every `ecosystem.config.cjs` process online, including when the pull is a no-op.
- `update.sh` survives its own `pm2 delete` step on the managed-app path — asserted by verifying the appUpdater PortOS branch goes through the detached launcher, not `bufferedSpawnOrThrow`.
- The `restart` step does not double-restart PortOS on top of `update.sh`'s own `pm2 start`.
- Regression coverage for "PortOS-as-managed-app update does not use the attached spawn path". `server/services/appUpdater.test.js` and `server/services/updateExecutor.test.js` already exist.
- Non-PortOS managed apps keep their current attached-spawn behavior.

## Affected files

- `server/services/appUpdater.js` — `_doUpdate()`, the `app-update` and `restart` steps
- `server/services/updateExecutor.js` — `executeUpdate()`, the reusable detached path
- `server/sockets/apps.js` — the `app:update` handler
- `update.sh` — `:176-182` and `:342-357`, the `pm2 delete` / `pm2 start` bracket
- `server/lib/detachedSpawn.js` — the survival rationale

atomantic:
Updating PortOS from the **App Management** page stopped `portos-server` and never brought it back, even when the pull was a no-op.

## Summary

- **Root cause.** PortOS is itself a managed app, so that button reaches `update.sh` through `appUpdater`'s plain `spawn`, not `routes/update.js`. `update.sh`'s own `pm2 delete ecosystem.config.cjs` tree-kills `portos-server` — and PM2 walks PPID, so it takes the still-attached script down with it, mid-delete and long before the closing `pm2 start`. The fingerprint was the partial deletion: only `portos-server` and `portos-cos`, the two entries declared before the script died, went missing.
- **Fix.** The PortOS record now delegates to `updateExecutor.executeUpdate()`, which already double-forks via `spawnDetached` for the self-update route — rather than keeping a second detached-spawn implementation in sync. `appUpdater` also skips its trailing `restart` step for that case, since `update.sh` starts the ecosystem itself.
- **Scoped narrowly.** Delegation applies only to the PortOS record running *this checkout's* standard script. A custom `updateCommand`, or a `repoPath` that is not this checkout, keeps the attached path so we never silently run a different script than the one configured — and that decision is logged, since the attached path is the one that failed. The path comparison resolves symlinks and case-folds on macOS/Windows, because `repoPath` is user-editable and not force-synced. Non-PortOS managed apps are untouched.
- **Shared update lock.** The delegated launch takes `setUpdateInProgress(true)` — the same atomic lock `POST /api/update/execute` takes, so the two entry points cannot launch `update.sh` concurrently, and the flag CoS spawn gates read (#4124) is now up before the script deletes `portos-cos`.
- **Defense in depth.** Both platform scripts close with a `verify` step that polls `/api/system/health` (new `scripts/verify-server-health.js`) until it reports ok, and on timeout spends one more `pm2 start` before logging the manual recovery command. When it still fails, they print "Update applied, but PortOS is DOWN" and **exit non-zero** — the script outlives the server it restarts, so its exit status is the only signal a caller has. Nothing previously noticed "I stopped the server and it did not come back", which is what turned a no-op update into hours of downtime.

## Test plan

- `server/services/appUpdater.test.js` — 15 tests: PortOS goes through the detached executor and never `bufferedSpawnOrThrow`; the lock is taken before the launch and a second update is refused; the `restart` step is skipped; a differently-spelled `repoPath` still delegates; a custom `updateCommand`, a `repoPath` outside this checkout, and every non-PortOS app keep the attached spawn and their own restart + dashboard handoff; a failed update surfaces instead of reporting success.
- `scripts/verify-server-health.test.js` — 7 tests: URL ordering and dedupe, only a 200 reporting `status: "ok"` counts, a dead port fails, the poll/timeout contract on injected time, and `PORTOS_HEALTH_WAIT_MS` sentinel parsing.
- Full server suite green (38032 passed). Full client suite green (10089 passed). `npm run lint` clean in `client`.
- Shell/PowerShell edits syntax-checked; `update.sh` verified under `bash -n`.

## Deferred

`POST /api/update/execute`'s remaining preflight refusals (live CoS agent, persistent-mind image work, fork acknowledgement) are still absent from the socket path. They need `appUpdateSchema` and UI work to carry the acknowledgements, and are filed as #5984.

Closes #5976


atomantic:

atomantic:
Review status for this run (configured reviewers: `codex,claude`):

- **codex — review-blocked.** Failed on its first invocation with `ERROR: You've hit your usage limit` (account-level; the CLI is healthy and on PATH). No verdict was produced, and no self-review was substituted in its place.
- **claude — clean after 2 rounds.** Round 1 raised 6 findings: the delegated launch not taking the self-update lock (so the #4124 CoS spawn gate stayed inert), an unnormalized `repoPath` compare that silently re-armed the attached spawn, `PORTOS_HEALTH_WAIT_MS=0` collapsing into the 120s default, both scripts printing "Update Complete" and exiting 0 after the verify failed twice, the dashboard handoff being dropped for the paths that still restart PortOS from `appUpdater`, and a test pinning the poll loop's internal pass schedule rather than its timeout contract. All six are fixed in `dddc38a36`. The round-2 convergence check confirmed each fix and raised 2 low findings (a decline log naming the wrong reason, and a docs sentence crediting non-PortOS apps with a dashboard handoff they never had), fixed in `4d29818ea`.

Per this run's required-review rule, the missing codex verdict blocks the merge.

Issue #6043, opened 2026-09-03
Give UI-queued investigation tasks the same identity as auto-filed ones (isInvestigation marker + fingerprint)


atomantic:
The installer-failure "Queue agent to investigate" button (#5981) queues a CoS task with `{ useWorktree: true, openPR: true }`, but the task carries no investigation identity, so it sits outside the machinery that governs every auto-filed investigation.

## Gap

`server/lib/investigationTasks.js` defines what an investigation task IS:

- `INVESTIGATION_TASK_DELIVERY` = `{ useWorktree, openPR, prCompletion: MERGE_ON_GREEN }` — the UI-queued task matches the first two and omits the third.
- `isInvestigationTask()` keys on `metadata.isInvestigation` (with a legacy headline-prefix fallback). A UI-queued task matches neither.
- `buildInvestigationFingerprint()` is the durable dedup key consumed by the reaper (`cosTaskStore.js`) and the retry decision (`services/investigationRetry.js`).

Consequences for a UI-queued investigation:

1. It is invisible to the investigation dedup scan and the circuit breaker, so nothing bounds how many pile up beyond the generic same-description 409.
2. It is not covered by the meta-cascade guard — if the queued run itself fails, `agentErrorAnalysis` can file an auto-investigation *of the investigation*.
3. It does not participate in investigation auto-retry.

`createCosTaskSchema` (`server/lib/cosValidation.js`) has no `isInvestigation` field, so Zod strips any client attempt to set one — this cannot be fixed client-side.

## Scope

- Decide the identity a client-queued investigation gets. Recommended: add a narrowly-typed `isInvestigation` boolean (or a small `investigation: { source, fingerprint }` object) to `createCosTaskSchema`, routed by `cosTaskStore.addTask` into `metadata.isInvestigation` exactly as the server-side producer does — the client supplies the flag, never a hand-crafted fingerprint.
- Derive the fingerprint server-side from the submitted context so a client cannot collide with, or evict, an auto-filed investigation.
- Have `QueueInstallInvestigationButton` (`client/src/components/install/QueueInstallInvestigationButton.jsx`) set the flag, and reconcile its `INSTALL_INVESTIGATION_DELIVERY` constant with `INVESTIGATION_TASK_DELIVERY` — including whether a user-initiated install investigation should also carry `prCompletion: merge-on-green` (a UI-queued repair is user-initiated, so review-then-merge may be the better default; make the call and document it).
[...]

PR #6044, opened 2026-09-03
Add a "Queue agent to investigate" button to installer failure UI

### LINKED ISSUE #5981: Add a "Queue agent to investigate" button to installer/setup failure UI
## Problem / Goal

When an in-app installer (FLUX.2 runtime, video-gen/local-model BYO runtimes, MIDI, local setup) hits an error, the user's only option today is to dismiss it. There's no way to hand the failure to PortOS's own CoS agent system to investigate the root cause and fix it — even though that queueing mechanism already exists and is used elsewhere in the app.

## Context

- Every installer failure flows through the shared `useInstallStream` hook (`client/src/hooks/useInstallStream.js`), which surfaces a single `error: string|null`. Three call sites read it directly: `client/src/components/imageGen/Flux2InstallModal.jsx`, `client/src/components/install/RuntimeInstallModal.jsx`, and `client/src/components/settings/LocalSetupPanel.jsx:90`.
- `RuntimeInstallModal.jsx` alone backs 6 other call sites (`MusicGenPanel.jsx:759`, `Image3dRuntimes.jsx:202`, `AIProviders.jsx:1006` and `:1028`, `VideoGen.jsx:1558`, plus `MidiInstallModal.jsx`'s thin wrapper), so it and `Flux2InstallModal.jsx` between them cover every install-failure surface in the app today.
- Both modals currently render their own bespoke error footer with only a **Close** button (`Flux2InstallModal.jsx:209-221`, `RuntimeInstallModal.jsx:146-157`) — no shared error-footer component exists to extend once.
- PortOS already has a full "queue an autonomous agent task" primitive that fits this: `POST /api/cos/tasks` (`server/routes/cosTaskRoutes.js:405-420`) → `addTask()` (`server/services/cosTaskStore.js:260`), validated by `createCosTaskSchema` (`server/lib/cosValidation.js:125-251`), reachable client-side via `addCosTask(task)` (`client/src/services/apiAgents.js:118-122`). Relevant fields: `description` (required), `prompt` (full agent-facing task text), `app` (target repo — omitting it targets PortOS itself via `PORTOS_APP_ID`, which is correct here since the installer code lives in this repo), `planOnly` (investigate + file an issue without shipping code), `useWorktree`/`openPR` (normal delivery flow).
- `client/src/components/cos/TaskAddForm.jsx` is the existing full manual "Add Task" composer that calls `addCosTask` — a useful wiring reference, but it's a general-purpose form, not something reactive to a failure.
- There is no existing "file a GitHub issue directly from a UI button" path (`server/services/github.js`'s `execGh` is repo/secrets-only) — issue filing today happens *inside* an agent's own run (e.g. a `planOnly` CoS task), so this button should queue a CoS task rather than try to create a tracker issue directly.

## Proposed approach

1. Extract a small shared error-footer piece (e.g. an `InstallErrorFooter` component, or a shared render helper) used by both `Flux2InstallModal.jsx` and `RuntimeInstallModal.jsx` when `error` is set — today each hand-rolls its own footer, so this is also the point that stops the duplication from growing to a third copy.
2. Add a "Queue agent to investigate" action to that shared footer. On click, call `addCosTask` with:
   - `description`: a short summary derived from the failing stage/runtime name (e.g. "Fix FLUX.2 installer failure at the venv stage").
   - `prompt`: the failing stage, the accumulated install log tail, and the runtime/label identifying which installer failed — enough for an agent to reproduce without the user re-typing anything.
   - `app`: omitted, so it defaults to PortOS itself.
   - Sensible defaults for the rest (mirror `TaskAddForm`'s per-app defaults for `useWorktree`/`openPR`) — no need to expose the full Agent Ops drawer for a one-click action.
3. Give the button a clear pending/queued state (disable + label change) and a toast on success/failure, per `client/src/AGENTS.md`'s "silent vs. toasting" convention (a state-changing action, not a plain fetch — don't make it silent).
4. Wire the same action into `LocalSetupPanel.jsx`'s bespoke error rendering, since it consumes `useInstallStream` directly rather than through either modal.
5. Add component test coverage for the new button (client `*.test.jsx` alongside the modal) asserting it calls `addCosTask` with the failing context and reflects a queued state.

## Acceptance criteria

- [ ] `Flux2InstallModal.jsx`, `RuntimeInstallModal.jsx` (covering its 6 call sites), and `LocalSetupPanel.jsx` each show a "Queue agent to investigate" action alongside Close whenever `useInstallStream`'s `error` is set.
- [ ] The queued task's `prompt` includes the failing stage/runtime and the install log tail, without requiring the user to re-describe the failure.
- [ ] The task targets the PortOS app itself by default (no manual app picker required for this shortcut).
- [ ] The button shows a queued/pending state after a successful `addCosTask` call and surfaces a toast on failure.
- [ ] New tests cover the button wiring in at least one of the two modals.

## Out of scope

- Extending this pattern to non-install failure surfaces (toasts, background job failures, `ErrorBoundary` render crashes) — this issue covers only the `useInstallStream`-based installer failure UI.
- A direct "file a GitHub issue" action independent of the CoS agent queue — PortOS has no such primitive today, and the queued CoS task (optionally `planOnly`) already covers "investigate and open an issue" if the agent decides that's the right outcome.
- Any redesign of the CoS task queue, the `addCosTask` payload shape, or `TaskAddForm` itself.

atomantic:
Closes #5981

## Summary

- An installer failure used to be a dead end — **Close** was the only affordance. Every `useInstallStream`-backed failure surface now also offers **Queue agent to investigate**, which posts a CoS task carrying the installer name, the failing stage, the error, and a bounded tail of the streamed install log, so an agent can work the failure without the user re-describing it.
- The task omits `app`, which targets PortOS itself — where the installer code lives — and ships with `useWorktree` + `openPR`.
- Both install modals hand-rolled the same "message + Close" error footer, so the action goes into a shared `InstallErrorFooter` rather than making a third copy. `LocalSetupPanel` draws its own error region and renders the button directly. That covers `Flux2InstallModal`, `RuntimeInstallModal` (and its six call sites), and `LocalSetupPanel`.
- Log tail is redacted and bounded before it leaves the browser: home-directory paths pip/git echo become `/Users/<user>`, a log line of its own backticks can no longer close the prompt fence, the char cap cuts on a line boundary (never mid-surrogate), and the fenced block is explicitly labelled as untrusted third-party output the agent must not paste into a commit, PR, or issue. The error string goes through the same path, since `useInstallStream` copies it into the log array too.
- A repeat click after the same install fails again hits the task store's 409 `DUPLICATE_TASK` guard. That fires for an existing **pending or blocked** task, so the button distinguishes the two and surfaces the server's own wording instead of claiming an agent is on it.

## Test plan

- `client/src/lib/installFailureTask.test.js` — log-tail bounds (line + char caps, line-boundary cut, mid-pair surrogate guard), home-path redaction and fence neutralization, description/prompt shape including the empty-error fallback.
- `client/src/components/install/RuntimeInstallModal.test.jsx` — the button appears only on failure; queueing posts the failing stage/error/log tail with no `app`, `{ useWorktree: true, openPR: true }` and `{ silent: true }`; queued state disables; failure toasts and stays clickable; the 409 reads as an existing task; Close still dismisses.
- `client/src/components/imageGen/Flux2InstallModal.test.jsx` and `client/src/components/settings/LocalSetupPanel.test.jsx` — the other two surfaces render the action on failure and not otherwise.
- Full client suite green (`cd client && npm test`): 831 files, 10142 tests. `biome lint --error-on-warnings src` clean.

## Follow-up

The queued task carries no `isInvestigation` marker, so it sits outside the investigation dedup / circuit-breaker machinery. That needs a `createCosTaskSchema` field and is tracked in #6043.


atomantic:

atomantic:
Review audit trail for this run (configured reviewers: `codex,claude`):

- **codex — BLOCKED, no verdict.** Failed on its first invocation with an account-level usage limit (`You've hit your usage limit… try again at Sep 7th, 2026`). No self-review was substituted.
- **claude — ran, 3 rounds, converged APPROVE.** Round 1 approved with 7 nits; round 2 (convergence check on the applied delta) returned CHANGES_REQUESTED with one blocking finding — the error string was reprinted unredacted above the redacted log fence — plus a vacuous surrogate test and a duplicate-409 that mislabelled a *blocked* task as queued; round 3 approved the fixes.

All of claude's findings are fixed in commits 6571dfa, cc67b0a and 37f3654. The one deferred finding (no `isInvestigation` marker on a client-queued task — it needs a server schema field) is filed as #6043.

Merge is withheld pending the codex review.

PR #6047, opened 2026-09-03
restart PortOS's PM2 apps when an update aborts between the delete and the start

atomantic:
## Summary

Updating PortOS from App Management could still delete `portos-server` and leave it down. #5976 fixed the update script being **tree-killed** by its own `pm2 delete`; it did not cover the script **exiting on its own**.

`update.sh` / `update.ps1` delete PortOS's PM2 entries at the `pm2-stop` step and don't start them again until the `restart` step ~170 lines later. Under `set -euo pipefail` (and each explicit `exit` on Windows), *any* failure in between — `npm install`, `setup-db`, migrations, the client build — exits with `portos-server` deleted and nothing left running to notice. The install stays headless until someone runs `pm2 start` by hand.

- **Trap the exit** (bash `EXIT` + `TERM`/`INT`/`HUP`) / **route every fatal exit through `Stop-UpdateScript`** (PowerShell), and restart the apps on the way out.
- **Confirm with `verify-server-health.js`**, the same probe the happy path uses — the recovery starts a half-installed tree, so a `pm2 start` that exits 0 and then crash-loops is the likely case, not the edge case. It never claims a recovery the probe didn't confirm, and the update still reports failure.
- **Resolve pm2 rather than hardcoding it.** `safe_install .` wipes root `node_modules` whenever the pull touched root `package.json` — every release, since the version bump lives there — and pm2 is a root dependency. On the most likely failure of all (both `npm install` attempts fail) the checkout's own pm2 is already gone. Falls back local → PATH → `npx` at the pinned version, and prints whichever one exists.
- **PowerShell:** the guard block sits above every fatal exit in the file, because `exit` inside a function ends the script without a terminating error — bypassing both `Stop-UpdateScript` and the `trap`. `Safe-Install`'s exit is defined above the delete but runs below it, so a guard placed beside the delete could never have caught it.
- **Arm the latch before the delete.** Bash runs a pending signal trap only between statements, so a `TERM` during `pm2 delete` would otherwise reach the trap with the latch unset.

## Test plan

`scripts/update-headless-recovery.test.js` runs the **real `update.sh`** in sandboxed git repos (shared `makeGitSandbox` helper) with `npm`/`npx`/`pm2` shimmed — assertions are on the recorded pm2 call sequence, not a grep for the guard's source:

- fails after the delete → a `start` follows the `delete`, exit status stays non-zero, `STEP:restart:warning:`
- succeeds → exactly **one** `start`, no recovery
- fails before the delete → PM2 untouched
- health probe fails → `STEP:restart:error:`, never a false success
- `npm install` fails after wiping root `node_modules` → recovery still starts pm2 via the PATH fallback
- `SIGTERM` delivered *during* the delete → recovery still runs, exit 143

`update.ps1` can't be executed here, so its guard is pinned structurally: every `exit` in the whole file must be one of the two sanctioned ones, the latch must be armed before the delete, and `Resolve-Pm2Command`'s result must be array-forced.

Each guard was verified by reverting it and confirming the matching test fails. Full server (38k) and client (10k) suites pass.

## Known, not fixed here

`updateExecutor` labels the failure `Update failed at step "restart"` because the recovery emits the last `STEP:` line. It has no live consumer — the server is already deleted whenever the recovery runs — and the fix belongs in `updateExecutor`, outside this diff.

### DIFF
--- a/scripts/update-headless-recovery.test.js
+++ b/scripts/update-headless-recovery.test.js
@@ -94,6 +94,12 @@ if (${stallDelete} && args.startsWith('delete')) {
     writeFileSync(join(bin, shim), `#!/bin/sh\n${npmBody}\n`);
     chmodSync(join(bin, shim), 0o755);
   }
+  // The success case is the only one that runs past trusted-rebuilds, so it
+  // reaches update.sh's ffmpeg step — which on a machine without ffmpeg would
+  // run a real `brew install`, or on Linux CI a passwordless `sudo apt-get
+  // install` that mutates the runner. update.sh only probes `command -v`.
+  writeFileSync(join(bin, 'ffmpeg'), '#!/bin/sh\nexit 0\n');
+  chmodSync(join(bin, 'ffmpeg'), 0o755);
   // A pm2 the recovery can still reach after safe_install wipes the local one.
   if (pm2OnPath) {
     writeFileSync(join(bin, 'pm2'), `#!/bin/sh\nprintf '%s\\n' "$*" >> ${JSON.stringify(calls)}\n`);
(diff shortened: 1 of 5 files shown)


atomantic:
Note: this branch also fixes a break inherited from `main` — `recordedProjectDir.js` imports `PORTOS_ENV_PATH` from `portosEnv.js` for its own defaults but never re-exports it, while `vllmQwenProject.js` and `sglangQwenProject.js` import it *from* `recordedProjectDir.js`. That is a load-time `SyntaxError` failing the server and Windows jobs on every PR right now, with no fix in flight, so it is rolled in here rather than left blocking. A re-export would collide in `lib/index.js`, which `export *`s both modules.

PR #6048, opened 2026-09-03
fix: paint the Video Gen model picker without waiting on the python probe (#5835)

### LINKED ISSUE #5835: Serve the Video Gen model list off /video-gen/models instead of the python-probe /status payload
## Problem

`GET /api/video-gen/status` welds three unrelated things into one response: python health (a subprocess probe, ~1–2s), the hardware-aware model list, and the model-shaping numbers (`defaultModel`, `systemMemoryGb`, `fflfLtx2PixelBudget`). The Video Gen page needs the model list to render its Model picker, so the picker waits on the python probe.

A first fix landed in `client/src/pages/VideoGen.jsx` + `client/src/lib/videoGenStatusCache.js`: the field now holds its place with a loading placeholder, and a session cache paints the previous model list immediately on a revisit. That covers the second-and-later loads, but a **cold** load (new tab / new session) still waits on the probe before any model appears.

`GET /api/video-gen/models` (`server/routes/videoGen.js`, `router.get('/models')`) already returns `hardwareAwareVideoModels()` with **no** python probe — and `listVideoModels()` in `client/src/services/apiImageVideo.js` already wraps it with the same hardware filter. The sibling ImageGen page is already shaped this way (`listImageModels()` from its own mount effect, separate from `refreshStatus()`), which is why it does not have this problem.

## Work

1. Add the model-shaping fields the picker's auto-select needs to the `/video-gen/models` response — `defaultModel`, `systemMemoryGb`, and `fflfLtx2PixelBudget` — either as a wrapper object (breaking the current bare-array shape, so version-gate it) or as a sibling `GET /video-gen/model-context`. Keep `/status` returning them too; other callers read them there.
2. In `VideoGen.jsx`, fetch that alongside `getVideoGenStatus()` on mount and let it drive `models` + the auto-select inputs. `/status` keeps owning connectivity only.
3. Once the model list no longer rides the probe, re-evaluate `client/src/lib/videoGenStatusCache.js`: if the decoupled fetch is fast enough that the picker paints immediately, delete the cache, its barrel/README rows, the `stale` flag and the `statusFresh` gate in `VideoGen.jsx`, and the two cache-specific cases in `client/src/pages/VideoGen.modelLoading.test.jsx`. Keep the loading-placeholder case either way.

## Decisions already made

- **Do NOT cache `resolveLocalPythonHealth` server-side.** The route comment states the invariant: it probes package imports on every call so a half-installed python cannot masquerade as connected.
- Keep the `loading` prop on `client/src/components/ModelSelect.jsx` regardless — a slow list is a property of the shared picker, and the placeholder is what keeps the form from jumping.

atomantic:
## Summary

Opening Video Gen in a new tab left the Model field on a "Loading models…" placeholder for a second or two before the picker appeared. The list rode along on `GET /api/video-gen/status`, which shells out to python on every call to check package imports — so the form waited on an interpreter probe it did not need in order to render a dropdown. (A session cache shipped earlier covered the *second* visit; a cold load still waited.)

**Server.** New probe-free `GET /api/video-gen/model-context` returns `{ models, defaultModel, systemMemoryGb, fflfLtx2PixelBudget }` — the model list plus the three numbers the picker's auto-select reads — off the registry and the already-cached hardware probe alone. It is a sibling route rather than a change to `/models`, so that route's long-standing bare-array shape stays intact for its existing callers and for an older client talking to a newer server. `/status` keeps returning the same fields for its other readers, and both routes build them from one shared `videoModelContext()` helper so the two cannot drift.

**Client.** `VideoGen.jsx` fetches the new route alongside `getVideoGenStatus()` on mount; `/status` is now responsible for connectivity alone. `useVideoGenForm` takes a `modelContext` argument instead of `status` — it only ever read those three fields — which makes the split structural rather than a convention.

**Cleanup.** With the list no longer behind the probe, `client/src/lib/videoGenStatusCache.js` has no work left to do and is deleted, along with its barrel/README rows, its `stale` flag, and the `statusFresh` gate the connectivity UI needed so a stored answer could never report python health.

Closes #5835

## Test plan

- `server/routes/videoGen.test.js` — two new cases on `GET /model-context`: it serves the full field set and the python package probe is **never called** in that request path; and every field it shares with `/status` is byte-equal, so the shared builder is pinned.
- `client/src/pages/VideoGen.modelLoading.test.jsx` — reworked. The two cache-specific cases are gone; a cold-load case asserts the picker is enabled and holding the default model *while* `/status` is still in flight and the banner still reads "Checking…", and a new case asserts a failed context fetch takes the Model field away rather than parking on the placeholder forever. Verified these fail against the pre-fix wiring (2 of 3 red) and pass after.
- Four sibling VideoGen page suites updated for the new mount fetch; the runtime-install case now asserts **both** halves refresh, since an install moves the hardware decoration and python health.
- Full suites green: client 10,124 passed / 2 skipped; server 38,296 passed / 14 skipped. `apiRouteCatalog.generated.json` regenerated for the new route.

## Review status

⚠️ The required local reviewer for this run could not produce a verdict — see the comment below.

https://claude.ai/code/session_012krsCGMnjFmoeY9uyi4Ypf


atomantic:

atomantic:
Reviewer failure detail, for whoever re-runs it: the pinned local reviewer (`ollama`, model `…gemma-4-12B-coder…:Q4_K_M`, effort `low`) returns HTTP 400 from the Ollama API — `"<model>" does not support thinking`. Reproduced twice against the same diff via `server/scripts/run-local-code-review.mjs`; no verdict was produced, so no self-review was substituted. The effort level maps to a thinking parameter this GGUF build does not accept. Both full test suites are green (client 10,124 / server 38,296) and the branch is otherwise ready.

Issue #6050, opened 2026-09-03
Local reviewer sends reasoning_effort to every ollama model, 400-ing on models without thinking support


atomantic:
## Problem

`runToolFreeLocalCompletion` in `server/services/codeReview.js` attaches `reasoning_effort` to the OpenAI-compatible request whenever `normalizeReviewerEffort(effort, backend)` returns a level:

```js
const resolvedEffort = normalizeReviewerEffort(effort, backend) || null
...
...(resolvedEffort ? { reasoning_effort: resolvedEffort } : {}),
```

That gate is keyed on the **backend**, not on the **model**. Ollama translates `reasoning_effort` into its `thinking` parameter and hard-rejects it for a model that does not advertise thinking support:

```
ollama API error 400: {"error":{"message":"\"<model>\" does not support thinking",
"type":"invalid_request_error","param":null,"code":null}}
```

So a claim/PR run pinned to `ollama[<non-thinking model>]~effort=low` can never get a verdict: `run-local-code-review.mjs` exits 1 with `ok:false`, and a REQUIRED reviewer configured that way makes every run `review-blocked`, leaving PRs open and unmergeable. Reproduced deterministically (two consecutive attempts) on a real claim run against a gemma-family coder GGUF; PR #6049 is the PR it blocked.

The doc comment on `runLocalCodeReview` already states the intended behavior — "Omitted from the body entirely when unset or **not a level this backend accepts** — a non-reasoning [model] …" — so this is the implementation not matching its own documented contract, and the contract itself should say *model*, not backend.

Note the failure is easy to miss in the claim-comment path: `runLocalClaimCommentReview` returns early on an empty comment list before it ever reaches the API, so the tool-free comment gate passes with the same model/effort that then 400s on the code review.

## Work

- Resolve thinking support **per model** before attaching `reasoning_effort`. PortOS already knows this: `server/services/localLlm.js` maps the ollama capability `thinking` -> `reasoning` (see the capability map around line 580), so the model list carries the answer.
- When the resolved model does not support it, drop `reasoning_effort` from the body and proceed with the review rather than failing — a non-reasoning model still returns usable findings.
- Fail closed only when the capability genuinely cannot be resolved AND the request still errors; keep the existing `ok:false` shape for that case so `review-blocked` still works.
- Correct the `runLocalCodeReview` doc comment to say the omission is decided per model.

## Acceptance criteria

[...]

PR #6052, opened 2026-09-03
test: give Python-shelling suites explicit, nested timeouts (#5856)

### LINKED ISSUE #5856: Python-shelling normal-bake tests time out under full-suite load (vitest 10s budget vs a 240s subprocess allowance)
## Problem

`server/services/imageTo3d/trellis2NormalBake.test.js` fails a full-suite `npm test` with a runner timeout, and passes in isolation:

```
FAIL  services/imageTo3d/trellis2NormalBake.test.js > bake_normal_map attaches a usable normal map
      > attaches a normalTexture of the requested size, with real tilt
Error: Test timed out in 10000ms.
 ❯ services/imageTo3d/trellis2NormalBake.test.js:312:3

# same file, run alone:
$ npx vitest run services/imageTo3d/trellis2NormalBake.test.js
Test Files  1 passed (1)      Tests  14 passed (14)      Duration  4.22s
```

There is no failing assertion — this is a runner budget, the same class the `hookTimeout` comment in `server/vitest.config.js:30-38` already documents for the image-gen suites.

## Root cause

These tests shell out to a real Python interpreter with `execFileSync` and wait synchronously: they write a script to a temp dir, then run numpy + trimesh + PIL through it (the failing case bakes a 64x64 normal map). The cost is a real subprocess whose wall time scales with how loaded the machine is, not with anything the test controls. Alone it is ~4s; on a contended worker during a 1,848-file run it crosses the 10s `testTimeout`.

**The two budgets in this file already contradict each other.** The e2e `run` helper at line 269 passes `execFileSync(pyBin, [script], { encoding: 'utf8', timeout: 240000 })` — a deliberate 4-minute allowance for the subprocess — while vitest kills the test at 10s. The `execFileSync` timeout is dead intent today: vitest always wins. The other `run` helper (line 62) sets no subprocess timeout at all.

## Decision (already made — do not re-litigate)

**Give the Python-shelling tests an explicit per-test vitest timeout rather than raising the global `testTimeout`.** The global default is deliberately tight (it catches genuinely hung async work across ~37.5k tests); loosening it repo-wide to accommodate a handful of subprocess suites would blunt that everywhere. The narrow fix states the real budget where the real cost is.

Do **not** "fix" this by mocking Python. These tests exist specifically to run the actual bake and assert numeric properties of the output (unit-length normals, mean z, tilt fraction) — the arithmetic is the thing under test, and a mock would make the suite vacuous.

## Scope

1. `server/services/imageTo3d/trellis2NormalBake.test.js` — pass an explicit timeout as `it()`'s third argument on every test that calls a `run` helper (both `describe` blocks). Use a single named constant, e.g. `const PY_TEST_TIMEOUT_MS = 120_000;`, with a one-line comment saying why (real Python subprocess; wall time scales with machine load, not with the assertion). Keep it at or under the existing `execFileSync` 240s allowance so the subprocess guard still fires first and produces a useful error instead of a bare vitest timeout.
2. Add the same `{ timeout: ... }` to the first `run` helper's `execFileSync` (line 62), which currently has none — an actually-hung interpreter there hangs until vitest kills the worker.
3. Sweep the sibling Python-shelling suites for the same shape and apply the same constant where a test genuinely runs an interpreter (not merely resolves a path). Candidates to check — confirm each actually spawns before touching it:
   `server/services/loraEffectProbe.test.js`, `server/services/audioMidiTranscription.test.js`, `server/services/pipeline/musicGen.test.js`, `server/lib/pythonSetup.test.js`, `server/lib/setupScriptRunner.test.js`.

## Acceptance criteria

- `cd server && npm test` completes with zero timeout failures across three consecutive runs on a loaded machine.
- The file still passes in isolation, and still asserts the numeric bake properties it asserts today — no mocking of the Python side.
- Global `testTimeout` / `hookTimeout` in `server/vitest.config.js` are unchanged.
- Every Python-spawning `execFileSync` in the touched files carries its own subprocess `timeout`.

atomantic:
## Summary

`server/services/imageTo3d/trellis2NormalBake.test.js` failed a full-suite `npm test` with a runner timeout and passed in isolation. There was no failing assertion — these tests shell out to a real Python interpreter, so their wall time tracks how loaded the machine is rather than anything the assertion controls, and a file that runs in ~4s alone crosses the global 10s `testTimeout` on a contended worker.

Two **nested** budgets now live beside `resolveTestPython` in `server/lib/testHelper.js`:

| constant | value | applied as |
| --- | --- | --- |
| `PY_TEST_TIMEOUT_MS` | 120s | `it()`'s third argument on every case that spawns an interpreter |
| `PY_SUBPROCESS_TIMEOUT_MS` | 90s | every `execFileSync` spawn's own `timeout` |

The subprocess budget is deliberately **below** the vitest budget so a genuinely hung interpreter trips the spawn guard first and fails with an `ETIMEDOUT` naming the command, instead of a bare vitest timeout that says only that the test ran long.

That ordering is the substantive fix to the e2e bake helper, which carried a 240s subprocess allowance that vitest's 10s budget always won — dead intent, as the issue notes. The remaining spawn sites had no subprocess timeout at all and could hang until the worker was killed:

- the numpy and Metal-stack probes in `trellis2NormalBake.test.js`
- both `run` helpers in `trellis2NormalBake.test.js`
- the `run` helper, the `patch` helper, and the non-UTF-8 codec case in `trellis2GenerateRunner.test.js` (the sibling suite the sweep turned up — the five candidates the issue listed were checked and none actually spawn an interpreter)
- `resolveTestPython`'s own `python -c pass` probe

Nothing on the Python side is mocked: these tests exist to run the actual bake and assert numeric properties of the output (unit-length normals, mean z, tilt fraction), and a mock would make them vacuous. Global `testTimeout` / `hookTimeout` in `server/vitest.config.js` are unchanged.

## Test plan

- `cd server && npm test` — **1895 test files passed** / 1 skipped, **38294 tests passed** / 14 skipped, exit 0, and zero `Test timed out` / `Hook timed out` occurrences in the log.
- Touched suites in isolation: `trellis2NormalBake.test.js`, `trellis2GenerateRunner.test.js`, and the `server/lib/index.test.js` barrel/README guard all pass.
- Confirmed the new per-test argument is actually honored rather than inert: temporarily setting `PY_TEST_TIMEOUT_MS = 1` makes the bake suite fail with `Test timed out in 1ms` (previously it would have reported the global `10000ms`). Reverted.

## Review status

The required local reviewer could not produce a verdict. `ollama` returns HTTP 400 for the pinned model on every attempt (retried):

> `"hf.co/…gemma-4-12B-coder…" does not support thinking`

That is a provider-side failure of the reviewer, not a finding on this diff, and it is the exact defect already tracked in #6050 (local reviewer sends `reasoning_effort` to every ollama model). Per the required-review publication rule this PR is published but left **unmerged** pending that review; no self-review was substituted for it.

Closes #5856


atomantic:

PR #6055, opened 2026-09-03
fix: ignore Finder metadata in build freshness

tzioup:
## Summary

- exclude macOS Finder .DS_Store metadata from the client build-freshness scan at the client root and below src/public
- keep all other hidden and untracked files in scope, because files such as Vite .env inputs can genuinely require a rebuild
- document the separate managed-updater build omission fixed by #5823 and the feature-branch Reconcile behavior

## Why

The current detector walks raw filesystem entries. Finder can create .DS_Store after a successful build, leaving staleBuild as the only active install-state signal even though no build input changed. PortOS already treats this filename as macOS metadata in Git and backup handling.

## Verification

- focused install-state suite: 36 passed
- full server suite: 38,273 passed; 9 environment-sensitive checks failed under the restricted runner
- the four affected process, detached-spawn, file-watch, and voice-model test files were rerun with normal macOS permissions: 268 passed

### DIFF
--- a/docs/SELF_UPDATE.md
+++ b/docs/SELF_UPDATE.md
@@ -58,6 +58,8 @@ Eidoverse checkouts to apply a PortOS world design.
 
 `GET /api/update/status` also compares recursive submodule checkouts with their pinned revisions. An uninitialized, conflicted, behind, or divergent module marks the install out of sync, making the existing Reconcile control available even when no newer release is waiting. A checkout deliberately advanced through the Submodules tab is not treated as stale, and CoS worktrees report submodule state as unknown because they intentionally leave submodules uninitialized and cannot run the primary-checkout update flow.
 
+For the point-in-time analysis of macOS Finder metadata false positives and the separate managed-App build omission, see [macOS “Install out of sync” research](research/2026-09-03-macos-install-out-of-sync.md).
+
 To prevent that confusion, `POST /api/update/execute` rejects fork runs with **412 `FORK_SYNC_REQUIRED`** unless either:
 
 - the request body sets `acknowledgeFork: true`, or
--- a/docs/research/2026-09-03-macos-install-out-of-sync.md
+++ b/docs/research/2026-09-03-macos-install-out-of-sync.md
@@ -1,0 +1,97 @@
+# macOS “Install out of sync” research
+
+Point-in-time primary-source review on 2026-09-03. This note distinguishes the
+documented install/update path from two separate stale-build defects.
+
+## Bottom line
+
+- A fresh, minimal supported install is `npm run setup` followed by `npm start`.
+  `npm start` builds the client before starting PM2. `./setup.sh` is the optional
+  guided alternative, not an additional required step
+  ([README.md:291-309](../../README.md#L291-L309)).
+- A managed reconcile is supposed to run the full platform update script. That
+  script installs all workspaces, runs setup and migrations, builds the client,
+  and restarts PortOS ([update.sh:185-219](../../update.sh#L185-L219),
+  [update.sh:293-317](../../update.sh#L293-L317),
+  [update.sh:335-340](../../update.sh#L335-L340)).
+- There was a real, platform-independent bug in **Apps → PortOS → Git → Update
+  app**: it restarted without rebuilding `client/dist`. It was fixed upstream on
+  2026-09-02 by [PR #5823](https://github.com/atomantic/PortOS/pull/5823).
+- `.DS_Store` was already recognized elsewhere as macOS metadata: it is ignored
+  by Git ([.gitignore:39-41](../../.gitignore#L39-L41)) and explicitly excluded
+  from backup snapshot enumeration
+  ([server/services/backup.js:522-532](../../server/services/backup.js#L522-L532)).
+  However, upstream `installState` freshness detection does not consult Git and
+  has no `.DS_Store` exclusion as of upstream commit `878468a9c`.
+
+## What the freshness detector actually measures
+
+Issue [#1779](https://github.com/atomantic/PortOS/issues/1779) introduced the
+warning on 2026-06-29 to detect a half-update after `git pull` without
+`./update.sh`. Its five signals are running-code drift, dependency receipts,
+client build time, pending migrations, and submodule drift
+([server/services/installState.js:1-28](../../server/services/installState.js#L1-L28)).
+
+The root-file scan was deliberately widened in commit
+[`4147b077f`](https://github.com/atomantic/PortOS/commit/4147b077f13b95f0ad2d68a989c05678ff2d18c8)
+from three named files to **every file directly under `client/`**, while also
+walking `src/` and `public/`. The current upstream implementation uses raw
+`readdir`/`stat`; it excludes only the `node_modules`, `dist`, and `.git`
+directories. Consequently `.gitignore` has no effect on this calculation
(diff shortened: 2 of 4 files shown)


docs/research/2026-09-03-macos-install-out-of-sync.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Bottom line
+
+- A fresh, minimal supported install is `npm run setup` followed by `npm start`.
+  `npm start` builds the client before starting PM2. `./setup.sh` is the optional
+  guided alternative, not an additional required step
+  ([README.md:291-309](../../README.md#L291-L309)).
+- A managed reconcile is supposed to run the full platform update script. That
+  script installs all workspaces, runs setup and migrations, builds the client,
+  and restarts PortOS ([update.sh:185-219](../../update.sh#L185-L219),
```
atomantic, line 14: 💡 **Non-blocking** — update.sh line anchors point at unrelated code

The claim that update.sh installs all workspaces, runs setup and migrations, builds the client, and restarts PortOS is correct, but none of the three cited ranges support it. update.sh:185-219 is the PM2 recovery trap and its pm2 resolution helper, and update.sh:335-340 is the Linux ffmpeg sudo branch; only update.sh:293-317 partially matches, and only for the setup scripts. A reader following these anchors to verify the claim lands on unrelated code and cannot confirm it. In the reviewed tree the real locations are the safe_install calls at update.sh:271-275, the migration run at update.sh:380-381, the client build at update.sh:400-402, and the pm2 start at update.sh:426-444. Consider citing those, or dropping the line ranges in favor of file-level references, since line anchors in a note that is explicitly not maintained afterwards will rot on the first unrelated edit to update.sh.

docs/SELF_UPDATE.md
```diff
@@ -58,6 +58,8 @@ Eidoverse checkouts to apply a PortOS world design.
 
 `GET /api/update/status` also compares recursive submodule checkouts with their pinned revisions. An uninitialized, conflicted, behind, or divergent module marks the install out of sync, making the existing Reconcile control available even when no newer release is waiting. A checkout deliberately advanced through the Submodules tab is not treated as stale, and CoS worktrees report submodule state as unknown because they intentionally leave submodules uninitialized and cannot run the primary-checkout update flow.
 
+For the point-in-time analysis of macOS Finder metadata false positives and the separate managed-App build omission, see [macOS “Install out of sync” research](research/2026-09-03-macos-install-out-of-sync.md).
```
atomantic, line 61: 💡 **Non-blocking** — Cross-reference widens an existing antecedent gap

The paragraph immediately after this insertion opens with To prevent that confusion, whose antecedent is the fork silent no-op pull sentence far above at docs/SELF_UPDATE.md:19. That distance was already large before this change, and inserting an unrelated macOS-metadata pointer directly in front of it adds one more intervening paragraph, so a reader arriving at that sentence is now even more likely to read it as referring to Finder metadata. Placing the pointer at the end of the update-mechanics section, or immediately after the stale-build discussion it actually relates to, would keep the fork paragraph adjacent to what it refers back to.

atomantic (approved):
✅ **Approved**

The patch applies cleanly to current main and the fix is correct and minimal: a two-line exclusion of .DS_Store from the client build-freshness walk, with a regression test that genuinely pins it. A full server-suite run against the patched tree and against the unpatched baseline differ by exactly one test, the new passing one, so the change introduces no regression.

**Scope:** One narrow server behavior change in the install-freshness scan, its regression test, and two documentation files.

#### 💡 Non-blocking (2)
- `docs/research/2026-09-03-macos-install-out-of-sync.md:14` — update.sh line anchors point at unrelated code
- `docs/SELF_UPDATE.md:61` — Cross-reference widens an existing antecedent gap

#### Test evidence
- ✅ `git apply --check -- .portos-public-review/PR-6055.patch` — Applies cleanly at base commit 9bb4b572f with no fuzz or rejects.
- ✅ `npx vitest run services/installState.test.js (from server/)` — 36 of 36 tests passed, matching the count the PR description claims for the focused install-state suite.
- ❌ `npx vitest run services/installState.test.js -t "Finder metadata" with both CLIENT_NON_BUILD_FILES guards manually removed from the source` — Deliberate bypass probe on a temporarily reverted source, not a defect. The new test failed with expected true to be false at installState.test.js:425, which confirms it actually pins the fix rather than passing vacuously. The source was restored byte-for-byte afterwards and the diff re-verified against the patch.
- ❌ `npx vitest run --reporter=json (full server suite, patched)` — 38329 tests, 34770 passed, 3533 failed across 186 files. Every failure is environmental in this restricted review sandbox: EPERM on listen and spawn, writes denied outside the sandbox allowlist, missing Python virtualenvs, no Metal device, and EMFILE. installState.test.js itself reported file status passed with 36 of 36 assertions passing inside this full run.
- ❌ `npx vitest run --reporter=dot (full server suite, unpatched baseline at 9bb4b572f)` — 38328 tests, 34769 passed, 3533 failed across 186 files. Identical failure count and failed-file count to the patched run, with exactly one fewer passing test, so the entire 3533-failure population is pre-existing and none of it is attributable to this PR.
- ✅ `npx vitest run ../scripts/repo-scan-guards.test.js ../scripts/ci-test-plan.test.js ../scripts/agent-instructions-files.test.js lib/generatedManifests.test.js` — 47 of 47 passed. Run after marking the new research document intent-to-add so the git-tracked-file-scoped guards could actually see it, since a patch-only apply leaves a new file untracked and invisible to them.
- ⏭️ `npm run lint` — The only lint script in the repository is client-scoped, biome lint over client/src, and this change touches no client source.
- ⏭️ `npm test -w client` — No client source or client test is touched by this change.
- ⏭️ `npm run test:db` — No database-backed code path is touched, and the review sandbox has no provisioned portos_test database.

#### Notes
- The exclusion covers only .DS_Store. Other macOS metadata that Finder can drop under client/, notably AppleDouble ._ files and a Finder custom-icon file, would still produce the same false stale-build signal. The narrow set is a documented deliberate choice rather than an oversight, so this is an observation about residual scope, not a defect.
- The full local server suite carries 3533 pre-existing failures in this restricted sandbox on unpatched main, so a local run cannot substitute for CI on this branch even though the delta attributable to this PR is provably zero.

- The exclusion set is deliberately narrow and its comment explains why hidden files such as .env stay in scope, at server/services/installState.js:53.
- The exclusion is applied on both scan paths, the flat client-root pass at server/services/installState.js:138 and the recursive src and public walk at server/services/installState.js:151, so a .DS_Store at either depth is skipped.
- isClientSourceNewer has no other callers besides getInstallState at server/services/installState.js:267 and the test export at server/services/installState.js:328, so no sibling walker needs the same exclusion kept in sync.
- The new test exercises the real filesystem through getInstallState rather than a mocked clientSourceNewer, which is what makes it capable of catching the regression, at server/services/installState.test.js:416.
- The research note's claim that .DS_Store is already Git-ignored is accurate, the entry is at .gitignore:40.
- The research note's claim that backup handling already treats .DS_Store as macOS metadata is accurate, the handling is at server/services/backup.js:530.
- The research note's claim that the update banner hook sets out-of-sync state from its mount-time status request and never clears it is accurate, the set has no clearing counterpart at client/src/hooks/useUpdateChecker.jsx:53.
- The new document follows the stated convention for its directory, a dated point-in-time note cross-linked from a living document rather than added to an index, per docs/research/README.md:5.
</details>

PR #6076, opened 2026-09-03
fix: guard overlapping ImageGen picks with per-target sequence token (#5944)

### LINKED ISSUE #5944: ImageGen orphans a blob URL when two init-image picks overlap
**Problem** — Two overlapping init-image picks in `client/src/pages/ImageGen.jsx` orphan a blob URL. `handlePickInitImage` awaits `normalizeImageOrientation(raw)` before it mints its object URL and revokes the previous one via `initImagePreviewRef.current`. If the user re-picks before the first normalization resolves, both continuations read the SAME stale `initImagePreviewRef.current`, both revoke it (idempotent, fine), and both call `URL.createObjectURL`. Only the last `setInitImage` survives, so the loser's url is unreachable from state — no replace path, no clear path, and not the unmount sweep can ever revoke it.

Same shape applies to `handlePickReferenceImage` for a single slot.

**Why the existing guards don't cover it** — PR #5943 fixed the unmount race (`useMounted()` re-check before creating) and the StrictMode double-mint (url minted outside the state updater). Both are *lifecycle* guards; this one is an *ordering* problem between two in-flight picks, and a mounted flag is true for both.

**Decision** — Use a per-target sequence token, not a lock or an abort. Keep a `pickSeqRef` (`{ init: 0, refs: [0,0,0,0] }`); each handler increments its slot's counter, captures the value, and after the await bails when the captured value no longer matches (a newer pick has superseded it) — before calling `createObjectURL`, so the losing pick never mints at all rather than minting and cleaning up. This mirrors the existing `statusRequestToken` idiom already in the same file.

**Tests** — Extend `client/src/pages/ImageGen.objectUrls.test.jsx` (it already stubs `URL.createObjectURL`/`revokeObjectURL` and defers `window.createImageBitmap`, so the harness exists): issue two picks with the first normalization still pending, release both, and assert exactly one url was created and it is the one rendered.

**Acceptance criteria**
- [ ] Two overlapping picks on the init image create exactly one object URL.
- [ ] Same for two overlapping picks on one reference slot.
- [ ] The url left in state is the one from the LAST pick.
- [ ] `cd client && npm test` passes.

**Out of scope** — the unmount and StrictMode paths (fixed in #5943), and the gallery picker (mints no object URL).

Severity is low: it needs a sub-second re-pick, and leaks one image rather than compounding. Found by a local review round on #5943.

atomantic:
## Summary

Fixes the blob-URL orphan when two init-image (or same-slot reference) picks overlap in the EXIF-normalization await: both continuations read the same stale preview ref and each mint a `URL.createObjectURL`, but only the last `setState` survives, leaving the loser's url unreachable from state, the clear path, and the unmount sweep.

Per the issue decision, uses a per-target sequence token (`pickSeqRef`: `{ init, refs[]}`), mirroring the existing `statusRequestToken` idiom in the same file. Each handler bumps its slot's counter and bails when superseded **before** calling `createObjectURL`, so a losing pick never mints at all. The init handler's follow-up dimension probe gets the same staleness guard so a superseded pick can't clobber width/height either.

Closes #5944

## Test plan

- Extended `client/src/pages/ImageGen.objectUrls.test.jsx` (its `createImageBitmap` deferral harness already existed): two overlapping picks on the init image, and two overlapping picks on reference slot 0 — each asserts exactly one url is created, it belongs to the last pick (`photo-2.jpg`), it is the live (unrevoked) one, and it is the url rendered.
- Verified both new tests **fail without the fix** (2 failed / 6 passed) and pass with it.
- `npx vitest run src/pages/ImageGen.objectUrls.test.jsx` — 8 passed.
- `npx vitest run src/pages/ImageGen.probeGating.test.jsx src/pages/ImageGen.federatedTarget.test.jsx src/pages/ImageGen.flux2VenvShare.test.jsx` — 9 passed.
- Local reviewer (mtplx, low) was unavailable — no model configured for that backend — so shipping with self-review + tests; reviewer status recorded as review-blocked.

### DIFF
--- a/client/src/pages/ImageGen.jsx
+++ b/client/src/pages/ImageGen.jsx
@@ -186,6 +186,13 @@ export default function ImageGen() {
   // updater React skips once unmounted — an implementation detail to guard
   // against, not to rely on.)
   const mountedRef = useMounted();
+  // Per-target pick sequence: two overlapping upload picks both survive the
+  // EXIF-normalization await, read the same stale preview ref, and each mint a
+  // url — only the last setState survives, orphaning the loser's url. Each
+  // handler bumps its slot's counter and bails when superseded, BEFORE minting,
+  // so a losing pick never creates a url (same token idiom as
+  // statusRequestToken below).
+  const pickSeqRef = useRef({ init: 0, refs: [] });
   const [initImageStrength, setInitImageStrength] = useState(0.4);
   // Visual gallery picker target: null (closed), { kind: 'init' }, or
   // { kind: 'reference', slot: i }. The search/browse alternative to the plain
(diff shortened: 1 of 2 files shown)


atomantic:
Review-blocked: the mtplx local reviewer (effort low) returned no verdict — "No model configured for mtplx reviewer — set one on the Settings → Code Reviewers page." Code and tests are otherwise shippable (ImageGen.objectUrls 8/8, sibling ImageGen suites 9/9; new overlap tests verified to fail without the fix), so this PR stays open for human/CI review. Status recorded in the worktree-private review file.

Issue #5703, opened 2026-09-01
Inline edit controls in the health-logging tabs are 22px tap targets, half the 44px floor the rest of the app enforces


atomantic:
**Problem** — 163 files in `client/src` already use `min-h-[44px] min-w-[44px]` for icon-only controls, and `PageSkeleton.jsx:32` documents 44px as the project's tap-target floor ("a default-size TabPills button is … 44px"). The MeatSpace tabs are the app's most phone-centric surface — logging a drink or a nicotine entry is done on a phone — yet their inline Save/Cancel buttons are `p-1` or `p-1.5` around a `size={14}` icon, giving a 22-26px hit area. Two of them sit adjacent to each other, so a mis-tap on Save/Cancel discards or commits the wrong edit.

**Evidence** — `client/src/components/meatspace/tabs/NicotineTab.jsx:495`:

```jsx
                                <button aria-label="Save" onClick={saveEdit} className="p-1 text-port-success hover:text-port-success/80"><Check size={14} /></button>
                                <button aria-label="Cancel" onClick={cancelEdit} className="p-1 text-gray-500 hover:text-gray-300"><X size={14} /></button>
```

`p-1` (4px) x2 + a 14px icon = a 22x22px target. The convention it should match, from the same tree (`client/src/components/cos/tabs/MemoryEditModal.jsx:114`):

```jsx
            className="p-2 min-h-[44px] min-w-[44px] flex items-center justify-center text-gray-500 hover:text-white transition-colors rounded-lg"
```

Sixteen sub-44px icon buttons live in `client/src/components/meatspace/`; the rest of the offenders are single controls in `components/settings/LocalLlmTab.jsx:989`, `components/wiki/tabs/GraphTab.jsx:230`, `components/wiki/tabs/LogTab.jsx:85`, `components/sharing/ConflictsTab.jsx:107`, `components/media/PromptFromMedia.jsx:360`, `pages/Loops.jsx:398`, `pages/Review.jsx:747`.

**Plan**
[...]

PR #6079, opened 2026-09-03
fix: sweep sub-44px icon buttons tree-wide and widen tap-target guard (#5904)

### LINKED ISSUE #5904: Sweep the remaining sub-44px icon buttons outside MeatSpace and widen the tap-target guard to the whole client tree
**Deferred from #5703 / #5901**, which fixed the MeatSpace health-logging tree and seven named single controls but deliberately left the rest of the tree alone.

**Problem** — a tree-wide scan of git-tracked `client/src/**/*.jsx` still finds ~250 static `className` values carrying `p-0.5`/`p-1`/`p-1.5` with no `min-h-[44px]`. A good share of those are icon-only `<button>`s: a 12-16px glyph in 4-6px of padding is a 20-28px tap target, well under the 44px floor `PageSkeleton.jsx:32` documents and 163 other files already honour. Known clusters:

- `pages/Review.jsx` — the review-card action rows (`:452`, `:660`, `:667`, `:818`, `:826`, `:833`, `:840`).
- `pages/Loops.jsx` — the loop-card run/stop/resume/delete row (`:254`, `:257`, `:262`, `:266`).
- `components/wiki/tabs/GraphTab.jsx:213,220` and `components/settings/LocalLlmTab.jsx:1218` — siblings of controls already fixed in #5901.

**Scan** — `client/src/a11yConventions.test.js` already carries the matcher. Its rule "meets the 44px touch-target minimum on the MeatSpace log-row icon buttons (#5703)" is scoped to `src/components/meatspace/` and to the tight-padding shape; dropping the path filter reports the full offender list, and `isIconOnlyButton` there is the shared definition of "icon-only".

**Plan**
1. Sweep the offenders the way #5901 did: add `min-h-[44px] min-w-[44px] inline-flex items-center justify-center`, widen `gap-1` to `gap-2` where two of them sit adjacent, and leave every icon `size=` untouched — glyph size is what sets each row's density.
2. Watch the dense card/table rows: a 44px min-height grows rows that were shorter, so eyeball the affected views rather than trusting the grep alone.
3. Widen the guard's scope from `src/components/meatspace/` to the whole tree in the same PR, so the sweep can't silently regress. Keep the padding-shape filter and the self-probe.

**Out of scope** — buttons carrying text labels (padding plus text already clears 44px), the OpenWorld HUD (`.openworld-hud-action` in `index.css`), and icon buttons with **no** padding class at all or with roomy padding (`p-2`+): a different judgement call per site, and folding them in would make the sweep unreviewable.

atomantic:
## Summary

Sweeps the remaining sub-44px icon buttons outside MeatSpace the way #5901 did, and widens the tap-target guard to the whole client tree in the same PR.

- Every icon-only `<button>` with tight `p-0.5`/`p-1`/`p-1.5` padding now carries `min-h-[44px] min-w-[44px] inline-flex items-center justify-center` (~230 buttons across ~120 files). Sub-44 explicit mins (`min-h-[24/28/32/36/40px]`) are raised to the floor rather than stacked beside it. Icon `size=` values are untouched.
- Action rows holding two or more adjacent fixed buttons widen from `gap-1` to `gap-2` (Loops run/stop/delete, Review queue/item rows, EntryCard-style action rows, etc.).
- Also fixed the same shape with dynamic `className` templates the guard skips (lock toggles, feedback thumbs, star/pin toggles, mute/review toggles) found by hand in the touched rows.
- `a11yConventions.test.js`: the `#5703` tap-target rule moves from `src/components/meatspace/` to the whole tree (same padding-shape filter and self-probe, same `isIconOnlyButton` definition). Sibling-owned SongBook surfaces (`pages/SongBook.jsx`, `pages/SongBookViewer.jsx`, `components/songbook/`) are excluded for the parallel change that owns them.

Left alone, deliberately: text-label buttons, icon buttons with no padding class or roomy `p-2`+ padding, icon-only anchors, the absolute `top-0.5 right-0.5` pin-star overlay on CanonCard reference thumbnails (a per-site overlay-density call), and the sibling-owned SongBook/PracticeLogger tree plus client ImageGen init-image code.

Closes #5904

## Test plan

- `npx vitest run src/a11yConventions.test.js` — 50/50 pass (widened guard reports zero offenders).
- Full client suite (`npm test` in `client/`): 832 files passed, 10161 tests passed, 2 skipped.
- Updated `VideoTimelineLanes.test.jsx` (28px remove-target assertion → 44px floor); collocated tests of all 62 touched components with test files pass.
- Local mtplx reviewer unavailable in this environment (no model configured) — see review-blocked comment below.

### DIFF
--- a/client/src/a11yConventions.test.js
+++ b/client/src/a11yConventions.test.js
@@ -32,9 +32,9 @@
  *   6. An `<img>` with no `alt`, which is announced by its `src` — a hashed
  *      filename or a blob URL. `alt=""` is the correct spelling for a
  *      decorative image and passes; only the omission is the bug.
- *   7. An icon-only `<button>` in the MeatSpace health-logging tree sized to
- *      its bare icon (`p-1` around a 12-14px glyph = a 22px target) instead of
- *      the 44px floor the rest of the app enforces.
+ *   7. An icon-only `<button>` sized to its bare icon (`p-1` around a
+ *      12-14px glyph = a 22px target) instead of the 44px floor the rest of
+ *      the app enforces.
  *
  * Scoped to git-tracked `.jsx` under `client/src` so an untracked scratch file
  * can't fail the suite.
@@ -3266,7 +3266,7 @@ describe('a11y conventions', () => {
     expect(offenders, `Close button under the 44px touch-target minimum — add min-h-[44px] min-w-[44px] + flex items-center justify-center (see Drawer.jsx:106):\n${offenders.join('\n')}`).toEqual([]);
   });
 
-  it("meets the 44px touch-target minimum on the MeatSpace log-row icon buttons (#5703)", () => {
+  it("meets the 44px touch-target minimum on icon-only buttons (#5703, #5904)", () => {
     // The MeatSpace tabs are the app's most phone-centric surface — a drink or
     // a nicotine entry is logged one-handed — and their inline row controls
     // kept shipping as a bare `p-1`/`p-1.5` around a 12-14px icon: a 22-26px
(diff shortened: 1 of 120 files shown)


atomantic:
Review status: review-blocked. The mtplx local reviewer (effort low) returned no verdict: "No model configured for mtplx reviewer — set one on the Settings → Code Reviewers page." Code and tests are otherwise shippable — full client suite green (832 files / 10161 tests). Leaving the PR open for human/CI review.

PR #6093, opened 2026-09-03
replace real sleeps and a 22.7M-element Python fixture in the detachedSpawn and trellis2 runner suites

### LINKED ISSUE #6007: CI: replace real sleeps in detachedSpawn and trellis2GenerateRunner tests
## Why

- `server/lib/detachedSpawn.test.js` (4.7s Linux, 28 tests) has 13 real `setTimeout` sleeps polling for child-process state instead of awaiting an event.
- `server/services/imageTo3d/trellis2GenerateRunner.test.js` (17s both runners, 24 tests) uses no fake timers and appears to have real waits/polling somewhere in its 24 tests — needs investigation to find the source before it can be fixed.

## What to do

- `detachedSpawn.test.js`: replace the 25–50ms polling loops with an `await once(child, 'spawn')` / readiness-event await where the child process signals readiness, instead of sleeping and re-checking.
- `trellis2GenerateRunner.test.js`: profile which of the 24 tests carry real wall-clock time (`vitest --reporter=verbose`), determine whether they're waiting on a real timer, a real async I/O op, or something else, and convert to fake timers (`vi.useFakeTimers()`) or a mocked equivalent where safe.
- Per repo test-strategy guidance (`AGENTS.md`): "Real timeout behavior gets one focused contract test with injected/fake time — never repeated production sleeps." If any of these tests are asserting real timeout/retry *behavior* rather than just being slow, keep exactly one such test with injected time and convert the rest.

## Acceptance criteria

- `detachedSpawn.test.js` and `trellis2GenerateRunner.test.js` show materially reduced wall time (no more real-sleep waits) with no test skipped or weakened.
- Record the before/after `⏱` line from `scripts/run-ci-tests.js`'s job summary in the PR.

## Files/areas

- `server/lib/detachedSpawn.test.js`
- `server/services/imageTo3d/trellis2GenerateRunner.test.js`

Part of #5902

---

When you ship this slice, check its box in #5902 (edit the epic body checklist), and if this was the LAST open child, close #5902 with a summarizing comment naming all the slices that delivered it.

atomantic:
## Summary

Removes the real wall-clock waits from two of the slowest server suites.

- **`server/lib/detachedSpawn.test.js`** — 13 fixed `setTimeout` sleeps (25–200ms, paid unconditionally) become a bounded `waitUntil` condition poll on the state each test is actually about: the job is alive under its recorded PID, the cleanup `rm` landed, the log holds a line. The two `reattachDetached` fixtures that ran a real `sleep 1` inside `sh` now block on a marker file the test writes, so the replay window stays open exactly as long as the test needs it. The "`replayed` fires before the live line" assertion is now ordered by the event instead of by winning a race against a one-second sleep.
- **`reapAndCleanDetachedDirs`** gains an opts passthrough to `reapDetached` (default unchanged), so the sweep test uses the same 25ms cadence as every other job it spawns rather than the 250ms production default, twice.
- **`server/services/imageTo3d/trellis2GenerateRunner.test.js`** — this one was not sleeping at all. Its fixture built the 22.7M-face decoder mesh as `list(range(...))`: ~300ms of Python list allocation per test and hundreds of MB per worker. Nothing (neither the adapter nor either stub) does more than `len()` on it, so `range` yields identical numbers for free. Under the parallel-worker contention CI actually runs at, that allocation dominated everything else.

No test is skipped, and no assertion is weakened. One assertion got **stronger**: bypass-probing the change surfaced that the pre-decimation-capture test matched `no pre-decimation mesh was captured`, a wording the adapter no longer prints — it passed with the capture ripped out entirely. It now matches the real `normal map skipped` message.

## Test plan

Both suites pass unchanged (45 + 24 tests). Timings from the same machine:

| run | before | after |
| --- | --- | --- |
| `detachedSpawn.test.js` alone, idle | 4.44s wall / 4.25s tests | 2.56s wall / 1.99s tests |
| `trellis2GenerateRunner.test.js` alone, idle | 5.02s wall / 4.78s tests | ~2.9s wall / 2.38s tests |
| both together, under parallel-agent contention (two back-to-back runs) | 28.42s / 22.20s wall | 3.73s / 3.61s wall |

The contended row is the one that matters for CI: the fixture allocation is what turns a 9s pair into a 28s pair once workers compete for memory.

Bypass probes confirming the converted tests still catch their regressions:

- disabling the `cleanup` `rm` in `finish()` → `removes the control dir after the job ends when cleanup is set` fails.
- disabling `_CAPTURED["source"]` in `trellis2GenerateRunner.py` → `captures the pre-decimation mesh for the normal bake` fails (it passed before the assertion fix).

Also ran `server/services/mediaJobQueue` (the only production caller of `reapAndCleanDetachedDirs`): 116 tests pass.

Closes #6007

### DIFF
--- a/server/lib/detachedSpawn.js
+++ b/server/lib/detachedSpawn.js
@@ -645,15 +645,16 @@ export async function reapDetached(controlDir, { graceMs = REAP_GRACE_MS, pollMs
  * new job starts (every dir present then is an orphan from the prior process).
  *
  * @param {string} parentDir - the `.detached` parent (e.g. PATHS.videos/.detached)
+ * @param {object} [opts] - forwarded to reapDetached (graceMs, pollMs)
  * @returns {Promise<{reaped: number, scanned: number}>}
  */
-export async function reapAndCleanDetachedDirs(parentDir) {
+export async function reapAndCleanDetachedDirs(parentDir, opts = {}) {
   const entries = await readdir(parentDir).catch(() => []);
   let reaped = 0;
   for (const name of entries) {
     const dir = join(parentDir, name);
     // eslint-disable-next-line no-await-in-loop
-    const res = await reapDetached(dir).catch(() => ({ reaped: false }));
+    const res = await reapDetached(dir, opts).catch(() => ({ reaped: false }));
     if (res.reaped) reaped += 1;
     // eslint-disable-next-line no-await-in-loop
     await rm(dir, { recursive: true, force: true }).catch(() => {});
--- a/server/lib/detachedSpawn.test.js
+++ b/server/lib/detachedSpawn.test.js
@@ -1,5 +1,5 @@
 import { describe, it, expect, afterEach, vi } from 'vitest';
-import { mkdir, mkdtemp, rm, stat, writeFile } from 'fs/promises';
+import { mkdir, mkdtemp, readFile, rm, stat, writeFile } from 'fs/promises';
 import { tmpdir } from 'os';
 import { basename, join } from 'path';
 import { ChildProcess, execFile } from './childProcess.js';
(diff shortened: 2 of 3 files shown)


atomantic:
CI-run confirmation (Server tests job on this PR, Linux runner) — the numbers the issue asked to record:

```
✓ services/imageTo3d/trellis2GenerateRunner.test.js (24 tests) 1774ms
✓ lib/detachedSpawn.test.js                          (45 tests) 1461ms
```

Against the baselines quoted in #6007 — `detachedSpawn.test.js` 4.7s and `trellis2GenerateRunner.test.js` ~17s (both runners) — that is roughly 3x and 10x. Same test counts, nothing skipped.

Issue #5992, opened 2026-09-03
[repo-study-fable-advisor-orchestration-profiles] Add an orchestrated CoS execution mode with per-role provider/model and per-step reasoning effort


atomantic:
Phase 1 of #5991.

**Provenance:** studied `DannyMac180/fable-advisor` (MIT, © 2026 Dan McAteer) on 2026-09-03. Clean-room — the technique is reimplemented against PortOS modules, nothing is copied.

## Rationale

A CoS task resolves one provider and one model for its entire run, so a user who wants strong reasoning for the planning pays that model's rate for the mechanical editing too. This phase adds an opt-in **orchestration profile** — architect / implementer / reviewer roles, each with its own provider+model — and the per-step spec contract that makes delegating to a context-free lane safe.

## What exists today (the gap is real)

- `selectModelForTask` (`server/services/agentModelSelection.js`) picks ONE model tier at spawn from `task.metadata.model` / `metadata.provider`, complexity heuristics and `taskLearning`'s `suggestModelTier`. There is no role dimension.
- `resolveThinkingLevel` / `getModelForLevel` (`server/services/thinkingLevels.js`) resolve a single effort for the whole run from `task.metadata.thinkingLevel` (task → agent → provider precedence). Per-step effort does not exist; #3641 shipped only a soft run-level default for pipeline stages.
- `agentProviderResolution.js` resolves one provider per run.
- `agentPromptBuilder.js` builds one prompt for one agent.

## Fix

- `server/lib/validation.js` — add an `orchestrationProfile` schema: `{ architect: {provider, model, effort}, implementer: {provider, model, defaultEffort}, reviewer: {provider, model, effort} }`, plus `executionMode: 'direct' | 'orchestrated'` on the CoS task input schema. Keep `direct` the default so existing tasks are untouched.
- `server/services/agentModelSelection.js` — add `selectModelForRole(task, role, provider, agent)` alongside `selectModelForTask`; the existing entry point becomes the `direct`-mode / architect-role caller so current behavior is byte-identical when no profile is set.
- `server/services/thinkingLevels.js` — let a resolved effort be carried per delegated step rather than only per run; the architect names `REASONING: <rung>` in each spec and the lane passes it through unchanged (never rounds it — an unsupported rung is an error, not a downgrade).
[...]

PR #6095, opened 2026-09-03
Add an opt-in orchestrated CoS mode with per-role provider/model and per-step reasoning effort

atomantic:
## Summary

Adds the opt-in **orchestration profile** from #5992: a CoS run can be split into three roles — architect (plans, writes specs, delegates), implementer (executes one spec), reviewer (checks it) — each with its own provider, model, and reasoning effort, instead of resolving one provider + one model + one effort for the whole run.

- `server/lib/orchestrationProfile.js` — vocabulary, normalizer, role gating, and the six-part spec contract a context-free delegated lane needs. `parseReasoningDirective` **never rounds**: an unsupported rung is an error, because substituting the nearest supported level would run a step at an effort nobody chose while still reporting success.
- `selectModelForRole` sits alongside `selectModelForTask`, which now resolves through the architect role. With no profile the selection is unchanged, and there is a test asserting that equivalence.
- `resolveStepEffort` resolves a delegated step's rung as spec directive → role default → run effort.
- `buildOrchestrationDoctrineSection` renders the architect doctrine into all three prompt paths (template, full, light). It returns `''` for every `direct`-mode task — the default — so an install that configures no profile is untouched.
- The mode + profile persist as CoS task metadata (`addTask`, `updateTask`, and the create/update routes), with the repo's absent-vs-cleared semantics.

No migration or `data.reference/` seed: both keys ride the existing free-form task-metadata blob, additively, so an older peer ignores them.

**Naming decision:** the issue proposed `executionMode: 'direct' | 'orchestrated'`. Agent metadata already carries an `executionMode` meaning `tui` / `runner` / `runner-tui` / `direct` — reusing the name, including the value `direct`, for an unrelated axis would have made both unreadable. Shipped as `orchestrationMode` instead.

## Test plan

- `cd server && npm test` — full server suite green (1 unrelated load-timeout flake in `routes/settings.secretsStrip.test.js`, passes on its own re-run).
- New: `server/lib/orchestrationProfile.test.js`, `server/services/promptSections/orchestrationDoctrine.test.js`.
- Extended: `agentModelSelection.test.js` (role pin wins; unpinned role and direct-mode task are identical to `selectModelForTask`), `agentProviderResolution.test.js` (architect provider pin resolves; a direct-mode task keeps its own metadata pin), `thinkingLevels.test.js` (spec → role → run precedence; unsupported rung errors), `cosTaskRoutes.test.js` (pass-through, unknown role rejected, bad rung rejected, clear-on-update).

## Remaining

Deliberate partial ship — the rest of #5992 stays open:

- Client profile picker on the CoS task form and in `/ai`.
- A machine-local named-profile library in Settings (per the issue's "profiles are per-install config, not a federated record"), so a profile is configured once rather than per task.
- Feeding `resolveStepEffort` from a returned spec at the point PortOS re-dispatches a delegated unit, once that lane exists — today the rung reaches the lane through the prompt contract, which is the only channel available while sub-agent dispatch happens inside the agent CLI.

Refs #5992


atomantic:

Issue #6003, opened 2026-09-03
CI: speed up Windows git-sandbox suites (primaryCheckoutGuard, worktreeReap)


atomantic:
## Why

`server/lib/primaryCheckoutGuard.test.js` (78s Windows, 6.5s Linux, 33 tests) and `server/services/worktreeReap.test.js` (53s Windows, 3.4s Linux, 18 tests) each build a fresh real-git sandbox per test via `useGitSandbox()`'s `beforeEach`/`afterEach` (`makeGitSandbox()` copies a cached template dir, `destroyGitSandbox()` `rm -rf`s it). The template build itself is already cached per worker (`server/lib/gitTestRepo.js` `getTemplate()`), but the per-test `fs.cp` + `rm -rf` of the sandbox tree is what dominates on Windows' slower filesystem syscalls.

## What to do

Build the sandbox once per `describe` in `beforeAll` and reset it between tests with real git commands instead of a filesystem copy+delete cycle:

- `git checkout -- . && git clean -fd` to reset working-tree state between tests.
- Re-attach/re-push the bare origin only for tests that actually need one (most currently call `addOrigin()`/`attachBareOrigin()` per test — hoist what's shared, keep an explicit re-attach only where a test mutates the remote).
- Keep the existing `SKIP_HEAVY_INTEGRATION` describe-skip for `VITEST_FAST=1`.
- Watch for cross-test leakage: a `beforeAll` sandbox means one test's branch/commit state can bleed into the next if the reset step misses something (stray branches, detached HEAD, dangling refs) — verify with `git status --porcelain` in the reset helper if needed.

Consider adding a `useGitSandbox({ scope: 'describe' })`-style option to `gitTestRepo.js` shared by both files rather than duplicating the beforeAll/reset logic.

## Acceptance criteria

- `primaryCheckoutGuard.test.js` and `worktreeReap.test.js` each run under 15s on Windows CI (currently 78s / 53s).
- No test skipped or weakened; all existing assertions still pass on both Linux and Windows.
- Record the before/after `⏱` line from `scripts/run-ci-tests.js`'s job summary in the PR.

## Files/areas

- `server/lib/gitTestRepo.js`
- `server/lib/primaryCheckoutGuard.test.js`
- `server/services/worktreeReap.test.js`

Part of #5902

---

When you ship this slice, check its box in #5902 (edit the epic body checklist), and if this was the LAST open child, close #5902 with a summarizing comment naming all the slices that delivered it.

PR #6099, opened 2026-09-03
test: fix ~3% flake in resetGitWorktreeSandbox worktree-listing assertion

### LINKED ISSUE #6059: gitTestRepo worktree-reset test flakes ~3% of runs on a mkdtemp name collision
## What happens

`server/lib/gitTestRepo.test.js > resetGitWorktreeSandbox > removes every worktree it grew, including a locked one, and restores main` fails intermittently in CI:

```
AssertionError: expected 'worktree /tmp/portos-git-fx-reset-wt-…' not to contain 'wt-b'

- wt-b
+ worktree /tmp/portos-git-fx-reset-wt-bvAY8J
+ HEAD fc42a146eeb0f35de344c2b449750cf7f4eeb2e6
+ branch refs/heads/main
```

Observed on PR #6058, which touches none of this code. It cancels every sibling CI job, so a whole run reads red from one unrelated assertion.

## Why

The parent sandbox and the two worktrees share a name prefix:

```js
const dest = await mkdtemp(join(tmpdir(), 'portos-git-fx-reset-wt-'));   // line 171
const wtA  = await mkdtemp(join(tmpdir(), 'portos-git-fx-reset-wt-a-')); // line 176
const wtB  = await mkdtemp(join(tmpdir(), 'portos-git-fx-reset-wt-b-')); // line 177
…
expect(listing).not.toContain('wt-a');   // line 186
expect(listing).not.toContain('wt-b');   // line 187
```

`git worktree list --porcelain` always lists the **main** worktree — i.e. `dest` itself, which the reset is supposed to leave in place. `mkdtemp` appends six random characters to `portos-git-fx-reset-wt-`, so whenever that suffix happens to start with `a` or `b`, `dest`'s own path contains the literal substring `wt-a` / `wt-b` and the assertion fires against a listing that is completely correct.

Six characters drawn from a 62-character alphabet ⇒ **P(first char is `a` or `b`) = 2/62 ≈ 3.2%** per run, on a suite that runs on every PR across several shards.

The assertion is also weaker than it reads: `toContain` on a path substring would miss a worktree whose *branch* survived under a different directory name.

## Fix

Assert on what the test actually means — that the two worktree **paths** are gone — instead of on a substring that the parent path can satisfy by accident. Either:

- compare against the resolved `wtA` / `wtB` paths (`expect(listing).not.toContain(wtA)`), or
- parse the porcelain output into worktree entries and assert the set is exactly `[dest]`.

The second is stronger and reads better against the test's name ("removes every worktree it grew … and restores main"). While there, the sibling assertion on `branch --format=%(refname:short)` already pins that the `wt-a` / `wt-b` **branches** are gone, so the path check does not need to carry that job too.

Introduced by #6003 (`6ec4e7008` / `ea04b6ef3` / `f56c6504f`), which added the `-wt-` prefix.

## Verify

Force the collision deterministically rather than waiting for the 3%: create the parent sandbox at a path that ends in `wt-b…` (e.g. `mkdtemp(join(tmpdir(), 'portos-git-fx-reset-wt-b'))`) and confirm the test still passes after the fix.

Bryandero98:
## Summary
- Fixes #6059
- The `resetGitWorktreeSandbox` test asserted `.not.toContain('wt-a'/'wt-b')` against the raw `git worktree list --porcelain` output, which always includes the parent sandbox's own entry. When that entry's random `mkdtemp()` suffix happens to start with the same letter as a child worktree's differentiator, the substring coincidentally appears inside the parent's own path and the assertion false-fails (~3.2% chance per prefix).
- Replaced the substring check with a parse of the porcelain output into worktree entries — mirroring the parsing `resetGitWorktreeSandbox` itself already does in `gitTestRepo.js` (which drops the first entry as always being the parent) — and asserts exactly 1 entry remains. This sidesteps both the collision and comparing full paths directly, which git can respell on Windows (see the `assertPath` test in this same file, #6003).

## Test plan
- `cd server && npx vitest run lib/gitTestRepo.test.js` → 10/10 passed.
- Forced the exact collision deterministically with a standalone script (parent sandbox created at a path ending in `wt-b-XXXXXX`): confirmed the old substring assertion fails in that scenario and the new entry-count assertion does not — proof the fix removes the flake rather than passing by luck.

### DIFF
--- a/server/lib/gitTestRepo.test.js
+++ b/server/lib/gitTestRepo.test.js
@@ -182,9 +182,15 @@ describe('resetGitWorktreeSandbox', () => {
 
     await resetGitWorktreeSandbox(dest, initialHead);
 
+    // Parse porcelain into worktree entries and assert only `dest` itself
+    // remains, rather than substring-matching 'wt-a'/'wt-b' against the raw
+    // listing — that substring can coincidentally appear inside dest's own
+    // random mkdtemp() suffix (~3.2% chance per prefix) and false-fail (#6059).
+    // Counting entries also sidesteps comparing absolute paths directly,
+    // which git can respell on Windows (see the assertPath test above, #6003).
     const listing = (await execGit(['worktree', 'list', '--porcelain'], dest)).stdout;
-    expect(listing).not.toContain('wt-a');
-    expect(listing).not.toContain('wt-b');
+    const worktreeEntries = listing.split(/\r?\n/).filter((line) => line.startsWith('worktree '));
+    expect(worktreeEntries).toHaveLength(1);
     expect((await execGit(['branch', '--format=%(refname:short)'], dest)).stdout.trim()).toBe('main');
     expect((await execGit(['rev-parse', 'HEAD'], dest)).stdout.trim()).toBe(initialHead);
   });


atomantic (approved):
✅ **Approved**

This is a test-only fix for a real ~3% flake in resetGitWorktreeSandbox's worktree-listing assertion, replacing a substring match that can collide with the parent sandbox's own random directory suffix with a parse of porcelain entries that mirrors logic already used in production. I could not run vitest in this sandbox (no node_modules and no network access to install), so I independently reproduced the exact git worktree sequence the test exercises using real git and confirmed the new assertion holds and the change introduces no regression.

**Scope:** single test file change in server/lib/gitTestRepo.test.js, replacing two substring assertions with a parsed worktree-entry count assertion

#### Test evidence
- 🚧 `cd server && NODE_ENV=test npx vitest run lib/gitTestRepo.test.js` — server/node_modules is not installed in this disposable worktree and the sandbox denies network access to the npm registry (403/deny on registry.npmjs.org), so vitest could not be installed or run. This is an environment limitation, not a patch problem.
- ✅ `manual git reproduction of the exact worktree add/lock/remove/prune sequence the test performs, using real git (no vitest/node_modules needed)` — Created a real git repo, added worktrees on branches wt-a and wt-b, locked wt-b, then reproduced resetGitWorktreeSandbox's own removal logic (unlock, remove --force, prune) from server/lib/gitTestRepo.js:242-262. Post-removal `git worktree list --porcelain` contained exactly one `worktree ` line (the parent repo itself), matching the patched assertion's expect(worktreeEntries).toHaveLength(1). Also confirmed resetGitSandbox's branch cleanup (checkout main + branch -D for all non-main branches, gitTestRepo.js:209-213) removes wt-a/wt-b afterward, so the test's later branch and HEAD assertions a

- server/lib/gitTestRepo.test.js:191-193 — the patched assertion parses `worktree ` lines from the porcelain listing and asserts exactly 1 remains, which I confirmed empirically against real git worktree removal output
- server/lib/gitTestRepo.js:246-262 — resetGitWorktreeSandbox's own porcelain-parsing logic treats the first entry as the parent repo and removes the rest; the patched test's approach (count worktree entries) mirrors this existing, already-relied-upon production parsing rather than introducing new logic
- server/lib/gitTestRepo.test.js:170-177 — the parent sandbox directory `dest` is created via mkdtemp with prefix 'portos-git-fx-reset-wt-', so a random suffix starting with 'a' or 'b' produces a path containing the literal substring 'wt-a' or 'wt-b', confirming the described flake mechanism in the old assertion was real, not hypothetical
</details>

Issue #6087, opened 2026-09-03
Thread ProviderModelSelector's loading gate to the remaining provider pickers


atomantic:
## Problem

`ProviderModelSelector` now takes a `loading` prop that disables its selects and renders "Loading providers…" instead of the caller's `emptyProviderOption` while the provider list is in flight. Without it, an empty `providers` array renders a picker whose only option is "Default (active provider)" / "Inherit (…)" — indistinguishable from a broken control.

The prop is wired for the CoS **Schedule** tab only (card quick-pins, plus all three pickers in the task config drawer). Every other picker still shows the bare sentinel during its own fetch window.

## Sites still unwired

**CoS page — same `providers` state, same in-flight window, already available at the render site as `providersLoaded`:**
- `client/src/pages/ChiefOfStaff.jsx` — `TasksTab`, `AgentsTab`, `WorkflowTab` receive `providers` but not `providersLoaded`.
- `TasksTab` → `TaskAddForm` (~9 `providers=` call sites via `TaskItem`/`SortableTaskItem`) — note `TaskAddForm` also *clears* a pinned provider whenever `enabledProviders` is empty, which mid-fetch it always is; the flag prevents that by construction.
- `AgentsTab` → `RelaunchAgentModal`, `ResumeAgentModal` (`emptyProviderOption="Auto (default)"`).
- `WorkflowTab` → `AppOverridePanel` → `AppProviderPin` (already accepts `loading`, just needs threading).

**Pickers that fetch through `useProviderModels`:** the hook already returns a `loading` flag for free at ~40 call sites; only 5 consume it (`LoomEpisodeFeedback`, `LoomSeriesPlan`, `LoomEditorialAutomation`, `FableLoom`, `LoomEpisodeOutlinePlanner`), and all 5 reach only `disabled`, never the label. Pass `loading={loading}` instead.

## Also worth doing

`getApps()` and `getCosLearningSummary()` in `ChiefOfStaff.fetchData` still sit in the `Promise.all` whose tail is `getCosActionableInsights` (which fans out to `cos.runHealthCheck()` **and** `detectIdleLeftoverBranches()` — git work across repos). `apps` feeds the Schedule/Tasks/Agents app pickers, so it has the same user-visible empty-control symptom. The fix is the same `.then(applyX)` split; reuse the `sameJsonShape` identity guard so the earlier commit doesn't add a re-render per poll.

## Acceptance

[...]

PR #6106, opened 2026-09-03
fix: thread ProviderModelSelector's loading gate to the remaining provider pickers

Bryandero98:
## Summary
- `ChiefOfStaff.fetchData` now settles `apps` on its own read (`appsRead`, mirroring `providersRead`) instead of holding it hostage inside `secondaryRead`'s `Promise.all`, with a `sameJsonShape` guard to keep array identity stable on an unchanged poll.
- `providersLoaded` now flows from `ChiefOfStaff` down through `TasksTab` → `TaskAddForm`/`TaskItem`/`SortableTaskItem`, `AgentsTab` → `RelaunchAgentModal`/`ResumeAgentModal`, and `WorkflowTab` → `AppOverridePanel` → `PerAppOverrideList` (which already accepted the prop from the Schedule-tab fix, just wasn't fed by Workflow).
- `TaskAddForm`'s provider-reset effect was clearing a legitimately pinned provider on every mount, because `enabledProviders` is always empty mid-fetch — it's now gated on `providersLoaded`.
- `ResumeAgentModal` renders its own `<select>`s rather than `ProviderModelSelector`; applied the same "Loading providers…" + disabled treatment inline instead of restructuring the modal (no dedicated test file existed for it, so I kept the change surgical).
- The 5 `useProviderModels` consumers named in the issue (`LoomEpisodeFeedback`, `LoomSeriesPlan`, `LoomEditorialAutomation`, `FableLoom`, `LoomEpisodeOutlinePlanner`) now pass the hook's `loading` flag into `ProviderModelSelector` alongside the `disabled` prop they already wired.

Follow-up to the Schedule-tab loading-gate fix, per #6087.

## Test plan
- [x] `client/src/components/cos/tabs/AgentsTab.test.jsx`
- [x] `client/src/components/cos/tabs/RelaunchAgentModal.test.jsx`
- [x] `client/src/components/cos/tabs/schedule/AppOverrideRow.test.jsx`
- [x] `client/src/components/cos/tabs/TaskItem.test.jsx`
- [x] `client/src/components/cos/tabs/TasksTab.test.jsx`
- [x] `client/src/components/cos/tabs/WorkflowTab.providers.test.jsx`
- [x] `client/src/components/cos/TaskAddForm.test.jsx`
- [x] `client/src/components/fableloom/LoomEditorialAutomation.test.jsx`
- [x] `client/src/components/fableloom/LoomEpisodeFeedback.test.jsx`
- [x] `client/src/components/fableloom/LoomEpisodeOutlinePlanner.test.jsx`
- [x] `client/src/components/fableloom/LoomSeriesPlan.test.jsx`
- [x] `client/src/pages/FableLoom.test.jsx`
- [x] `biome lint --error-on-warnings` on every changed file

### DIFF
--- a/client/src/pages/ChiefOfStaff.jsx
+++ b/client/src/pages/ChiefOfStaff.jsx
@@ -190,6 +190,18 @@ export default function ChiefOfStaff() {
     return data;
   }, []);
 
+  // Same self-committing-read fix as `applyProviders`, for the same reason: apps
+  // feeds the Schedule/Tasks/Agents app pickers, so bundling it into
+  // `secondaryRead`'s Promise.all held it hostage to `getCosActionableInsights`
+  // (a server-side health check) and left those pickers showing an empty list
+  // for seconds. `sameJsonShape` keeps the array identity stable on an unchanged
+  // 30s poll payload so this doesn't cost a full-tree re-render each tick.
+  const applyApps = useCallback((data) => {
+    const filtered = (Array.isArray(data) ? data : []).filter(a => a.id !== 'portos-autofixer');
+    setApps(prev => (sameJsonShape(prev, filtered) ? prev : filtered));
+    return filtered;
+  }, []);
+
   // Derive agent state from system status
   const deriveAgentState = useCallback((statusData, agentsData, healthData) => {
     if (!statusData?.running) return 'sleeping';
@@ -232,8 +244,10 @@ export default function ChiefOfStaff() {
     const providersRead = api.getProviders()
       .catch(() => ({ providers: [] }))
       .then(applyProviders);
+    // Same rationale as providersRead above: apps commits on its own settle
+    // instead of waiting on the slower siblings in secondaryRead.
+    const appsRead = api.getApps().catch(() => []).then(applyApps);
     const secondaryRead = Promise.all([
-      api.getApps().catch(() => []),
       api.getCosLearningSummary().catch(() => null),
       // `silent: true` keeps transient poll blips quiet, matching the banner's
       // retired 60s poll; `.catch(() => null)` → preserve last-good below.
@@ -264,10 +278,10 @@ export default function ChiefOfStaff() {
     const runningAgent = agentsData.find(a => a.status === 'running');
     setActiveAgentMeta(runningAgent?.metadata || null);
 
-    const [appsData, learningSummaryData, insightsData] = await secondaryRead;
-    // Both self-committing reads are barriers, not values: `mergedHealth` below
-    // reads what `healthRead` wrote, so it must not run before they settle.
-    await Promise.all([healthRead, providersRead]);
+    const [learningSummaryData, insightsData] = await secondaryRead;
+    // All three self-committing reads are barriers, not values: `mergedHealth`
+    // below reads what `healthRead` wrote, so it must not run before they settle.
+    await Promise.all([healthRead, providersRead, appsRead]);
     // `getCosHealth` above reads the *pre-check* persisted health, while the
     // getCosActionableInsights call in this same batch triggers a fresh server
     // health check (cos.runHealthCheck) that emits `cos:health:check` — the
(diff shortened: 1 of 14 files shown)


client/src/pages/ChiefOfStaff.jsx
```diff
@@ -232,8 +244,10 @@ export default function ChiefOfStaff() {
     const providersRead = api.getProviders()
       .catch(() => ({ providers: [] }))
       .then(applyProviders);
+    // Same rationale as providersRead above: apps commits on its own settle
+    // instead of waiting on the slower siblings in secondaryRead.
+    const appsRead = api.getApps().catch(() => []).then(applyApps);
```
atomantic, line 249: 💡 **Non-blocking** — apps fetch failure clears last-good list, same as providers

The catch fallback on appsRead turns a transient fetch failure into an empty array, and applyApps (line 201) then overwrites the last-good apps state with an empty list instead of preserving it, so one failed poll blips the app pickers empty until the next successful fetch. This is not new: the providers path in applyProviders already has identical behavior and this PR mirrors it deliberately. The same function already preserves last-good for insights specifically to avoid this flicker, so a follow-up could make both the apps and providers self-commit reads keep the last-good list on a failed fetch instead of resetting to empty.

atomantic (approved):
✅ **Approved**

A clean, well-tested follow-up to #6087 that threads the providers-loaded gate and ProviderModelSelector's loading flag into the remaining COS and FableLoom provider pickers. The patch applies cleanly against main, all 14 covering test files pass (236 tests), every new prop defaults to a backward-compatible value, and the new self-committing apps read mirrors the already-shipped applyProviders pattern without introducing a new failure mode. No blocking findings; one non-blocking note on the shared wipe-on-transient-failure fallback.

**Scope:** Client-only: threads providersLoaded/loading from ChiefOfStaff through the Tasks/Agents/Workflow tabs, the task/agent modals, and five FableLoom ProviderModelSelector consumers.

#### 💡 Non-blocking (1)
- `client/src/pages/ChiefOfStaff.jsx:249` — apps fetch failure clears last-good list, same as providers

#### Test evidence
- ✅ `NODE_ENV=test npx vitest run src/components/cos/tabs/AgentsTab.test.jsx src/components/cos/tabs/RelaunchAgentModal.test.jsx src/components/cos/tabs/schedule/AppOverrideRow.test.jsx src/components/cos/tabs/TaskItem.test.jsx src/components/cos/tabs/TasksTab.test.jsx src/components/cos/tabs/WorkflowTab.providers.test.jsx` — 6 files, 106 tests passed
- ✅ `NODE_ENV=test npx vitest run src/components/cos/TaskAddForm.test.jsx src/components/fableloom/LoomEditorialAutomation.test.jsx src/components/fableloom/LoomEpisodeFeedback.test.jsx src/components/fableloom/LoomEpisodeOutlinePlanner.test.jsx src/components/fableloom/LoomSeriesPlan.test.jsx src/pages/FableLoom.test.jsx` — 6 files, 59 tests passed
- ✅ `NODE_ENV=test npx vitest run src/pages/ChiefOfStaff.test.jsx src/components/ProviderModelSelector.test.jsx` — 2 files, 71 tests passed (the ChiefOfStaff caller and the unchanged ProviderModelSelector contract)
- ⏭️ `npm test -w client` — Full client workspace suite deliberately not run as review evidence: the sandbox denies network binds, GPU, and background services a minority of suites need, so a from-zero full run reports thousands of environment-driven failures unrelated to this patch; the 14 targeted files covering every patched component were run instead.

#### Notes
- The PR description carries a Claude Code attribution footer, which the repo convention says PR descriptions should not include; a maintainer may ask the author to drop it.

- ProviderModelSelector's loading prop disables the provider, model, and effort selects and swaps the empty option for a Loading providers placeholder (client/src/components/ProviderModelSelector.jsx:212).
- providersLoaded is stamped true only inside applyProviders, the single provider-list write path, so the flag can never be true without the list present (client/src/pages/ChiefOfStaff.jsx:189).
- The new appsRead self-committing read is added to the settle barrier alongside healthRead and providersRead (client/src/pages/ChiefOfStaff.jsx:284).
- LoomSeriesPlan destructures the hook's loading flag (client/src/components/fableloom/LoomSeriesPlan.jsx:384) and threads it into ProviderModelSelector (client/src/components/fableloom/LoomSeriesPlan.jsx:457); the other four FableLoom consumers do the same with their respective loading variables.
- providersLoaded is passed from ChiefOfStaff into WorkflowTab, the tab that previously accepted the prop but was never fed it (client/src/pages/ChiefOfStaff.jsx:1247).
</details>

PR #6107, opened 2026-09-03
feat: add additive provider-model insert migration family, port 337 to it

Bryandero98:
## Summary
- Adds `makeAdditiveProviderInsertMigration({ targets, label })` to `scripts/migrations/_lib.js` as a new "family 7b": it splices a replacement model id after a retired one on a provider's `models` list, without rewriting or dropping anything else on a possibly-curated (user-edited) list — unlike family 7's exact-match pristine-list rewrite, which is the wrong tool when the list may have been hand-customized.
- Adds `runAdditiveProviderInsertMigrationTests({ migration, targets, prefix })` to `scripts/migrations/_testHelpers.js`, following the same convention as the existing shared test helpers: fixtures are derived from the migration's own `targets` table so behavior and coverage can't drift apart.
- Rewrites migration 337 (`claude-sonnet-5` additive repair) as a thin config wrapping the new factory, exporting `TARGETS` so its test file can build on the shared helper instead of hand-rolling its own fixtures.

## Test plan
- [x] `scripts/migrations/337-claude-sonnet-5-additive.test.js` — 96/96 (via the new shared helper)
- [x] Full `scripts/migrations/` directory run — no new failures introduced

### DIFF
--- a/scripts/migrations/337-claude-sonnet-5-additive.js
+++ b/scripts/migrations/337-claude-sonnet-5-additive.js
@@ -10,9 +10,10 @@
  * that record: the reviewer/task model pickers reading it offer the retired
  * sonnet and cannot offer the current one at all.
  *
- * ADDITIVE, deliberately — the opposite policy from 153/206 and from
- * `makeSeededProviderTierMigration`, because this one runs against lists the
- * user curated:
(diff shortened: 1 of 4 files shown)


atomantic (approved):
Reviewed — change is correct, focused, and covered by tests. Approving.

PR #6108, opened 2026-09-03
fix: PromptManager stage save/create/delete give no visible confirmation

Bryandero98:
## Summary
- `saveStage`, `createStage`, and `confirmDeleteStage` in `PromptManager.jsx` mutated the stage and silently reloaded — no toast ever confirmed the save/create/delete happened.
- `createStage` also never selected the stage it just created, so the user was left looking at whatever stage was selected before, with no visible confirmation their new stage exists.
- All three now call `toast.success(...)`; `createStage` additionally calls `setSelectedStage(payload.stageName)` so the URL and editor deep-link to the stage just made.

## Test plan
- [x] `client/src/pages/PromptManager.test.jsx` — 53/53, including a new `describe` block covering all three actions' feedback

### DIFF
--- a/client/src/pages/PromptManager.jsx
+++ b/client/src/pages/PromptManager.jsx
@@ -204,6 +204,7 @@ export default function PromptManager() {
       .catch((err) => { toast.error('Failed to save stage: ' + err.message); return false; });
     if (!ok) { setSaving(false); return; }
     setSaving(false);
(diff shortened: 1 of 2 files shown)


atomantic (approved):
Reviewed — change is correct, focused, and covered by tests. Approving.

PR #6110, opened 2026-09-03
fix: fableLoom asset manifest omits playbackAssets video clips

Bryandero98:
## Summary
- `buildFableLoomAssetManifest` (`server/services/sharing/peerSyncAssets.js`) only collected video ids from the legacy `node.videoHistoryId` field.
- A scene render's actual clips live under `node.playbackAssets` (`entryVideoHistoryId`, `holdLoopVideoHistoryIds[]`, and per-transition `exitByTransition`), none of which were ever included — so those clips were silently dropped from the peer-sync manifest and never federated to peers.
- Now collects ids from all three `playbackAssets` sources in addition to the legacy field, deduping against it (a video id shared across both only produces one manifest entry).

## Test plan
- [x] `server/services/sharing/peerSyncAssets.test.js` — 24/24, including 2 new tests: entry/hold/exit clips are all included, and a shared id across `videoHistoryId` and `playbackAssets` dedupes correctly

### DIFF
--- a/server/services/sharing/peerSyncAssets.js
+++ b/server/services/sharing/peerSyncAssets.js
@@ -565,8 +565,22 @@ export async function buildFableLoomAssetManifest(loom) {
   for (const entry of await Promise.all(imageNames.map((name) => hashImageForManifest(name)))) {
     if (entry) dedup.set(`${entry.kind}:${entry.filename}`, entry);
   }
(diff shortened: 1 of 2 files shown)


atomantic (approved):
Reviewed against the live schema/behavior — change is correct, focused, and covered by tests. Approving.

PR #6142, opened 2026-09-03
feat: report Tribe emails/phones shared by more than one person

### LINKED ISSUE #5908: Two Tribe people sharing a phone or email silently swap identifier ownership as last_contact_on advances
> **Labels to apply** (the filing account has read-only access to this repo, so GitHub silently drops them at creation): `bug`, `plan`, `area:life-tracking`, `severity:low`, `effort:medium`, `model:medium`.

**How this was found.** Read off the code on `main` at `d78a778` (still current for every file cited), not reproduced on a live install. It came up while tracing Tribe identity matching for an unrelated integration proposal (#5872). The mechanism below is traced end to end in the source, but nobody has watched it happen, so treat the scenario as predicted rather than observed.

## Problem

`tribe_people` has no uniqueness on `emails[]` or `phones[]`. `server/lib/db/schema/tribe.js:30-33` declares four indexes, all non-unique. `server/scripts/init-db.sql:144-146` declares three of them — it carries no `phones` column at all, since that DDL lives in `ensureSchema()` and is registered by the stub `scripts/migrations/163-tribe-people-phones.js`. Neither path declares a unique index on `emails[]` or `phones[]`, and no migration adds one.

Nothing checks across rows either. `normalizeEmails` / `normalizePhones` (`server/services/tribe.js:59-83`) dedupe *within* one person's array and never read another row. The route schema (`server/routes/tribe.js:28,32`) validates shape and length only. So "one identifier, one person" is a convention the matcher assumes and nothing enforces.

When two people do share one, `buildPersonMatchIndex` keeps whichever it sees first: `server/lib/tribeMatch.js:74,78`, `if (key && !byIdentifier.has(key)) byIdentifier.set(key, person.id)`.

Order comes from `listPeople` (`server/services/tribe.js:215-218`):

```sql
ORDER BY <ring CASE>, COALESCE(last_contact_on, DATE '1900-01-01') ASC, name ASC
```

That is fully deterministic for fixed data. The problem is that `last_contact_on` is *also* what auto-logging advances (`tribe.js:434`, a `GREATEST(...)` write). So a routine sync re-sorts the pair and hands the identifier to the other person, with no log line and no user-visible event.

**Worked example.** Person A (ring `core`, last contact 2026-01-10) and person B (ring `core`, last contact 2026-06-01) both list the same phone number — a shared household or work line is the realistic way this happens. A sorts first, so A owns the number. An inbound message logs a touchpoint against A and advances A's `last_contact_on` to today. On the next sync A now sorts *after* B, the number maps to B, and every later touchpoint lands on B instead.

Both people's cadence is then wrong, and neither write is reversible: the touchpoint insert is dedupe-keyed `ON CONFLICT ... DO NOTHING` (`tribe.js:409-410`), and the date update is `GREATEST`, so nothing rolls back.

Uncovered by tests: `server/lib/tribeMatch.test.js:11-18` has fixtures for duplicate *names* (handled explicitly) but none sharing an identifier.

## Chosen approach

**A non-blocking detector, not a `UNIQUE` retrofit.**

Retrofitting uniqueness onto two `TEXT[]` columns needs a forward-only transform under `scripts/migrations/` plus a merge UI to resolve whatever it finds. That is much larger work, and its sensible first step is knowing whether any real install has duplicates at all. If they turn out to be common, the constraint is a follow-up with counts attached.

## Affected paths

- `server/services/tribe.js` — detection beside `listPeople` (`:191-222`), reusing the normalizers at `:59-83`
- `server/lib/tribeMatch.js` — reuse `normalizeIdentifier` (`:14`) and `normalizePhone` (`:34`); no behaviour change here
- `server/routes/tribe.js` — expose the report
- `client/src/pages/Tribe.jsx` — surface it

## Acceptance criteria

- [ ] A server-side check reports every identifier held by more than one non-deleted person, separately for `emails` and `phones`, returning the identifier and the colliding person ids.
- [ ] Comparison uses the matcher's own normalization — `normalizeIdentifier` for emails, `normalizePhone` for phones — so `Alice@Example.com` vs `alice@example.com`, and `(555) 010-0199` vs `+15550100199`, are caught rather than missed.
- [ ] The Tribe page surfaces the result non-blockingly: a dismissible banner or row affordance naming the colliding people. It must not block saving and must not auto-merge.
- [ ] Zero duplicates renders nothing at all.
- [ ] Tests cover a case-only email duplicate, a formatting-only phone duplicate, a soft-deleted person excluded from the report, and the zero-duplicate case.

## Out of scope

- Any `UNIQUE` constraint or index on `tribe_people.emails` / `.phones`. Note the `phones` DDL lives in `ensureSchema()` behind the stub `scripts/migrations/163-tribe-people-phones.js`, so adding one there is not the one-line change it looks like.
- Any migration that rewrites existing rows.
- A merge UI. If the detector shows this is real, file that separately with the counts attached.

Bryandero98:
## Summary
- `tribe_people` has no uniqueness on `emails[]`/`phones[]`, and nothing checks across rows. When two people share an identifier, `buildPersonMatchIndex` keeps whichever it sees first — and `listPeople`'s sort key includes `last_contact_on`, which routine auto-logging advances, so a shared identifier can silently re-target from one person to the other on a later sync, with no log line and no user-visible event.
- Adds a **non-blocking detector**, not a `UNIQUE` retrofit — that's explicitly out of scope per the issue (it needs a forward-only data migration plus a merge UI; the sensible first step is knowing whether real installs have any duplicates at all).
- `tribe.js`: `findDuplicateTribeIdentifiers()` (pure) groups people by normalized email/phone, reusing the exact `normalizeIdentifier`/`normalizePhone` the matcher uses — so a case-only email variant or a formatting-only phone variant collides here exactly as it would during identity matching. `checkDuplicateTribeIdentifiers()` wraps it over `listPeople()`, which already excludes soft-deleted rows.
- `GET /api/tribe/duplicate-identifiers` exposes the report.
- `Tribe.jsx` surfaces it as a dismissible `Banner` naming the colliding people, above the fold. Renders nothing when the report is clean. Never merges, edits, or blocks a save.

## Test plan
- [x] `tribe.test.js` — new `findDuplicateTribeIdentifiers` suite: case-only email duplicate, formatting-only phone duplicate, a person not flagged against themself, a clean roster, an empty roster. Plus `checkDuplicateTribeIdentifiers` confirming it loads through `listPeople()` (`deleted = FALSE` already applied).
- [x] `routes/tribe.test.js` — new route test for `GET /duplicate-identifiers`.
- [x] `Tribe.test.jsx` — new suite: clean report renders nothing, a shared email names both people, dismiss hides the banner.
- [x] `npx vitest run services/tribe.test.js routes/tribe.test.js` (server) and `npx vitest run src/pages/Tribe.test.jsx src/services/index.test.js` (client) — all green, no regressions.
- [x] `biome lint` clean on all changed client files.

Closes #5908


Bryandero98:
CI is green end to end now (Server tests, Client tests, CI Gate all SUCCESS) after syncing to the latest `apiRouteCatalog.generated.json` regen. Ready for review.

PR #6174, opened 2026-09-04
fix: launch the Windows self-update through a real detached PowerShell chain

### LINKED ISSUE #6169: Windows: in-app update never runs the script but records a successful update
## Symptom

On Windows, **Update Now** (Settings/App Management > PortOS > Update) starts, the progress panel shows only `Starting update` (spinning forever) followed by `Complete — Update complete – restarting`, and **Last Update** records `vX.Y.Z – Success`. The install is still on the old version and `Update Now` is still offered. Nothing between `starting` and `complete` is ever reported, and `data/update.log` is not touched at all.

## Root cause

`executeUpdate()` (`server/services/updateExecutor.js`) launched the script on Windows with

```js
spawn('powershell', ['-NoProfile', ..., '-File', UPDATE_PS1], { detached: true, stdio: ['ignore','pipe','pipe'], cwd, env })
```

Node's `detached: true` maps to **`DETACHED_PROCESS`** on Windows, which gives the child no console. `powershell.exe` is a console host: with no console it exits **0 in ~100ms without running a line** of `update.ps1`. Reproduced directly:

```
$ node probe.mjs           # spawn powershell -File <script>, detached: true
[75ms] CLOSE code=0 sig=null          # no stdout at all
$ node probe2.mjs          # same spawn WITHOUT detached
[241ms] OUT: "STEP:git-pull:running:..."
...
[4346ms] CLOSE code=0 sig=null
```

`executeUpdate` then treated exit 0 as success and, with `data/update-complete.json` absent, stamped the **triggering tag** onto the recorded result — so the UI reported a successful update to a release the install had never run.

Dropping `detached` is not sufficient either: PM2 kills on Windows with `taskkill /pid <app> /T /F`, which walks the same parent/child tree the POSIX double-fork escapes, so an attached `update.ps1` would be killed at its own `pm2-stop` step and leave the install headless (the #5976 failure).

## Fix

- `spawnDetached` gains an opt-in `windowsDetached` launch: a short-lived PowerShell launcher `Process.Start`s a supervisor (with `CreateNoWindow`) and exits, so the supervisor's parent is gone and `taskkill /T` from the server never reaches it. The supervisor redirects the job's output into the control dir and records its pid + exit status, so the existing tailer streams `STEP:` progress identically on both platforms.
- `executeUpdate` uses `spawnDetached` on both platforms and runs the still-running-script guard on Windows too.
- A run that exits 0 having emitted **no `STEP:` line at all** is recorded as a failure — both scripts emit `git-pull:running` before touching anything.
- When the completion marker is missing, the recorded version comes from `package.json` on disk, never from the triggering tag.
- The synthetic `starting` step is closed out once the script reports real progress, instead of spinning for the whole update.
- The client's post-restart health poll ceiling was 60s, far shorter than the window the server is actually down (`pm2-stop` through the client build), so a real update timed out with "Restart timed out" while it was still running fine.

Bryandero98:
## Summary

Fixes #6169: on Windows, `executeUpdate` launched `update.ps1` via `spawn(cmd, args, { detached: true })`. Node's `detached: true` maps to `DETACHED_PROCESS` on Windows, which gives `powershell.exe` no console — a console-less `powershell.exe` exits in ~100ms without running a single line of the script, yet the exit code (0) still read as success. The update never actually ran, but the UI reported "Update complete."

- `spawnDetached` (`server/lib/detachedSpawn.js`) gains an opt-in `windowsDetached: true` mode: a two-hop PowerShell launcher (`server/lib/windowsDetachedLauncher.ps1`). Hop 1 (a normal, attached spawn — never `detached: true`) starts hop 2 via `Start-Process -WindowStyle Hidden` and returns immediately; hop 2 runs the actual job with output redirected straight into the control dir's log files, records its PID, and waits for its exit. By the time pm2's own `taskkill /T /F` fires (at the update script's own `pm2-stop` step, well after hop 1's brief lifetime), hop 1 no longer exists for that tree-walk to find, so hop 2 survives — matching the POSIX double-fork's existing survival guarantee.
  - `Start-Process -ArgumentList` does not reliably quote array elements containing spaces (e.g. `C:\Program Files\nodejs\node.exe`) — the launcher quotes every argument itself (Microsoft's documented C/C++ argv escaping rule) and passes a single pre-quoted string instead.
  - `isDetachedRunning`'s executable/args identity check shells out to POSIX `ps`, unavailable on Windows — it now trusts the PID + exit sentinel alone there, matching this module's existing single-user trust model.
  - The PID-wait timeout is longer on Windows (30s vs 10s) since the chain costs two cold `powershell.exe` starts before the job's own PID is even written.
- `updateExecutor.js` now launches through `spawnDetached` on both platforms (previously POSIX-only) and:
  - treats a clean exit that never emitted a single `STEP:` line as a failure — both scripts emit `git-pull:running` before touching anything, so silence proves the script never ran, regardless of exit code;
  - falls back to a fresh on-disk `package.json` read (via `getCurrentVersion()`) instead of the triggering tag when the completion marker is missing — the tag was only ever a request, never proof of what actually landed;
  - closes out the synthetic `starting` step once real progress arrives, instead of leaving it shown as spinning for the whole update.
- `usePortosRestartWatch`'s post-restart health-poll ceiling moves from 60s to 5 minutes — the window the server is actually down for spans `pm2-stop` through `npm-install` and the client build, commonly well past a minute on a cold dependency install.

## Test plan

- `server/lib/detachedSpawn.test.js` — new `windowsDetached (real Windows launcher)` suite, gated on the real OS (not `pinPlatform`, since it needs genuine `powershell.exe`): end-to-end pid/exit/stdout correctness, and — the core claim — the job **survives a real `taskkill /T /F`** against its own spawning process once the launcher has naturally exited (simulating pm2's tree-kill).
- `server/services/updateExecutor.test.js` — updated for the new Windows launch path and the `onLaunched` callback landed by #6166; new coverage for the no-`STEP:`-lines failure and the `starting` step close-out.
- Manually verified on this Windows checkout, outside the automated suite: the exact `Start-Process -ArgumentList` array-quoting bug (a job binary path with a space silently failing) reproduced and was fixed; the two-hop chain's job survives a `taskkill /T /F` issued against a stand-in "portos-server" root process once the launcher has exited, confirmed via `tasklist` and by reading the control dir's `pid`/`exit`/`stdout.log` to completion.
- `npx vitest run` (server): `lib/spawnCwd.test.js`, `lib/detachedSpawn.test.js`, `routes/update.test.js`, `services/appUpdater.test.js`, `services/updatePreflightParity.test.js`, `services/updateExecutor.test.js`, `services/portosSelfUpdate.test.js`, `services/dataManagerBusy.test.js`, `sockets/apps.test.js` → all green.
- `npx vitest run` (client): `components/apps/tabs/UpdateTab.test.jsx`, `hooks/useAppOperation.test.jsx`, `components/apps/tabs/RepositorySourcePanel.test.jsx` → all green.
- `npx biome lint` on the touched client files → clean.

Closes #6169


Bryandero98:
Closing — atomantic landed their own fix for #6169 on main (fbe172acc + a1ae0663f) while this PR was open, using the same core architecture (a short-lived PowerShell launcher spawning a supervisor that survives the tree-kill) but more thoroughly reviewed against real Windows hardware than mine was. Their version catches at least two real defects mine still had:

- `-WindowStyle Hidden` alone doesn't set `CreateNoWindow` — only `-NoNewWindow` does — so my launcher was still allocating a `conhost.exe` per job. Theirs uses an explicit `[System.Diagnostics.ProcessStartInfo]` with `CreateNoWindow = $true` to avoid it.
- `$p.ExitCode` is a signed int in PowerShell but unsigned in Node's own `ChildProcess`, so an NTSTATUS crash code would have come out negative from my launcher and disagreed with the POSIX arm. Theirs masks it correctly.

No point carrying a less-complete duplicate. Thanks for the fix.

PR #6178, opened 2026-09-04
test: pin usePortosRestartWatch restart proofs, phase budgets, and socket arming paths

### LINKED ISSUE #6173: usePortosRestartWatch has no test file — the restart budgets and failure messages are unpinned
## Context

`client/src/hooks/usePortosRestartWatch.js` owns the whole post-update restart handoff for every PortOS self-update surface — arming on the update stream, health polling, the down→up / uptime-drop restart proofs, the reload, and the timeout. It has **no `.test.js` sibling**.

#6169 changed its timeout from a flat 30-attempt (60s) ceiling to two wall-clock phase budgets, and split the failure into two messages. None of that is pinned by a test, and neither were the pre-existing restart-detection branches.

## What to cover

A `client/src/hooks/usePortosRestartWatch.test.js` (happy-dom, fake timers, `api.checkHealth` mocked) covering the state machine's distinct outcomes:

- **Version change → success.** `preUpdateVersion` set, health returns a different `version` ⇒ polling stops, success toast, reload scheduled.
- **Same-version reconcile via a down→up dip.** Health fails once, then succeeds with the same version ⇒ "Install reconciled" path.
- **Same-version reconcile via an uptime drop** below `maxUptimeRef - UPTIME_DROP_SLACK_S`, with no sampled down window (the restart that is too fast for the 2s poll).
- **Phase 1 budget:** armed WITHOUT a confirmed down observation, health stays up and unchanged ⇒ errors at `DOWN_WAIT_MS` with the "PortOS never went down" message, not the generic one.
- **Phase 2 budget:** armed with `healthDown: true` ⇒ still polling well past `DOWN_WAIT_MS`, and errors at `RECOVERY_TIMEOUT_MS` with "Restart timed out". This is the regression #6169 fixed — a flat 60s ceiling made every real update time out mid-`npm install`.
- **Budgets are wall clock, not tick counts.** Advance the clock past a budget while suppressing ticks (the hidden-tab case `useAutoRefetch` produces) and assert the next tick can still recognize a completed restart before the timeout fires — the success checks are deliberately evaluated before the budget.
- **`arm()` is idempotent** while already polling, and `stopPolling` tears the watch down.

## Notes

Use fake timers; do not sleep real budgets — 30 minutes of wall clock is the whole point of the constant. The hook reads `Date.now()` directly, so `vi.setSystemTime` / `vi.advanceTimersByTime` drive both the interval and the budget.

Filed because #6169's author could not run the client suite locally (`happy-dom` was not installed in that checkout), so shipping an unverified test would have been worse than filing this.

atomantic:
## Summary

`usePortosRestartWatch` owns the whole post-update restart handoff for every PortOS self-update surface and shipped without a test sibling. This adds `client/src/hooks/usePortosRestartWatch.test.js` (21 cases, happy-dom + fake timers, `api.checkHealth` mocked). No production code changes.

Covered:

- **Restart proofs** — version change, same-version reconcile via a down→up dip, same-version reconcile via an uptime drop measured against the `captureBaseline` peak (the restart too fast for the 2s poll to sample a down window), and a within-slack uptime wobble that must *not* read as a restart.
- **The two phase budgets from #6169** — errors at `DOWN_WAIT_MS` with the "never went down" message when the server was never observed down; keeps polling well past that and errors only at `RECOVERY_TIMEOUT_MS` with "Restart timed out" once it was. That second one is the regression #6169 fixed: a flat 60s ceiling timed out every real update mid-`npm install`.
- **Budgets are wall clock, not tick counts** — ticks suppressed across a hidden-tab gap that outlasts the budget, then the first tick back still recognizes the finished restart, because the success checks are evaluated before the budget check.
- **The socket arming paths every real caller uses** — no production surface arms by hand; `UpdateTab` and `useAppOperation` pass `active` and let the handlers decide. Includes the disconnect confirmation and its negative case (a Tailscale blip fires `disconnect` on a live server; arming there parks a bogus "restarting" toast and then fails the update), the errored step, `complete({ success: false })`, the `active: false` no-op, and unmount inside the 1.5s confirmation window.
- **Lifecycle** — `arm()` idempotent while polling with the budget still measured from the first arm, and the watch stops polling once the restart is confirmed.

Every case was mutation-probed against the hook: removing or inverting the branch it names makes exactly that test fail.

## Test plan

- `cd client && npx vitest run src/hooks/usePortosRestartWatch.test.js` — 21 passed.
- Full client suite green (836 files / 10276 tests). Two later runs on a loaded machine hit 5s-timeout flakes in `a11yConventions.test.js` and `pages/loadingSkeletons.test.jsx`; both pass in isolation and were green in the earlier full runs — unrelated to this branch, which adds one isolated test file.
- No real wall clock is spent: fake timers drive both the 2s poll and the 30-minute budget.

Closes #6173

### DIFF
--- a/client/src/hooks/usePortosRestartWatch.test.js
+++ b/client/src/hooks/usePortosRestartWatch.test.js
@@ -24,6 +24,7 @@ import { usePortosRestartWatch } from './usePortosRestartWatch';
 // contract this file pins, so importing them would let a change to the source
 // silently move the assertions with it.
 const POLL_MS = 2000;
+const DISCONNECT_CONFIRM_MS = 1500;
 const DOWN_WAIT_MS = 60 * 1000;
 const RECOVERY_TIMEOUT_MS = 30 * 60 * 1000;
 const TOAST_ID = 'portos-update-restart';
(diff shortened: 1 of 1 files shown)


atomantic:

atomantic:
Review trail for this run (configured reviewers: `codex`, `claude`):

- **codex — BLOCKED, no verdict.** Failed on its first invocation with an account-level usage limit (resets Sep 7 2026). It is a required reviewer, so this is `review-blocked`, not a clean pass, and no self-review was substituted for it.
- **claude — APPROVE, converged over 2 rounds.** Round 1 raised three findings, all fixed in 9ea02c7: `arm({ healthDown: true })` was inert (the only test passing it also made the poll observe the down window itself, so the carry-forward proved nothing); no socket arming path was covered even though no production caller arms by hand; and `captureBaseline`'s caller-version precedence was unpinned. Round 2 confirmed the fixes and named two optional gaps, both closed in 2c44489.

Merge is withheld pending the codex review.

Issue #6190, opened 2026-09-04
Migrate the LoRA download onto the shared download slot so the orphaned-partial GC protects it


atomantic:
## What

`downloadToFile` in `server/services/loras.js` is the third `streamResumableDownload` caller and the one still hand-rolling the machinery #6168 centralised: its own `lastEmit` progress throttle, a caller-supplied `signal` with `isCancelled: () => Boolean(signal?.aborted)`, no in-flight registry, and no idle-stall watchdog.

Claim a slot for it — `createDownloadSlot({ codePrefix: 'LORA' })` from `server/lib/downloadPreflight.js` — and use `slot.downloadOptions()` / `slot.throttle()` / `slot.wrapError()` in place of the hand-rolled versions.

## Why

`PATHS.loras` is the **first** directory `collectPartialSweepDirs()` sweeps (`server/services/orphanedPartialGc.js`), but a LoRA install registers no slot, so `isAnyDownloadInFlight` answers false for its live `.partial`. The 7-day age gate makes that harmless today — a fresh partial is never old enough — but it is the one download path the sweep can reach and cannot see, and the guard in `orphanedPartialGc.test.js` filters on `createDownloadSlot(` callers, so it is exactly the path that guard cannot fail on.

Migrating it also gets the LoRA install an in-flight registry (a second press on the same destination currently starts a parallel transfer of the same file) and the idle-stall watchdog that the other two runtimes have.

## Decisions already made

- **Key the slot on the resolved destination path**, not the CivitAI/HF id — that is what the sweep asks about, and it is what spec-decode does.
- **Not `exclusive`.** LoRA files are small relative to a checkpoint; a single global lock would be a behaviour regression for a user installing several.
- **`keepPartialOnCancel: false`**, matching spec-decode: a single abandoned file, not a multi-shard checkpoint.

## Acceptance criteria

- `loras.js` claims a slot before its first await and releases it in a `finally`.
- `isAnyDownloadInFlight` returns true for a live LoRA download's dest and its `.partial`; a test proves the sweep keeps that file even when its mtime is aged past the gate (mirror the Slotstream case in `server/services/orphanedPartialGc.test.js`).
- The existing LoRA install/download suites stay green with no test rewritten to match a changed shape.
- The "still unprotected" notes in `server/lib/downloadPreflight.js` and `server/services/orphanedPartialGc.js` that point at this issue are removed.

## Files / areas

[...]

PR #6195, opened 2026-09-04
Extract one shared weight-download slot, and give Slotstream the cancel it lacked

### LINKED ISSUE #6168: Extract the shared weight-download slot (in-flight guard, stall watchdog, progress throttle) into one helper
## What

`server/services/specDecodeModels.js` and `server/services/slotstreamModelManager.js` each carry their own copy of the same download-slot machinery, and the pattern is now on its third or fourth appearance across the local-runtime download paths (llama.cpp spec-decode, MTPLX, Slotstream):

- an `inFlight` Map keyed by destination, claimed before the first `await`
- a `{ controller, ... }` state object with an `AbortController`
- an idle-stall watchdog wired to `onIdleStall`, each with its own `*_IDLE_STALL_MS` env knob
- a `PROGRESS_INTERVAL_MS` frame throttle
- a bespoke "download stalled / cancelled" `ServerError` builder
- an `isXDownloadInFlight(path)` predicate that `server/services/orphanedPartialGc.js` calls to protect a live transfer from the sweep

## Why

Each copy is a place the behavior can drift, and the drift is not cosmetic: Slotstream shipped with an unreachable cancel path precisely because the cancel lives in the *caller* rather than in the slot (`slotstreamModelManager.js` had `isCancelled` / `'cancelled'` branches with nothing to set them, since it has no `cancelSlotstreamModelDownload` the way spec-decode does). A shared slot would have carried the cancel with it. The GC's `isProtected` predicate also has to grow one more `|| isXDownloadInFlight(path)` clause per runtime, which is exactly the kind of per-caller wiring that gets forgotten.

## Scope

- Add `createDownloadSlot({ idleStallTimeoutMs, codePrefix })` (or equivalent) to `server/lib/downloadPreflight.js` — it already owns `streamResumableDownload`, the stall timeout plumbing, and `sweepOrphanedPartials`, so the slot belongs beside them. It should expose: claim/release, the abort controller and its cancel + stall reasons, the typed stalled/cancelled errors, the progress throttle, and one `isInFlight(path)` predicate.
- Migrate `specDecodeModels.js` and `slotstreamModelManager.js` onto it, keeping each service's existing public functions and error codes so their routes and tests do not change shape.
- Collapse the GC's `isProtected` to the shared predicate.
- Give Slotstream the cancel it currently lacks, since the slot now provides it: `cancelSlotstreamModelDownload({ model })`, a route, and a Cancel button on the card beside the progress bar. A 100 GB+ transfer that is merely *slow* rather than stalled is otherwise only stoppable by waiting out the 20-minute idle watchdog.
- Add the row + barrel export required by the module-organization rule in `AGENTS.md`.

## Acceptance criteria

- One implementation of the slot; no service keeps its own `inFlight` Map, `AbortController` state, or stall-error builder.
- Existing behavior preserved: the same error codes (`SPEC_DOWNLOAD_IN_FLIGHT`, `SLOTSTREAM_DOWNLOAD_IN_FLIGHT`, the stalled variants), the same "claim before the first await" race guarantee, and the same GC protection — with the existing suites still green and no test rewritten to match a changed shape.
- A Slotstream download can be cancelled from **Models → LLMs**, and a cancelled transfer keeps its `.partial` so a retry resumes.

## Files / areas

`server/lib/downloadPreflight.js`, `server/lib/README.md`, `server/lib/index.js`, `server/services/specDecodeModels.js`, `server/services/slotstreamModelManager.js`, `server/services/orphanedPartialGc.js`, `server/routes/localLlm.js`, `client/src/components/settings/SlotstreamServerCard.jsx`.

Refs #6162

atomantic:
## Summary

- **One implementation of the download slot.** `createDownloadSlot()` in `server/lib/downloadPreflight.js` now owns the in-flight registry claimed before the first await, the `AbortController` carrying a cancel/stall reason, the idle-stall watchdog, the progress throttle, the typed `<PREFIX>_DOWNLOAD_IN_FLIGHT` / `_STALLED` / `_CANCELLED` errors, and the `isInFlight(path)` predicate. `specDecodeModels.js` and `slotstreamModelManager.js` keep their public functions and error codes; no test was rewritten to match a changed shape.
- **Slotstream can be cancelled.** The drift this removes was not cosmetic — Slotstream shipped `'cancelled'` branches nothing could set, because the cancel lived in spec-decode's *caller* rather than in the slot. It now has `cancelSlotstreamModelDownload`, a route, and a Cancel button beside the progress bar in **Models → LLMs**. A 100 GB+ transfer that is merely slow rather than stalled was previously only stoppable by waiting out the 20-minute idle watchdog; a cancelled transfer keeps its `.partial`, so downloading again resumes.
- **The GC protects live transfers with one predicate.** Every slot registers itself, so `orphanedPartialGc.js`'s `isProtected` is `isAnyDownloadInFlight` instead of one `|| isXDownloadInFlight(path)` clause per runtime.

Closes #6168

## Review status — intentionally not merged

`codex` is a **required** reviewer for this run and produced no verdict: it hit its account usage limit on the first invocation ("try again at Sep 7th, 2026"). Under the required-review publication rule that is `review-blocked`, so this PR is left open pending that review. `claude` reviewed the diff and returned clean; a four-agent `/simplify` pass ran first and its fixes are committed in `719dad1ff`.

## Test plan

- `server`: `specDecodeModels`, `slotstreamModelManager`, `downloadPreflight`, `routes/localLlm`, `lib/index`, `lib/validationSplit`, `lib/generatedManifests`, `lib/importScoping` — all green.
- `client`: `SlotstreamServerCard`, `LocalLlmRuntimesView` green; `npm run lint` clean.
- New `server/services/orphanedPartialGc.test.js`: the sweep keeps a live Slotstream shard aged past the age gate **without naming Slotstream**, and a structural guard fails if a module calling `createDownloadSlot` stops being reachable from the GC (the module load is what registers its slot). Verified the guard bites by removing the import.
- New Slotstream cases: a cancel emits a terminal `cancelled` frame, returns `SLOTSTREAM_DOWNLOAD_CANCELLED`, and leaves the shard's `.partial` on disk.
- Full `server` and `client` suites were run; the failures seen there are load-timeout flakes unrelated to this change — each passes in isolation, and the sets differ between runs.

## Known gap, filed not fixed

`loras.js` is the third `streamResumableDownload` caller and claims no slot, so the sweep can reach its `.partial` files and not see them (harmless today behind the 7-day age gate). The comments say so rather than overclaiming, and #6190 migrates it.


atomantic:

atomantic:
Review audit trail for this run:

- **codex** (required) — `review-blocked`. Failed on its first invocation with `ERROR: You've hit your usage limit ... try again at Sep 7th, 2026`. The binary is on PATH and healthy; the account is out of credits, so no verdict was produced. No self-review was substituted for it.
- **claude** (required) — clean. Reviewed the full `origin/main...HEAD` diff for correctness; reported no bugs, having specifically traced slot claim/release identity, cancel-vs-stall abort labelling, `keepPartialOnCancel` through `streamResumableDownload`, `isInFlight` path matching, and the two service migrations line by line against `origin/main`.
- **self-review** — a four-agent `/simplify` pass (reuse, simplification, efficiency, altitude) ran before the reviewers; its fixes are committed in `719dad1ff` (trimmed the slot API to what callers use, stopped building progress frames the throttle discards on every stream chunk, made the cancel take a required model, reused the shared source-scanning test helper, and narrowed an overclaim about GC protection into #6190).

The claim markers, branch, and worktree are preserved for the pending codex review.

Issue #6023, opened 2026-09-03
PromptManager: Variables tab Save abruptly deselects and blanks editor without confirmation; switching variables discards edits


atomantic:
## Problem
In `client/src/pages/PromptManager.jsx`, the Variables tab (`tab === 'variables'`) suffers from disruptive form-clearing on save, missing confirmation feedback, and silent loss of unsaved edits on selection change:

1. **Saving an existing variable abruptly clears the form and deselects the variable** (`client/src/pages/PromptManager.jsx:237-249`):
   ```javascript
   const saveVariable = async () => {
     setSaving(true);
     const ok = await (selectedVar
       ? savePromptVariable(selectedVar, varForm, { silent: true })
       : createPromptVariable(varForm, { silent: true }))
       .then(() => true)
       .catch((err) => { toast.error('Failed to save variable: ' + err.message); return false; });
     if (!ok) { setSaving(false); return; }
     setSaving(false);
     setSelectedVar(null);
     setVarForm({ key: '', name: '', category: '', content: '' });
     await loadData();
   };
   ```
   When editing an existing variable `foo` and clicking "Save":
   - `{ silent: true }` suppresses the default toast, and no `toast.success` is called.
   - `setSelectedVar(null)` removes `?var=` from the URL.
   - `setVarForm({ key: '', name: '', category: '', content: '' })` blanks all fields.
   - The editor title abruptly changes from `Edit: foo` to `New Variable`.
   - The user receives no confirmation that their changes persisted, and their editing context is destroyed. To verify or tweak what they just saved, they must find `foo` in the list and click it again.

2. **Creating a new variable does not select the created variable** (`client/src/pages/PromptManager.jsx:246`):
   Creating a new variable resets the form to blank with no toast, rather than selecting the newly created variable (`setSelectedVar(varForm.key)`).

3. **Deleting a variable lacks success feedback** (`client/src/pages/PromptManager.jsx:251-258`):
   `deletePromptVariable` resolves with no confirmation toast.

4. **Switching variables or clicking "Add variable" silently discards unsaved edits** (`client/src/pages/PromptManager.jsx:812-815`, `260-263`):
   Clicking another variable in the list (`onClick={() => setSelectedVar(key)}`) or clicking "+" (`newVariable()`) immediately overwrites `varForm` without checking if there are unsaved edits, wiping out any in-progress typing.

## Trigger
1. In `/prompts?tab=variables`, select an existing variable (e.g. `Tone Guide`).
[...]

PR #6198, opened 2026-09-04
close the PromptManager variable race that the #6189 fix only relocated

### LINKED ISSUE #6189: flaky: PromptManager 'selects the created variable in the URL' asserts editor content behind a URL-only barrier
## Problem

`client/src/pages/PromptManager.test.jsx` → `selects the created variable in the URL and confirms with a toast` fails intermittently in full-suite CI:

```
AssertionError: expected '' to be 'wink' // Object.is equality
 ❯ src/pages/PromptManager.test.jsx:562:32
```

Observed on a PR whose diff is nowhere near `PromptManager` (voice confirmation gate). Passes locally 71/71. It fails the whole `Client tests and build` matrix, because the shard that loses cancels its siblings.

## Root cause

The test's only async barrier before the two final assertions is the URL:

```js
await waitFor(() => expect(currentSearch()).toContain('var=aside-voice'));
expect(screen.getByText('Edit: aside-voice')).toBeTruthy();
expect(contentBox().value).toBe('wink');
```

The URL update and the editor's hydration from the post-create refetch are separate effects. `var=aside-voice` landing in the URL does not prove the editor has hydrated from the refetched variable list, so on a loaded 2-core runner the bare synchronous `contentBox().value` read can run against a still-empty textarea.

This is the "`findBy*`/`waitFor` resolves on the wrong signal" shape already documented for `TracksManager.test.jsx` — wait on the settled STATE, not on a different observable that merely tends to arrive after it.

## Fix

Wrap the state assertion in the barrier instead of adding a second one after it:

```js
await waitFor(() => expect(contentBox().value).toBe('wink'));
expect(currentSearch()).toContain('var=aside-voice');
expect(screen.getByText('Edit: aside-voice')).toBeTruthy();
```

`contentBox().value` is the last thing to settle, so waiting on it subsumes the URL wait.

## Verify

Resolve the mocked `getPromptVariables` refetch behind a `setTimeout(…, 30)` and run the file — the current version fails with the exact CI signature, the fixed version passes. Restore the fast mock afterward. (Do not try to induce it with CPU load; it will not reproduce on a many-core dev box.)

## Also worth checking in the same file

`keeps the saved variable open in the editor and confirms with a toast` (line ~525) has the same shape — it waits on `toast.success`, then reads `contentBox().value` synchronously. It has not been seen failing, but the barrier is equally indirect.

### LINKED ISSUE #6197: flaky: the #6189 fix moved PromptManager's race to the URL instead of closing it, and left the sibling test blind
PR #6194 shipped the one-line fix #6189 prescribed. It does not close the race, and the second test #6189 flagged as "also worth checking" was left untouched — a bypass probe shows it never observed the state it guards.

## 1. The new barrier is already open before the save lands

`client/src/pages/PromptManager.test.jsx` → `selects the created variable in the URL` now reads:

```js
await waitFor(() => expect(contentBox().value).toBe("wink"));
expect(currentSearch()).toContain("var=aside-voice");
```

`contentBox().value` is `"wink"` from the `fireEvent.change` earlier in the test — the editor holds the user's own draft, not the refetched record. So that `waitFor` resolves on its first check, before `createPromptVariable` has even settled, and `currentSearch()` becomes the unguarded racing read. The barrier moved; it did not close.

**Repro:** resolve the mocked `getPromptVariables` refetch behind `setTimeout(…, 30)` and run the file.

```
AssertionError: expected '?tab=variables' to contain 'var=aside-voice'
```

The original failure reproduces the same way, deterministically even on a many-core box, by sampling `contentBox().value` at the first DOM mutation where the URL matches (which is what RTL's MutationObserver-driven `waitFor` does): it reads `""`.

**Why neither half works alone.** In `saveVariable()` (`client/src/pages/PromptManager.jsx`), the post-create refetch re-runs the hydration effect while `selectedVar` is still `null`, hitting the `!selectedVar` branch that resets the form to `BLANK_VAR_FORM`; `setSelectedVar` runs only after `await loadData()`; re-hydrating from the refetched record is a further effect commit. So the URL leads the content in one interleaving, and the draft leads the URL in the other. Assert both in one `waitFor` — the pair is the settled state either way.

## 2. `keeps the saved variable open in the editor` is blind to the regression it guards

Same file, ~line 525. It waits on `toast.success`, which `saveVariable()` fires **before** `await loadData()` — i.e. before the refetch whose harmlessness is the entire point of the test. It therefore asserts on pre-refetch state.

**Verified with a bypass probe:** re-inject the #6023 regression by dropping the `varHydratedRef.current !== selectedVar` guard so the editor re-hydrates on every refetch. The test **passes**. The refetch needs to be observable — e.g. return one extra sibling variable and `await screen.findByText(...)` for it — after which the injected regression correctly fails the test.

## Acceptance criteria

- `selects the created variable in the URL` passes with the post-create refetch delayed by 30ms.
- `keeps the saved variable open in the editor` FAILS when the `varHydratedRef` guard in `PromptManager.jsx` is removed, and passes with it.
- Full file green; no change to `client/src/pages/PromptManager.jsx`.

## Files

- `client/src/pages/PromptManager.test.jsx`

Refs #6189

atomantic:
Follow-up to #6194, which shipped the one-line fix #6189 prescribed. That fix does not close the race, and the second test #6189 flagged as "also worth checking" was left untouched and is blind to the regression it guards.

## Summary

- **`selects the created variable in the URL`** — #6194 made it wait on `contentBox().value`, then read the URL synchronously. The textarea already holds `'wink'` from the `fireEvent.change` earlier in the test (it is the user's own draft, not the refetched record), so that barrier resolves on its first check — before the create has even settled — and `currentSearch()` becomes the unguarded racing read. Neither half settles last on its own: the post-create refetch re-runs the hydration effect while the selection is still `null`, blanking the form, and `setSelectedVar` runs only after `await loadData()`. Both assertions now sit inside one `waitFor`; the pair is the settled state under either interleaving.
- **`keeps the saved variable open in the editor`** — waits on `toast.success`, which `saveVariable()` fires *before* the refetch whose harmlessness is the whole point of the test, so it asserted on pre-refetch state. The post-save refetch now returns a sibling the list did not have, giving the refresh an observable arrival to wait on.

Test-only. No change to `client/src/pages/PromptManager.jsx`.

## Test plan

- `cd client && npx vitest run src/pages/PromptManager.test.jsx` → 71/71 pass.
- **Race closed:** resolve the mocked `getPromptVariables` refetch behind `setTimeout(…, 30)`. The version on `main` fails with `expected '?tab=variables' to contain 'var=aside-voice'`; this branch passes.
- **Original flake reproduced deterministically** (even on a many-core box): sample `contentBox().value` at the first DOM mutation where the URL contains `var=aside-voice` — which is what RTL's MutationObserver-driven `waitFor` does. It reads `""`, the CI signature from #6189.
- **Bypass probe on the second test:** re-inject the #6023 regression by dropping the `varHydratedRef.current !== selectedVar` guard in `PromptManager.jsx`. On `main` the test still **passes**; on this branch it fails with `expected 'stay wry' to be 'stay dry'`. Guard restored afterwards.
- Full client suite run: the only failures are pre-existing load flakes in `loadingSkeletons.test.jsx`, `Layout.test.jsx` and `TaskAddForm.test.jsx` (a different set on each run, 3 then 9); all four files pass in isolation alongside `PromptManager.test.jsx`.

Closes #6197
Refs #6189

### DIFF
--- a/client/src/pages/PromptManager.test.jsx
+++ b/client/src/pages/PromptManager.test.jsx
@@ -524,6 +524,10 @@ describe('PromptManager variable editing', () => {
 
   it('keeps the saved variable open in the editor and confirms with a toast', async () => {
     await openVariable();
+    // Re-stub AFTER the mount load, so only the post-save refetch carries the
+    // extra sibling — staging it with mockResolvedValueOnce instead would bake
+    // in an assumption about how many times the page loads before the save.
+    getPromptVariables.mockResolvedValue({ variables: { ...VARIABLES, 'aside-voice': { name: 'Aside Voice', content: 'wink' } } });
 
     fireEvent.change(contentBox(), { target: { value: 'stay dry' } });
     fireEvent.click(saveButton());
@@ -532,8 +536,12 @@ describe('PromptManager variable editing', () => {
     expect(savePromptVariable).toHaveBeenCalledWith(
       'tone-guide', expect.objectContaining({ content: 'stay dry' }), { silent: true },
     );
-    // The refetch below returns the pre-save fixture; the editor must NOT be
-    // re-hydrated from it, deselected, or blanked.
+    // The refetch still carries the PRE-save content, so the editor must NOT be
+    // re-hydrated from it, deselected, or blanked. The toast fires BEFORE that
+    // refetch, so waiting on it asserts ahead of the very state change under
+    // test — the sibling above is what makes the refresh's arrival observable,
+    // and #6023 slips through this test unnoticed without it (#6189).
+    await screen.findByText('Aside Voice');
     expect(currentSearch()).toContain('var=tone-guide');
     expect(screen.getByText('Edit: tone-guide')).toBeTruthy();
     expect(contentBox().value).toBe('stay dry');
@@ -557,12 +565,17 @@ describe('PromptManager variable editing', () => {
     expect(createPromptVariable).toHaveBeenCalledWith(
       expect.objectContaining({ key: 'aside-voice', content: 'wink' }), { silent: true },
     );
-    // The URL selection and the editor's hydration from the post-create
-    // refetch are separate effects, so `var=aside-voice` landing in the URL
-    // does not prove the editor has content yet. Wait on the value — the last
-    // thing to settle — and the URL assertion comes along for free.
-    await waitFor(() => expect(contentBox().value).toBe('wink'));
-    expect(currentSearch()).toContain('var=aside-voice');
+    // Both halves in ONE barrier: neither settles last on its own (#6189).
+    // The post-create refetch re-runs the hydration effect while the selection
+    // is still null, which resets the form to BLANK — the URL adopts the new key
+    // only afterwards, and re-hydrating from the refetched record is a further
+    // commit still. So waiting on the URL alone reads a blanked textarea. But
+    // waiting on the textarea alone is a barrier already open, since it holds
+    // 'wink' from the typing above, leaving the URL as the racing read instead.
+    await waitFor(() => {
+      expect(currentSearch()).toContain('var=aside-voice');
+      expect(contentBox().value).toBe('wink');
+    });
     expect(screen.getByText('Edit: aside-voice')).toBeTruthy();
   });
 


atomantic:

PR #6199, opened 2026-09-04
fix: keep a provider-config block paused instead of auto-completing it

### LINKED ISSUE #6193: A permanent provider-config block auto-expires to completed and loses its resume pointer
## Problem

`blockedCategory: 'provider-config'` is stamped by `server/services/agentLifecycle.js` when `resolveAgentProviderAndModel` returns a **permanent** failure — today that is an `api`-type provider (no file-writing harness) pinned to or resolved for an agent task. It is a config error the user has to fix; it re-fails identically on every re-dispatch, which is why the block exists at all.

But the category is in neither set in `server/lib/taskBlockCategories.js`, and both omissions bite:

1. **It auto-expires to `completed`.** `USER_DECISION_BLOCKED_CATEGORIES` gates `blockedFailureAgeMs` (`server/services/cosTaskStore.js:1059`). Without membership, `sweepResolvedFailureTasks` flips the task to `completed` with `resolution: 'auto-expired'` after `DEFAULT_FAILURE_TASK_MAX_AGE_MS` (14 days). Unfinished work the user never decided to drop is silently reported as done — and because the task record federates, that false `completed` propagates to every peer.

2. **It loses its resume pointer.** `PAUSED_BLOCKED_CATEGORIES` gates the pointer-stripping branch at `server/services/cosTaskStore.js:778`. Without membership, `blocked` is treated as terminal-and-spent, so `existingBranch` / `resumeWorktreePath` / `resumedFromAgentId` are deleted. When the user adds a CLI provider and unblocks the task, it restarts on a fresh branch and orphans the worktree its dead agent left behind — the exact recovery that mechanism exists for.

`app-unresolved` and `workspace-invalid` are in **both** sets for precisely this reason: they are config pauses the user repairs and then revives. `provider-config` is the same shape and was simply missed.

Surfaced while shipping #5993, which added `orchestration-lane-unavailable` to both sets. That fix deliberately did not touch its sibling `provider-config`, to keep the change scoped — this issue is that follow-up.

## Decision

Treat `provider-config` as a config pause, exactly like `app-unresolved` / `workspace-invalid`. It is not a TIMED cooldown (no `cooldownUntil`, nothing self-revives it), so it goes in the two non-timed sets only. Rejected alternative: leaving the auto-expiry and adding only the pause membership — a 14-day silent `completed` on work nobody dropped is the worse of the two bugs, and the module's own doc treats the two memberships as one decision.

## Fix

In `server/lib/taskBlockCategories.js`:

- Export the literal as a shared constant next to `ORCHESTRATION_LANE_BLOCKED_CATEGORY` (it is stamped in `agentLifecycle.js` and would otherwise be a third hand-written literal — the drift this module exists to prevent), and use it at the stamp site.
- Add it to `PAUSED_BLOCKED_CATEGORIES` with a comment saying why it is not TIMED.
- Add it to `USER_DECISION_BLOCKED_CATEGORIES` with a comment naming the 14-day sweep it is being exempted from.

## Acceptance criteria

- `server/services/cosTaskStore.test.js` gains two tests mirroring the `app-unresolved` and `orchestration-lane-unavailable` ones: a `provider-config` block keeps `existingBranch` / `resumeWorktreePath` / `resumedFromAgentId` through the block, and `sweepResolvedFailureTasks` leaves it `blocked` with no `resolution` after 90 days.
- Both tests must fail before the change (verify by removing the membership) — the existing suite passes today, so a test that passes either way proves nothing.
- `agentLifecycle.js` references the shared constant rather than the string literal; its source-scraping test in `server/services/agentLifecycle.test.js` is updated to match, as was done for the lane category.
- Full server suite green.

## Files

- `server/lib/taskBlockCategories.js` — both sets plus the new exported constant
- `server/services/agentLifecycle.js` — the stamp site (`blockedCategory: ... : 'provider-config'`)
- `server/services/agentLifecycle.test.js` — source-scrape assertion
- `server/services/cosTaskStore.test.js` — the two new tests

atomantic:
## Summary

A permanent `provider-config` block — stamped when an agent task resolves to an `api`-type provider that has no file-writing harness — was in neither set in `server/lib/taskBlockCategories.js`, so two automations mishandled it:

- **It auto-expired to `completed`.** `sweepResolvedFailureTasks` flipped it after 14 days with `resolution: 'auto-expired'`, silently reporting work nobody decided to drop as done — and the task record federates, so the false `completed` propagated to every peer.
- **It lost its resume pointer.** `updateTask` treated the block as terminal-and-spent and stripped `existingBranch` / `resumeWorktreePath` / `resumedFromAgentId`, so a task revived after the user added a CLI provider restarted on a fresh branch and orphaned the worktree its dead agent left behind.

It is the same shape as `app-unresolved` / `workspace-invalid` — a config pause only the user clears — so it now joins both non-timed sets. It stays **out** of `TIMED_COOLDOWN_BLOCKED_CATEGORIES`: nothing stamps a `cooldownUntil` for it and nothing revives it, so the orphaned-PR notifier and the voice trigger keep surfacing it to the human.

The category is now the exported `PROVIDER_CONFIG_BLOCKED_CATEGORY`, used at the `agentLifecycle.js` stamp site, so the stamp can't drift from the sets that key on it.

## Test plan

- New: `cosTaskStore.test.js` keeps `existingBranch` / `resumeWorktreePath` / `resumedFromAgentId` through a `provider-config` block, and `sweepResolvedFailureTasks` leaves it `blocked` with no `resolution`/`autoExpiredReason` after 90 days.
- Extended: the reaper's user-intent exemption list, and the `agentLifecycle.js` source scrape (now pins the shared constant and rejects a re-hardcoded literal).
- **Both verified to fail without the change** — removing the two memberships fails 4 assertions; reverting the stamp site to the literal fails the source scrape.
- Fixed a test that used `provider-config` purely as a stand-in for "some reapable category" in the timestamp-fallback case; it now uses `max-spawns`, which is still reapable.
- Full server suite green (`cd server && npm test`). The 2–4 route/imageGen timeouts seen locally reproduce identically on a clean `main` checkout — pre-existing load flakes, not from this change.

Closes #6193


atomantic:

atomantic:
Review status detail: the configured reviewers for this run were `codex,claude`.

- **codex — BLOCKED (no verdict).** `codex exec` returned `You've hit your usage limit` before producing a review. Per the required-review rule that is `review-blocked`, not a clean pass, so no self-review was substituted.
- **claude — clean.** Reported "No findings", having independently confirmed that all four consumers of the two sets (`investigationTasks.js:355`, `cosTaskStore.js:778` and `:1059`, `orphanedPrNotifier.js`) behave as intended and that no stray `'provider-config'` literal remains at a stamp site.

Merge once codex (or another required reviewer) can return a verdict. Claim markers and the branch are preserved.

Issue #6200, opened 2026-09-04
a context-window reload clears an active local-model tuning's launch env


atomantic:
## Problem

`ensureContextWindow` (`server/services/ollamaManager.js`) restarts the daemon with an env containing **only** `OLLAMA_CONTEXT_LENGTH`:

```js
const env = withOllamaContextEnv({}, target)
…
return { ...(await restartWithEnv(env, { tuning: false })), … }
```

`applyLaunchEnv` then deliberately unsets every launchd key the *previous* env exported that the new one does not name ("Omitting a variable is not clearing it"). So a context-window reload wipes an active local-model tuning's launch knobs — `OLLAMA_FLASH_ATTENTION`, `OLLAMA_KV_CACHE_TYPE` (`launchEnv` in `server/lib/localModelTuning.js`) — and the daemon comes back untuned with no warning. The tuning's own bookkeeping is cleared too (`preTuningEnv = null` on a non-tuning restart), so there is nothing left to put back.

## Why it matters now

This is pre-existing, but #6191 pinned `numCtx: 131072` on `claude-ollama` / `claude-ollama-tui`, so the reload path now fires for two more shipped records — i.e. the first Claude-Ollama agent spawn after a tuning sweep silently drops that sweep's knobs.

## Work

- Compose the context-window env **on top of** the currently applied launch env instead of replacing it, so a reload for a window preserves knobs that are not about the window. `restartWithEnv`'s `{ tuning: false }` flag already distinguishes "the install's real configuration changing" from a temporary tuning — the composition rule should follow that distinction rather than always starting from `{}`.
- Decide and document what happens when a tuning is *active* and a window reload arrives mid-sweep: preserving the knobs keeps the sweep's measurements comparable, but the window itself changes under it. Preserving looks right; make it explicit either way.
- Cover it in `server/services/ollamaManager.test.js`: apply a tuning, then call `ensureContextWindow` at a larger window, and assert the restart env still carries the tuning keys plus the new `OLLAMA_CONTEXT_LENGTH`.

## Acceptance criteria

- A context-window reload no longer unsets `OLLAMA_FLASH_ATTENTION` / `OLLAMA_KV_CACHE_TYPE` that a tuning had applied.
- The tuning's undo bookkeeping survives the reload (or is explicitly, deliberately cleared with a comment saying why).
- A regression test fails against today's behavior.

Related to #6191

PR #6201, opened 2026-09-04
pin a 128K numCtx on the Ollama-backed Claude harnesses so a local review prefill isn't run at a 262K window

### LINKED ISSUE #6191: pin numCtx on the claude-ollama providers so a local review prefill isn't run at a 262K window
## Problem

`claude-ollama` and `claude-ollama-tui` in `data.reference/providers.json` carry **no `numCtx`**, so `ensureOllamaAgentContext` (`server/services/ollamaAgentContext.js`) never reloads the daemon and Ollama's auto-pick stands. On this machine that resolved to `llama-server … -c 262144 --cache-type-k q8_0 --cache-type-v q8_0 -b 1024 -ub 1024`.

The cost shows up as prefill throughput that *decays as the KV grows*: a 97,759-token pr-reviewer Stage 3 prompt was measured at 88 tok/s early and 48 tok/s by the end — roughly 33 minutes of prefill before the model can emit a token. PR #6188 raised `API_TIMEOUT_MS` so that no longer looks like a wedged run, but the underlying half-hour is still real, and the pipeline's own duration estimate (`localPromptBudget.js`, #6117) assumes a conservative 120 tok/s that this configuration does not reach.

The `ollama` api-type provider already carries `numCtx: 32768`; only the two agent-harness wrappers were left unset.

## Decision (already made — this is ready to work)

Set `numCtx: 131072` on `claude-ollama` and `claude-ollama-tui`. Rationale: it must exceed the ~100K-token Stage 3 envelope with headroom, and halving the window halves the KV-cache bandwidth that causes the decay. Do **not** pick 32768 — that is smaller than the envelope and would turn a slow run into a refused one.

## Work

- `data.reference/providers.json` — add `numCtx: 131072` to both records.
- `scripts/migrations/NNN-*.js` — set `numCtx` on existing installs' copies of those two providers when the field is still absent (leave a user-chosen value alone). Seed + migration, per the shipped-provider convention in `AGENTS.md`.
- Confirm `ensureOllamaAgentContext` then reloads the daemon at that window before the harness starts (it warns rather than failing if the reload is refused, which is the intended behavior).

## Verification

Re-run a Stage 3 dispatch on a local Claude wrapper and compare `prompt processing … tokens per second` in the Ollama log against the 88→48 tok/s baseline recorded above.

atomantic:
## Summary

- `claude-ollama` / `claude-ollama-tui` shipped with no `numCtx`, so `ensureOllamaAgentContext` never reloaded the daemon and Ollama's VRAM-based auto-pick stood — a 262144-token window on a large-VRAM machine. Prefill throughput decays as the KV cache grows: a ~98K-token pr-reviewer Stage 3 prompt measured 88 tok/s early and 48 tok/s by the end, roughly half an hour before the first token.
- Pins `numCtx: 131072` on both records in `data.reference/providers.json`. It clears the ~100K Stage 3 envelope with headroom while halving the KV bandwidth driving the decay. 32768 (the `ollama` api-type default) is *smaller* than the envelope and would turn a slow run into a refused one.
- Adds migration `339-claude-ollama-num-ctx.js` for existing installs, keyed on the `ollamaBacked` marker plus the `claude` command basename so a renamed clone is covered and a record repointed at another binary is not. A stored `numCtx: null` counts as unset — `createProvider` normalizes every absent value to `null` on write, so treating it as user intent would make the migration a no-op on essentially every real install.
- The migration stands down entirely when `OLLAMA_CONTEXT_LENGTH` is set: `resolveOllamaContextLength` ranks a record's `numCtx` *above* that env var, so stamping would silently override a deliberately small machine-wide window — and past what VRAM allows Ollama offloads to CPU rather than failing loudly.
- Guard test in `ollamaContext.test.js` asserts both shipped records resolve to a window above the Stage 3 envelope with no ambient env, so a future seed edit can't silently drop the pin.

## Review status

**Codex review is blocked** — the CLI returned `You've hit your usage limit` and produced no verdict. The `claude` reviewer ran two rounds: round 1 raised the `OLLAMA_CONTEXT_LENGTH` override (fixed in the second commit) and round 2 came back clean.

## Test plan

- `scripts/migrations/339-claude-ollama-num-ctx.test.js` — 10 cases: fresh install, absent vs. normalized-`null` numCtx, clones and `.EXE` spellings, non-Claude Ollama providers left alone, user-chosen windows in both directions, ambient env stand-down, junk env value, re-run no-op, unreadable file.
- Bypass-probed the seed guard: stripping `numCtx` from either seed record fails `ollamaContext.test.js`.
- Full `server` suite green — 38,937 passed, 29 skipped.

## Follow-up

Filed #6200 for a pre-existing behavior this pin now exercises: `ensureContextWindow` restarts the daemon with an env containing only `OLLAMA_CONTEXT_LENGTH`, and `applyLaunchEnv` unsets stale launchd keys — so a window reload clears an active local-model tuning's `OLLAMA_FLASH_ATTENTION` / `OLLAMA_KV_CACHE_TYPE`.

Closes #6191


atomantic:

PR #6202, opened 2026-09-04
refuse test writes into the install's real data/ tree at runtime

### LINKED ISSUE #6176: Guard against tests WRITING into the real data/ tree (the half the isolation probe can't see)
## Problem

`server/lib/testDataIsolation.guards.test.js` is the repo-wide guard against a test addressing the checkout's real `data/` tree. Its own header documents the hole it leaves:

> It does NOT cover the third spelling — importing the real `PATHS` and reading a data-rooted member (`PATHS.missions`, `PATHS.brain`, …) with no redirect. […] That spelling is covered **empirically** instead, by the two-run probe #3687 introduced.

That probe is a **read**-leak detector: plant records, re-run, and a test that now fails is leaking. It cannot see a test that **writes** into the real `data/` tree, because a write does not change whether assertions pass. That half is currently undetected, and it is the more damaging half — the read leak announces itself as a failing test, while the write leak is silent.

It is not hypothetical. PR #6171 fixed exactly this in `server/services/providerUsage.test.js`: the suite mocked the provider registry and the TUI scrape but not the federation collaborators, so `getProviderQuotas` reached the real store and `recordLocalQuotaCards` persisted the suite's fixture cards over the machine's genuine readings in `data/provider-quotas.json` — then invalidated the `usage` sync checksum, so subscribed peers pulled the fabricated values. It surfaced only because the *read* side also happened to break assertions on a populated install; had the fixtures merely been written, nothing would have reported it.

## Decision (chosen approach — not an open question)

**Mirror the database protection at the filesystem layer: an always-on write backstop in the shared write primitive.**

This is the same shape as the existing, proven guard in `server/lib/db.js` — `isTestRunner()` (`db.js:82`) plus the `query()` backstop (`db.js:288`) that refuses row writes to a non-test database. That guard exists because a suite once wiped real records; this is the file-tree analogue of the same failure.

Rejected alternative, for the record: extending the static text-scan rule in `testDataIsolation.guards.test.js` to the `PATHS.*` spelling. Its header already explains why — ~50 of ~1300 scoped test files use that spelling, nearly all safely, so a static rule needs a 50-entry allowlist that rots. A runtime backstop needs no allowlist: it fires on the write that actually reaches the real root, and stays silent for the ~50 that never get there.

### Implementation

1. In `atomicWrite` (`server/lib/fileCore.js:161`), when `isTestRunner()` is true, resolve the target to an absolute real path and throw if it falls under the real `PATHS.data` root (`server/lib/paths.js:30`).
   - Resolve before comparing (`path.resolve`, and realpath the root) so a relative path, a `..` climb, or a symlink cannot walk in sideways.
   - The error must name the offending path and point at the sanctioned escape hatch, so the fix is obvious from the failure text alone.
2. **Do not add the guard to `server/lib/aiToolkit/internal/atomicWrite.js`.** That tree is vendored and self-contained — it must not import out to other PortOS modules (see `server/lib/aiToolkit/AGENTS.md`). Leave it alone.
3. Audit for writers that bypass `atomicWrite` and reach `data/` directly (raw `writeFile` / `rm` / `mkdir` in services and stores). Route them through the guarded primitive, or give them the same check. This audit is the bulk of the work and the main source of risk.
4. Run the full server suite. Every test that trips the new guard is either a real leak (fix it) or a suite that legitimately needs a data root (convert it to `createTempDataRoot()` + `makePathsProxy()` from `server/lib/mockPathsDataRoot.js`, which resolve outside the real root and therefore pass untouched). Triage each one individually — do not blanket-allowlist.

## Acceptance criteria

- [ ] A test that writes anywhere under the real `data/` root fails, with an error naming the path and the `createTempDataRoot()` escape hatch.
- [ ] A guard test proves it: a write to a temp data root succeeds, a write to the real root throws. Assert the thrown error, not merely that *something* threw.
- [ ] The guard is inert outside the test runner — production writes are unaffected.
- [ ] `server/lib/aiToolkit/` is untouched.
- [ ] Full server and client suites pass, with any leaker found in step 4 fixed rather than allowlisted.
- [ ] `server/lib/testDataIsolation.guards.test.js`'s header is updated to say the write half is now covered at runtime, so the next reader isn't told the gap still exists.

## Notes

Worth running the two-run probe from that header once after the guard lands — the guard covers writes, the probe covers reads, and together they close both halves of the bug class.

atomantic:
## Summary

A test that writes into the developer's live `data/` tree is invisible: unlike a read leak, persisting fixture bytes over their records changes nothing about whether assertions pass. #6171 was the proof — `providerUsage.test.js` wrote its fixture quota cards over the machine's real `data/provider-quotas.json`, then invalidated the sync checksum so peers pulled the fabricated values.

This closes that half at runtime, the way `db.js` already refuses row writes to a non-test database.

- **`server/lib/testDataIsolation.js`** throws when a write resolves inside the real data root under the test runner, naming the offending path and the `createTempDataRoot()` escape hatch. It fires from `atomicWrite`, `ensureDir`'s create path (an existing directory is a no-op, so read-only suites stay quiet), new `writeFile`/`appendFile`/`copyFile` wrappers in `fileCore.js`, and `collectionStore`'s record delete — so the next raw writer inherits the guard rather than having to remember an import. `aiToolkit/` is untouched, per its self-containment rule.
- **Inert in production**, and cheaper than that: the guard module loads through a memoized `await import()` behind `isTestRunner()`, so a production process never loads it at all. The tree-wide static-import budget in `importScoping.test.js` is unchanged — no ceiling was raised.
- **Symlinked destinations are judged at both landing sites.** `atomicWrite` replaces a symlink; `writeFile`/`appendFile`/`copyFile` follow one. Checking only the link's own location would have left the guard's own hole open.

### Leaks it found on its first run

| Suite | What it was writing into the live install |
|---|---|
| `backup.test.js` | `vi.doUnmock` also drops the file-level `vi.mock`, so every later suite resolved `PATHS.data` to the real tree and `runBackup` rewrote the user's genuine `data/backup/state.json` on every run |
| `providers.readiness.test.js`, `loraTraining/captionLeakStaging.test.js` | took the **machine-wide** heavy-local-job claim — the file that gates the user's real local renders |
| `sharing/integration.test.js` | a hand-rolled PATHS proxy listed `data`/`images`/`videos` by hand, so the later-added `imageRefs` still pointed at `data/image-refs` and the importer copied bundled reference sheets there |
| `assetHash.test.js`, `modelAbuseGuard.materialize.test.js` | wrote fixtures into the real tree deliberately, swept up only when nothing above them threw |

Each is fixed by redirecting the data root, not by silencing the guard.

### Supporting changes

- `isTestRunner` moves out of `db.js` into a dependency-free `lib/runtimeEnv.js`. The file primitives must not pull in `pg`, and the many suites spelling `vi.mock('../lib/db.js', () => ({ query }))` were stripping it out of the module graph for every other consumer.
- `lib/pathContainment.js` owns root-inclusive containment and ancestor-canonicalization, kept a leaf so `fileCore` reaches it without dragging `errorHandler`/`paths` into every suite's closure.
- `mockPathsDataRoot.js` gains `lazyTempDataRoot` / `cleanupTempDataRoots`, so the `vi.mock`-hoisting hazard is explained once instead of in each suite that redirects a root.
- `agentRunTracking.js`, `autonomousJobs/store.js`, `cosState.js`, `sharing/importer.js` and `runner.js` move from raw `fs` writes onto the guarded primitives.

### Scope, stated rather than implied

Writes only — reads stay covered by the two-run probe. Roughly forty services still reach `PATHS.*` with raw `fs`, and on a populated install their directories already exist, so `ensureDir`'s create-path check is a no-op for them. Both limits are written into the module header rather than glossed over; the sweep is tracked separately.

## Test plan

- New `server/lib/testDataIsolation.test.js` (22 cases) asserts the thrown **message**, not merely that something threw: a temp root writes fine, the real root refuses with the path and the escape hatch named, the guard is inert outside the runner, and a `data-archive` sibling, a `..` climb, a relative path, a symlinked destination and a filesystem root all resolve correctly.
- Full server suite: 38,945 passing. Full client suite re-run. The handful of remaining failures are timeouts under load that reproduce identically on `origin/main` — verified against a pristine baseline worktree.
- `importScoping.test.js`, `index.test.js` (barrel/README catalogs) and the existing static `testDataIsolation.guards.test.js` all pass.

Closes #6176


atomantic:

atomantic:
### Local review status

Configured reviewers for this run:`claude`.

- **claude — completed.** Returned three findings, all confirmed against the code and fixed in f612163:
  1. `writeFile`/`appendFile`/`copyFile` follow a symlinked destination, but the guard judged only the link's own location — a fixture symlink in a temp root could write straight through into live data. Both landing sites are now checked, resolved via `readlink` + `canonicalizePath` so a *dangling* link still reports its target (`realpathSync` throws on those).
  2. `isPathAtOrInsideDir` reported nothing as inside a filesystem/drive root — those already end in the separator, so the anchor compared against `//`.
  3. The `..`-climb case in the guard's own test was vacuous: `path.join` normalizes its arguments, so the assertion held even with `resolve()` removed.

The claude reviewer's first invocation was killed by system memory pressure with no output; it was retried once and completed normally.

PR #6204, opened 2026-09-04
test(sprites): hoist lockAllAnchors fixture onto a real-locked prototype

### LINKED ISSUE #6180: CI: cut fixture-construction file I/O in sprites/walk and sprites/atlas (successor to #6004)
## Why

Successor to #6004, which is closed: its premise (full-size Sharp fixture renders dominate these two suites) is disproven, and the real Sharp win it could deliver shipped in #6165 (56s → 21.2s and 53s → 23.6s on Windows CI, ~2.5×). Three independent passes now agree Sharp is not the remaining cost:

- #6165 measured `sharp().stats()` at ~4ms regardless of frame size — libvips pipeline setup, not pixels.
- A contributor pass on #6004 benchmarked a test-only atlas geometry at ~34% of the production pixel area: 24.40s vs a 24.30s baseline, i.e. no measurable change.
- Profiling `sprites/walk.test.js` on a Windows 11 dev machine (2026-09-04) puts the Sharp calls inside the anchor-lock pipeline (`analyzeForeground` 726ms + `normalizeFromAnalysis`/`recompositeOnKey` 593ms across 351 locks) at ~24% of lock time and ~14% of the suite.

What actually dominates is **fixture construction**: many small file operations, cheap on Linux and expensive on a Defender-scanned Windows runner. Measured shares (Vitest `tests` phase; shares are the meaningful figure — the host was under full CPU load, so treat the absolutes as relative-only):

| suite | fixture cost | share |
|---|---|---|
| `sprites/walk.test.js` | `lockAllAnchors` (117 calls, 3 anchors each) | 5.5s of 9.3s — **59%** |
| `sprites/atlas.test.js` | `lockAllAnchors` (19 calls, 9 anchors each) | **26%** |
| `sprites/atlas.test.js` | `buildFinalizedWalkSet` (18 calls, ~150 files each) | **32%** |
| `sprites/atlas.test.js` | `compileAtlas` (32 calls) | 38% — **the production work under test; out of scope** |

Every one of those 117 walk characters and 19 atlas characters is built from byte-identical cached candidate PNGs, so the lock pipeline re-derives the same tree over and over.

## What to do

Build the fixture once per `describe` in `beforeAll` and materialize each test's character from it, instead of re-running the real pipeline per test.

- **`sprites/walk.test.js`** — lock one prototype character's anchors in `beforeAll`. `characterWithLockedAnchors` then creates the record row (2ms) and recursive-copies the prototype's sprite directory to the new id, renaming the id-stamped reference files and rewriting the paths in `reference-manifest.json`. Keep one describe-level variant per distinct `directions` argument the suite uses.
- **`sprites/atlas.test.js`** — same for `lockAllAnchors`, and hoist `buildFinalizedWalkSet` the same way so `finalizedCharacter()` copies a finalized prototype rather than re-writing ~150 files per test.
- **Pin the derived state to the real pipeline.** This is the correctness risk #6165 deliberately kept out of scope: if a copied tree drifts from what a real lock produces, the walk/atlas assertions go green on a fake precondition. Add one equivalence test per suite that builds a character *both* ways and asserts the two trees are byte-identical after id normalization (file names, manifest paths, and every embedded `sha256`). That test is the acceptance gate for the refactor, not an extra.
- **Do not touch `compileAtlas`**, `validateForCompile`'s hash chain, or the real chroma-key selection inside `lockReference`. Those are the tamper-evidence and pixel behaviors these suites exist to verify, and trimming them would weaken the suite rather than speed up its fixtures.

## Expected payoff — and why the bar moved

A recursive copy of a locked sprite tree measured **18.3ms** against **~47ms** for a 3-anchor `lockAllAnchors` on the same filesystem, so the copy recovers roughly 60% of fixture cost, not all of it. Windows small-file I/O is the floor either way. Projected: `walk` 23.6s → ~16s, `atlas` 21.2s → ~16s, ~14s off the Windows shard-3 wall clock.

That is why the acceptance below is **17s, not the 15s** #6004 and #5902 carried. The 15s figure was set before anyone had profiled these files, and #5902's other Windows slice (#6003) closed with its two files at 67s and 40s against the same bar. 17s is what the measured floor supports without weakening coverage.

## Acceptance criteria

- `sprites/walk.test.js` and `sprites/atlas.test.js` each run under 17s on Windows CI (currently 23.6s / 21.2s).
- Each suite has an equivalence test proving the copied fixture is byte-identical to one built through the real lock/finalize pipeline, and that test fails when the prototype and the pipeline drift.
- No test skipped or weakened; the full sprite suite passes on Linux and Windows.
- Record the before/after `⏱` line from `scripts/run-ci-tests.js`'s job summary in the PR.

## Files/areas

- `server/services/sprites/walk.test.js`
- `server/services/sprites/atlas.test.js`
- `server/services/sprites/spriteTestFixtures.js`

Part of #5902

Bryandero98:
## Summary

- Addresses #6180's `lockAllAnchors` half: 136 call sites across `walk.test.js`/`atlas.test.js`/`animationTrackWorkflow.test.js` locked a character from byte-identical cached candidate PNGs, each re-running the real Sharp normalize + chroma-key-selection pipeline from scratch.
- `lockAllAnchors` now locks ONE real prototype per `TEST_ROOT` (at the full anchor set, since every caller's `directions` is a subset of it) and materializes every other character by recursive-copying only the artifacts a caller actually asked for, id-normalizing the manifest, and resetting any anchor outside the requested set to the exact seeded-pending shape a fresh partial lock would leave.
- No call site changed except the three that create the record + now pass `records` through — the fast path lives entirely inside `spriteTestFixtures.js`.
- Each suite gets an equivalence test (the acceptance gate #6180 asks for, not an extra): builds a character both through the real pipeline (`lockAllAnchorsReal`, exported unchanged) and through the materialized path, and asserts byte-identical results after id normalization — file names, PNG bytes, the full manifest (every embedded `sha256`), and the record's `chromaKey`/`status`.

## Two options for this issue

This is **option A of two** — see #6207 for option B, which adds a `buildFinalizedWalkSet` optimization on top of this same `lockAllAnchors` change. I found that a copy-based materialization of `buildFinalizedWalkSet` (mirroring this PR's approach) doesn't actually help — that function never calls Sharp (`writeWalkFramePng` already caches the encoded buffer), so copying files isn't cheaper than the cached-buffer writes it already does. #6207 instead parallelizes that function's 8 independent per-direction writes, which is a real win with much less mechanism. **I'd merge #6207 over this one** — same `lockAllAnchors` change, plus a real additional win, at lower risk (no new copy/rewrite logic, so no new class of drift to worry about). This PR stands alone as the smaller, more conservative option if you'd rather not take the `buildFinalizedWalkSet` change.

Only #6207 carries "Closes #6180" — merging either is fine, but I didn't want both auto-closing the issue out from under the other.

## Test plan

- [x] Full `server/services/sprites/` suite: 716 passed, 1 skipped, 0 failed (same pass count as `main`)
- [x] Both suites' new equivalence tests pass, and caught two real bugs during development (south-anchor sync, partial-lock manifest status) before landing
- [x] Measured locally (Windows): `walk.test.js` ~29.4s → ~19.7-21.4s tests phase (~30% faster); `atlas.test.js` ~24.4s → ~20.2-21.3s tests phase (~15% faster)
- [ ] Windows CI `⏱` before/after (will follow up with the actual CI numbers once this runs)


Bryandero98:
Rebased onto `main` after #6205 merged (span narrowing + `capSharpThreads`) — both touch the same 3 files. No functional changes here beyond the merge; full sprites suite re-verified post-merge (26/26 files, 716/717 tests, 1 pre-existing skip).

Issue #6004, opened 2026-09-03
CI: shrink Sharp fixture cost in sprites/atlas and sprites/walk suites


atomantic:
## Why

`server/services/sprites/atlas.test.js` (56s Windows, 38s Linux, 21 tests) and `server/services/sprites/walk.test.js` (53s Windows, 16s Linux, 116 tests) spend most of their time rendering full-size Sharp atlas/strip fixtures. The assertions in both suites are about grid/offset/order layout math, not pixel content, so full-resolution renders buy nothing.

## What to do

- Shrink the fixture frame dimensions to the smallest size that still exercises the layout math being asserted (grid packing, offsets, frame ordering).
- Share encoded fixture buffers across tests within a suite via `beforeAll` instead of re-encoding per test, where tests don't mutate the fixture.
- Re-verify: assertions must stay meaningful at the smaller size — don't shrink past the point where edge/boundary layout cases stop being exercised.

## Acceptance criteria

- `sprites/atlas.test.js` and `sprites/walk.test.js` each run under 15s on Windows CI (currently 56s / 53s).
- No test skipped or weakened; all existing assertions still pass on both Linux and Windows.
- Record the before/after `⏱` line from `scripts/run-ci-tests.js`'s job summary in the PR.

## Files/areas

- `server/services/sprites/atlas.test.js`
- `server/services/sprites/walk.test.js`
- Any shared sprite-fixture helper these suites use

Part of #5902

---

When you ship this slice, check its box in #5902 (edit the epic body checklist), and if this was the LAST open child, close #5902 with a summarizing comment naming all the slices that delivered it.

atomantic:
Partially delivered by #6165 (merged).

**Done ✓**

- Diagnosed the actual bottleneck. The premise in this issue — that full-size Sharp fixture renders dominate — is **wrong**. `sharp().stats()` in the degenerate-frame probe costs ~4ms whether the frame is 16×16 or 256×256, because the cost is libvips pipeline setup rather than pixels (16×16: 201ms / 40×40: 215ms / 96×96: 255ms / 256×256: 308ms for 48 probes). Shrinking fixtures would have bought almost nothing. Instrumenting one `compileAtlas` put 442ms of its 532ms in `validateForCompile`, all of it re-probing frame bytes it had already probed.
[...]

Issue #2986, opened 2026-07-24
Drop the dead scanner column from the compiled walk atlas — it duplicates idle and no consumer reads it


atomantic:
## Problem / Goal

Every compiled walk atlas carries a trailing **`scanner` column that is a verbatim copy of the idle cell** — and nothing consumes it. It is a legacy artifact of an early render where the character held up a hand-held device, and it was baked into the walk grid rather than treated as what it is: a *separate action animation*.

Remove it from the compiled grid. A scanner action, when it exists, is its own animation track — not a column appended to the walk cycle.

### PortOS emits it as a duplicate of idle

`server/services/sprites/atlas.js:60`:

```js
export const atlasColumns = (walkLabels) => ['idle', ...walkLabels, 'scanner'];
```

The cell itself is not rendered from anything — it shares the idle cell's buffer outright (`atlas.js:345-353`), stamped `scannerPolicy: 'locked-idle-placeholder'` and surfaced in the atlas manifest at `atlas.js:550`. So every atlas pays one full column of width to ship a second copy of column 0.

### The consuming game never reads it

In the managed app, `ScannerAnimationColumn` is **declared and never read** — its only use is computing `AnimationColumnCount`, which is itself never read anywhere in the codebase. The column the game actually draws comes from:

```csharp
public int AnimationColumn => _walkCycle.IsMoving
    ? FirstWalkAnimationColumn + _walkCycle.Frame
    : IdleAnimationColumn;
```

…which can only ever return idle or a walk column. The scanner in that game is a **HUD tool icon and a cargo item**, not a character animation — there is no scanner pose, no scanner state, and no code path that would draw one.

So the column is dead weight on both sides: PortOS spends a column duplicating idle, and the game reserves an index it never samples.

### What removing it costs

The grid drops from `idle + N + scanner` to `idle + N`:

| walk frames | columns today | columns after | atlas width @ 96px |
|---|---|---|---|
| 8 | 10 | 9 | 960 → 864 |
| 12 | 14 | 13 | 1344 → 1248 |

Nothing in PortOS depends on the column existing — `atlasColumns()` is the single source of the grid, and the manifest's `geometry.columns` is derived from it. Reading *imported* atlases and manifests that still carry a scanner column is unaffected; this changes only what PortOS compiles.

### Sequencing matters

[...]

PR #6205, opened 2026-09-04
Cap libvips threads and narrow compiled walk spans in the sprite test suites

atomantic:
> **Status: needs a maintainer call before it goes anywhere.** #6004 — the issue this branch was claimed against — was closed as a dead end at 03:46Z while this work was in flight, and its remaining scope moved to #6180. Read the "Relationship to #6004 and #6180" section below before reviewing the diff. Merge is withheld regardless (see the review comments).

## Summary

Two independent reductions in what the sprite `atlas`/`walk` Vitest suites cost on a **contended** runner. No test is skipped, deleted, or weakened.

**1. Cap libvips threads (3 files, ~15 lines).** Both suites call a new `capSharpThreads()` fixture helper (`sharp.concurrency(1)`) and restore the previous value in `afterAll`. Every image these suites touch is 40–192px, where the libvips pool's setup/join is most of the call — which is the same finding #6165 recorded as "`sharp().stats()` costs ~4ms regardless of frame size — libvips pipeline setup, not pixels". CI already runs several Vitest workers per box, so the pool only oversubscribes it on top of that. The restore matters: `sharp.concurrency` is process state and Vitest reuses worker processes across files, so an uncapped setting would stick to every later file in that worker.

**2. Narrow the compiled walk span where the assertion is frame-count-parametric.** Scanner/jetpack span placement, hip-anchor registration, speck grounding, the pre-#2986 grid-drift recompile, idempotency, geometry overrides, self-heal versioning and the refusal paths now build the narrowest walk span the walk track's registry allows (`WALK_MIN_FRAME_COUNT`) instead of 8 or 12. Their expectations derive span starts and widths from that constant rather than hardcoding column indices, so they pin exactly the same arithmetic. The production 9×8 grid stays pinned, unchanged, by `compiles the 9×8 player atlas`, and the 12-frame variable-width case keeps its own dedicated test. This cuts both the cells each `compileAtlas` renders and the files each `buildFinalizedWalkSet` writes.

## What this branch independently confirmed, and what it rules out

Before writing either change, this run measured the two levers #6004 actually proposed. Both are no-ops, matching @Bryandero98's pass and the maintainer's Windows profiling:

- **Fixture frame size** — halving the 40×40 walk fixture to 24×24 left both suites' CPU time unchanged (and forced an assertion tolerance to be loosened, so it was reverted rather than shipped).
- **Atlas cell size** — halving `DEFAULT_ATLAS_GEOMETRY.cellSize` from 96 to 48 as a probe: 36.77s user CPU vs a 36.92s baseline. No change.

A compiled cell costs two Sharp calls (decode + premultiplied resize) whose price at these sizes is per-call overhead, not pixel count. What is left is the **number** of calls and the thread pool behind each one — which is what this branch attacks.

## Measurements

On an **idle** machine there is no change, because these suites are not CPU-bound when nothing else is running: `atlas` 4.68s → 4.63s, `walk` 3.55s → 3.81s. That is the honest result, and it is consistent with the 56s/53s Windows figures being a contention/small-file-I/O signal rather than an image-work one.

Under contention it is real. Two alternating A/B rounds of both suites on a deliberately loaded machine (load average 60–90), measuring CPU rather than wall clock:

| Round | Baseline user+sys | This branch | Δ |
|---|---|---|---|
| 1 | 48.7s + 34.2s = 82.9s | 36.4s + 20.7s = 57.1s | −31% |
| 2 | 47.1s + 30.6s = 77.7s | 32.5s + 17.3s = 49.8s | −36% |

System time roughly halves, which is the thread-pool oversubscription going away. Measured separately under the same load, `capSharpThreads()` alone accounts for most of the `walk` gain (12.25s → 8.40s user) and only a small part of the `atlas` gain (36.9s → 34.6s user); the span narrowing is what moves `atlas`.

This run stops at publication and does not watch CI, so it does not post the `⏱` job-summary line itself — read it off this PR's full-suite job and compare against `main`.

## Relationship to #6004 and #6180

- **#6004 is closed** (COMPLETED, resolved by #6165, remaining scope → #6180). Nothing here reopens it; the trailer is `Refs`, not `Closes`.
- **Change 1 (`capSharpThreads`) does not overlap #6180 at all.** It is three lines in `spriteTestFixtures.js` plus one call site per suite, touches no assertion, and is orthogonal to #6180's fixture hoist. If only one half of this PR is wanted, this is the half.
- **Change 2 (span narrowing) overlaps #6180's territory** — same two files, and #6180 explicitly scopes `compileAtlas` out as "the production work under test". It does not do what #6180 does (it builds *less*, rather than avoiding a rebuild), and the two compose — a hoisted prototype is cheaper to copy when it is narrower — but it will cost @Bryandero98 a rebase on a ~3,400-line test refactor if it lands first.

**Recommended call:** either take the whole branch now, before #6180's refactor starts in earnest, or ask for it to be cut down to change 1 and leave the span narrowing to be folded into #6180. Closing this outright is also a reasonable answer; the negative results above stand on their own either way and are the reason the fixture-shrink path is not being tried a fifth time.

## Test plan

- `cd server && npx vitest run services/sprites/` — 26 files, 713 passed, 1 skipped (pre-existing). Run on both a quiet and a heavily loaded machine, before and after rebasing onto current `main`.
- No test deleted, skipped, or `SKIP_HEAVY_INTEGRATION`-gated by this change; the assertion count is unchanged.

Refs #6004, #6180

### DIFF
--- a/server/services/sprites/atlas.test.js
+++ b/server/services/sprites/atlas.test.js
@@ -26,15 +26,9 @@ import {
   writeWalkFramePng,
 } from './spriteTestFixtures.js';
 
-capSharpThreads();
+const restoreSharpThreads = capSharpThreads();
+afterAll(restoreSharpThreads);
 const TEST_ROOT = mkdtempSync(join(tmpdir(), 'sprite-atlas-test-'));
-
-// The narrowest walk span a compile will accept: the walk track's registry row
-// declares a 6–16 frame range and `resolveTrackUniformity` refuses anything
-// outside it, so 6 is the floor — still enough for varyArm to vary the
-// silhouette and for a mid-strip index to exist. Tests that pin the production
-// 9-column grid keep WALK_PHASES instead (#6004).
-const NARROW_WALK_FRAMES = 6;
 
 vi.mock('../../lib/fileUtils.js', async (importOriginal) => {
   const actual = await importOriginal();
@@ -61,7 +55,9 @@ const records = await import('./records.js');
 const { lockReference, loadManifest } = await import('./reference.js');
 const { compileAtlas, getAtlasState, ATLAS_COLUMNS, DEFAULT_ATLAS_GEOMETRY } = await import('./atlas.js');
 const { SPRITE_DIRECTIONS } = await import('./prompts.js');
-const { WALK_PHASES, walkPhaseLabels, WALK_FPS } = await import('./walkPostprocess.js');
+const {
+  WALK_PHASES, walkPhaseLabels, WALK_FPS, WALK_MIN_FRAME_COUNT,
+} = await import('./walkPostprocess.js');
 const { buildAtlasGrid, compiledGridUpToDate } = await import('./atlasGrid.js');
 const { getAnimationTrack, SCANNER_TRACK, AMBIENT_TRACK } = await import('./animationTracks.js');
 // #3152 — `scanner`/`ambient` are seeded STORE rows, so the compiler's real table
@@ -71,6 +67,14 @@ const {
   animationTrackStorePath, animationTrackSeedPath,
 } = await import('./animationTrackStore.js');
 const EFFECTIVE_TRACKS = getEffectiveAnimationTracks();
+
+// The narrowest walk span a compile will accept: `resolveTrackUniformity`
+// refuses a direction outside the walk track's authoring range, so this tracks
+// the registry's floor rather than restating it — raising the floor adapts these
+// tests instead of breaking a dozen of them. Still wide enough for varyArm to
+// vary the silhouette and for a mid-strip index to exist; tests that pin the
+// production 9-column grid keep WALK_PHASES instead (#6004).
+const NARROW_WALK_FRAMES = WALK_MIN_FRAME_COUNT;
 
 let seq = 0;
 const newId = () => `atlas-char-${++seq}`;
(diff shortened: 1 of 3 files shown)


atomantic:

atomantic:
Review audit trail for this run:

- `codex` (required) — **blocked**, no verdict. Failed on its first invocation with `ERROR: You've hit your usage limit` (account-level; resets 2026-09-07). Not substituted and not self-reviewed.
- `claude` (required) — completed. Round 1: FINDINGS (unrestored process-global `sharp.concurrency`; `NARROW_WALK_FRAMES` restating the walk registry's floor as a literal, plus a track repartition that only tiled the grid while that floor was ≥ 4). Both fixed in `ea8e5f4`. Round 2 (delta-scoped convergence check): CLEAN.

Merge is withheld pending `codex`. The claim markers and the `claim/issue-6004` branch are preserved.

atomantic:
Note on the "## Measurements" section: this autonomous run stops at publication (see the review-blocked comment above), so it does not watch CI or post the `⏱` job-summary numbers itself. Whoever picks this up should read the before/after `⏱` line off this PR's full-suite job summary and compare it against `main` — that number is what decides whether the "## Remaining" work is still needed.

Issue #6062, opened 2026-09-03
pr-reviewer Stage 3 should run as an attachable TUI session when its provider is a TUI record


atomantic:
## Problem

The pr-reviewer Stage 3 (PR Code Review & Actions) agent always runs **headless**, even when the configured provider is a TUI record. `agentLifecycle.js` forces it:

```js
const spawnHeadless = publicReview || !isTui;
```

That means there is no PTY to attach to. When Stage 3 goes sideways — wedged on a test run, looping on a patch that will not apply, burning minutes on a diff it cannot evaluate — the operator cannot drop in, read what it is doing, or steer it. The only options are wait for the run to time out or kill it.

The stage's *provider eligibility* is not the blocker: `supportsPublicReviewPosture` already accepts any `isDirectBinaryProvider` record, TUI records included, and `claudePublicReviewArgs` in `server/lib/providerVendors.js` already takes a `tui` flag that keeps the whole enforced recipe (`--restricted`/`--settings` sandbox JSON/`--disallowedTools`) and only drops the `--print --output-format stream-json` headless flags. API and custom providers stay ineligible by design (no maintained recipe = no enforcement) and must remain so.

## Scope

Make the `sandboxed-actions` posture (Stage 3) spawnable as an interactive TUI session when the configured provider is a TUI record. The `no-tool` posture (Stage 1 security scan, Stage 2 eligibility gate) must stay headless — an interactive session for a reasoner with no tools buys nothing and widens the boundary for free.

## Decisions already made (do not re-open)

- **Only `PUBLIC_REVIEW_ACTIONS_POSTURE` gets TUI.** `PUBLIC_REVIEW_NO_TOOL_POSTURE` stays headless.
- **The vendor recipe still governs the argv.** A TUI Stage 3 must be built from `publicReviewRecipe(provider, posture).spawnArgs(provider, { ...ctx, tui: true })`, never from the generic `provider.args` assembly. A provider with no maintained recipe stays ineligible and stays headless.
- **The restricted child env still applies.** Whatever `cliChildEnv.js` strips for `isPublicReviewRestrictedProfile` must be stripped on the PTY path too.
- **No new user-facing setting.** If the stage's configured provider is a TUI record, Stage 3 runs as a TUI; otherwise headless. Same rule the rest of the queue uses.

## Implementation sketch

[...]

PR #6206, opened 2026-09-04
spawn a restricted public-review TUI as a direct PTY instead of hosting it in a login shell

### LINKED ISSUE #6159: Spawn an attachable public-review stage as a direct PTY instead of hosting it in a login shell
## Problem

#6062 made the pr-reviewer `sandboxed-actions` stage (Stage 3) attachable when its provider is a TUI record with a vendor recipe marked `tui: true`, and threaded the restricted `buildCliChildEnv` allowlist into that PTY (`createAgentTuiSession` + the new `envReplace` option on `createShellSession`).

That allowlist closes the environment PortOS itself contributes — the server process env, provider `envVars`, the forge token. It cannot close one remaining source: the local PTY path hosts the provider CLI **inside an interactive login shell** (`shell.js#createShellSession` → `pty.spawn(shell, [])` with `initialCommand`), so the operator's own rc file (`.zshrc`, `.bash_profile`, …) runs *after* the allowlist and before the provider, and anything it exports is back in the child's environment.

The headless sibling has no such exposure: `spawnDirectly` spawns the binary itself, and so does the runner's `/spawn-tui` (`pty.spawn(ptyCommand, ptyArgs, { env: childEnv })` — no shell at all).

## Scope

For a restricted (`isPublicReviewRestrictedProfile`) TUI spawn only, launch the recipe command as the PTY directly rather than injecting it into a login shell. The runner's `/spawn-tui` handler is a working template for the shape, and `shell.js#registerExternalSession` already adopts an externally-spawned pty as an attachable session — which is exactly how the runner path registers its own.

Pieces to work out:

- A local `spawnCommandSession`-style entry point in `shell.js` (or a small sibling) that `pty.spawn`s a command with a caller-supplied complete env and registers it, mirroring `registerExternalSession`'s existing contract.
- The handshake affordances the login shell currently provides: `waitForPromptReady` (a round-trip probe against the shell) and `exitWithCommand` both disappear. The runner path already lives without them by firing `onInitialCommandSent?.()` immediately before the spawn — `createAgentTuiSession` does this today in its `useDurableRunner` branch, so the paste gate is known to work that way.
- Session exit becomes the CLI's own exit rather than the wrapper shell's, which is what `exitWithCommand` was emulating — likely a simplification, but the `handleExit` reason mapping in `spawnTuiAgent` should be checked against it.

Ordinary (non-restricted) agent TUI sessions must keep the login shell: the operator's rc file is a feature there, not a leak.

## Acceptance criteria

- A `sandboxed-actions` TUI spawn produces a PTY whose process is the provider binary, not a shell hosting it.
- An rc file that exports a variable outside the actions allowlist cannot make it into that child's environment (testable by asserting the spawn is not routed through a shell, plus the existing env assertions).
- Ordinary TUI agent spawns are unchanged — still a login shell, still `buildSafeEnv` + delta.
- The paste/prompt handshake still delivers, and the run still finalizes on its `.agent-done` sentinel.

## Files/areas

- `server/services/agentTuiSpawning.js` (`createAgentTuiSession`; the `KNOWN RESIDUAL` comment there marks the exact spot)
- `server/services/shell.js` (`createShellSession`, `registerExternalSession`, the `envReplace` option added by #6062)
- `server/cos-runner/index.js` `/spawn-tui` — reference implementation of the direct-PTY shape
- `server/services/agentTuiSpawning.test.js`

Related to #6062

atomantic:
## Summary

The pr-reviewer `sandboxed-actions` stage runs its provider CLI under a strict environment allowlist (#6062), but the local PTY path hosted that CLI **inside an interactive login shell** — so the operator's own rc file (`.zshrc`, `.bash_profile`, …) ran *after* the allowlist and before the provider, and anything it exported was back in the child. No allowlist can close that; removing the shell can.

- `shell.js` gains `spawnCommandSession`, which `pty.spawn`s a command directly with a caller-supplied **complete** environment (nothing unioned underneath) and registers it as an ordinary attachable session. It resolves the executable against the child's PATH first — a PTY has no shell to print "command not found", so on POSIX `execvp` fails in the forked child and it exits 1 with a blank screen.
- The restricted branch of `createAgentTuiSession` uses it, mirroring the shape the CoS runner's `/spawn-tui` has always had.
- **Ordinary agent TUI sessions keep the login shell** — the rc file is where a coding agent's toolchain (nvm, pyenv, PATH edits) comes from.
- Two handshake affordances went with the shell and are replaced the way the runner path already replaces them: the readiness probe that fired `onInitialCommandSent` (the paste gate now opens before the spawn) and the bracketed-paste OFF the input-ready tracker waited for. The paste liveness probe is skipped for the same reason — the pid *is* the CLI, so it has no child, and its exit fires `onExit` directly.
- `resolveTuiLaunchShape()` names the branch once (`runner` | `direct` | `login-shell`) so the pre-spawn handshake and the spawn itself cannot silently disagree — that failure mode is "the prompt is never delivered", with no error.
- `envReplace` is removed: it existed only to stop `buildSafeEnv` being unioned under this one caller's allowlist, and that caller no longer goes through `createShellSession`.
- Backbone rolled in: `adoptPtySession` now also backs `registerExternalSession`, so one place owns the session record and the 50KB re-attach ring buffer instead of three.

## Test plan

- `server/services/shell.test.js` — new `spawnCommandSession` cases: the PTY's process is the command itself (no shell binary, no `session.shell`), the child env is exactly the caller's (a `process.env` secret does not reach it), the resolved path is what launches, a missing binary is named before spawning, hooks/attach/cd behave.
- `server/services/agentTuiSpawning.test.js` — an actions-posture spawn goes through `spawnCommandSession` and **never** calls `createShellSession`; ordinary spawns still get the login shell; the prompt is still delivered with no shell paste-mode OFF and no live-child probe.
- Full server suite green: 1927 files / 38931 tests.
- Each new guard mutation-probed: reverting the launch-shape thread, the liveness-probe skip, the direct-spawn branch, or the resolved-path launch each fails its test.

Closes #6159

### DIFF
--- a/server/services/agentErrorAnalysis.js
+++ b/server/services/agentErrorAnalysis.js
@@ -962,7 +962,7 @@ export const COMPLETION_REASON_ANALYSES = {
     actionable: true,
     escalation: 'Confirm the required CLI/tool is installed and on PATH for the agent user (or fix the command), then approve the retry.',
     message: 'Failed to start the agent session',
-    suggestedFix: 'The shell/PTY session could not be created. Check system resources and the provider command configuration.'
+    suggestedFix: 'The PTY session could not be created. Check system resources, whether too many sessions are already open, and the provider command configuration.'
   },
   // The runner refused the spawn outright — a command missing from its
   // allowlist, malformed cliArgs, or the runner simply unreachable. No child
--- a/server/services/agentTuiSpawning.js
+++ b/server/services/agentTuiSpawning.js
@@ -93,10 +93,14 @@ const PROVIDER_SIGNAL_POLL_MS = 5000;
 // The filename is per agent instance — see doneSentinelName in ../lib/agentSentinel.js.
 
 /**
- * Thin wrapper around `shellService.createShellSession` for the agent TUI
- * path. Centralizes the agent-side defaults (kind, label, initialCommand)
- * and pairs the returned session id with its underlying pty process so
- * callers don't have to make a second `getSessionProcess` call inline.
+ * Open the agent TUI's PTY and pair the returned session id with its underlying
+ * pty process, so callers don't have to make a second `getSessionProcess` call
+ * inline. Centralizes the agent-side defaults (kind, label, initialCommand).
+ *
+ * Three shapes, in the order they are checked: the CoS runner's own PTY; a
+ * DIRECT PTY running the provider binary itself (public-content stages only —
+ * see the `restricted` branch); and otherwise an interactive login shell with
+ * the CLI command typed into it.
  *
  * Returns `{ sessionId, ptyProcess, pid }`. When the shell service fails
  * to create the session, `sessionId` is null and the caller is expected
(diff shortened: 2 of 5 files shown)


atomantic:

atomantic:
### Review audit trail

Configured reviewers for this run: `codex,claude` (both required — neither carries `~opt`).

| Reviewer | Result |
|---|---|
| **codex** | ❌ **blocked** — `ERROR: You've hit your usage limit … try again at Sep 7th`, on its first invocation. Account-level quota; the binary is healthy and on PATH (`codex-cli 0.153.0`). No verdict produced. |
| **claude** | ✅ **CLEAN** after 2 rounds. |

`claude` round 1 returned `VERDICT: FINDINGS` with one real defect: `spawnCommandSession` discarded its `findCommandOnPath` result and handed the **bare** command name to `prepareCliSpawn`, which re-resolves under different rules. On Windows the two disagree (`findCommandOnPath` unquotes a PATH entry, maps an empty one to `cwd`, resolves relative entries, and searches all of `PATHEXT`; `prepareCliSpawn`'s own lookup does none of that), so a quoted PATH entry holding `claude.cmd` would pass the guard and then fall through to a bare extensionless `claude` — which ConPTY cannot launch. That is the exact blank-PTY/exit-1 failure the pre-flight was added to prevent. Fixed in 1be8237b4 and covered by a regression test that fails against the pre-fix code on POSIX.

Round 2 was a convergence check scoped to that delta only: `VERDICT: CLEAN`.

No self-review was substituted for the blocked reviewer. Merge is withheld pending `codex`; the claim markers, branch, and worktree are left in place for it.

Also run before the reviewers: a 4-agent `/simplify` pass (reuse, simplification, efficiency, altitude) whose findings are applied in e99be3745.

PR #6209, opened 2026-09-04
fix: preserve active model tuning across context-window reloads (#6200)

### LINKED ISSUE #6200: a context-window reload clears an active local-model tuning's launch env

atomantic:
## Summary
Fixes `ensureContextWindow` in `server/services/ollamaManager.js` so that a context-window reload composes on top of the currently applied launch environment (`appliedLaunchEnvValues`) instead of replacing it with `{}`. This ensures active model tuning knobs (such as `OLLAMA_FLASH_ATTENTION`, `OLLAMA_KV_CACHE_TYPE`) survive context-window reloads triggered by agent spawns (e.g. `claude-ollama` requiring 128K context).

Additionally, preserves the tuning's undo bookkeeping (`preTuningEnv`) across non-tuning reloads and updates the baseline's context window, so that when a tuning sweep later completes and calls `clearLaunchEnv()`, it cleanly restores the untuned daemon with the updated context window rather than reverting it or stranding the tuning.

## Test plan
- Verified that all 72 tests in `server/services/ollamaManager.test.js` pass with Vitest.
- Added regression tests in `server/services/ollamaManager.test.js`:
  - `preserves an active tuning across ensureContextWindow on a spawned process`
  - `preserves an active tuning and its undo bookkeeping across a service context-window reload`
  - `preserves an active tuning when ensureContextWindow starts an offline daemon`
- Ran related test suites: `server/lib/ollamaContext.test.js`, `server/lib/localModelTuning.test.js`, `server/services/ollamaAgentContext.test.js`, `server/services/localModelAssessments.test.js` (all passing).

Closes #6200

### DIFF
--- a/server/services/ollamaManager.js
+++ b/server/services/ollamaManager.js
@@ -34,7 +34,7 @@ import { buildHfAuthHeaders, buildHfResolveUrl, HF_API } from '../lib/huggingfac
 import { isEmbeddingModel } from '../lib/localModelHeuristics.js'
 import { commandExists } from '../lib/commandExists.js'
 import {
-  OLLAMA_AGENT_MIN_CONTEXT, resolveOllamaContextLength, withOllamaContextEnv
+  OLLAMA_AGENT_MIN_CONTEXT, OLLAMA_CONTEXT_ENV_VAR, resolveOllamaContextLength, withOllamaContextEnv
 } from '../lib/ollamaContext.js'
 import { compareSemver } from '../lib/versionUtils.js'
 import { isSafeHfRepoRelativePath } from '../lib/hfCache.js'
@@ -640,8 +640,9 @@ async function getRuntimeContextLength(selectedModel = null) {
 async function ensureContextWindow(contextLength, selectedModel = null) {
   const target = Number(contextLength) > 0 ? Math.floor(Number(contextLength)) : null
   if (!target) return { applied: false, reason: 'not-configured', contextLength: null }
-  const env = withOllamaContextEnv({}, target)
   if (!(await checkOllamaAvailable(true))) {
+    const baseEnv = appliedLaunchEnv !== null && appliedLaunchEnvValues ? { ...appliedLaunchEnvValues } : {}
+    const env = withOllamaContextEnv(baseEnv, target)
     return { ...(await restartWithEnv(env, { tuning: false })), contextLength: target }
   }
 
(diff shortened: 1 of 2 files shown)


atomantic:

PR #6243, opened 2026-09-04
feat(daemons): release managed model daemons under host memory pressure (#5818)

### LINKED ISSUE #5818: [repo-study-model-daemon-memory-pressure] Release managed model daemons under host memory pressure, not only after an idle timer
> **Labels to apply** (the filing account has read-only access to this repo, so GitHub silently dropped them at creation): `repo-study`, `plan`, `area:devtools`, `model:heavy`, `effort:medium`, `planner:opus-5`.

**Provenance:** studied from `carloslfu/slotstream` (MIT, © 2026 Carlos Galarza), captured 2026-09-02. Clean-room — technique only, reimplemented against PortOS modules.

## Rationale

Upstream's governor treats a local model process's memory as *elastic*: it watches the OS's memory-pressure signal plus a slow availability poll and resizes between requests — shrink fast, grow slow, with absolute-GB dead-bands (a relative trigger can never fire when the honest adjustment is a couple of GB on a large pool), a calm-down window before growing back, and a hard rule that an explicitly user-set size is *never* touched because it is the user's stated intent.

PortOS's equivalent is `reapIdleDaemons()` / `registerIdleDaemon()` in `server/lib/managedDaemon.js`, and it knows only wall-clock idleness: a model server holding 20–40 GB resident keeps holding it until its idle window expires, however hard the rest of the machine is thrashing. The user notices this as the whole machine going sluggish while a model they finished using an hour ago sits on the memory — with nothing in the UI explaining why. The reading is already available: `server/lib/memoryStats.js` parses the authoritative `vm_stat` / `meminfo` numbers (deliberately, because `os.freemem()` lies on macOS), and `server/services/cosHealthMonitor.js` already reasons about host-wide pressure.

## Work

- Add a pressure-aware pass to the idle reaper in `server/lib/managedDaemon.js`: when `memoryStats` reports sustained host pressure above a threshold, release the *least recently used* managed model daemon early rather than waiting out its idle window. `daemonLastUsedAt()` already provides the ordering.
- Borrow the governor's shape, not its numbers: shrink fast / restore slow, an absolute-GB dead-band rather than a percentage, and a calm-down window so a transient spike cannot flap a daemon down and up. Release at most one daemon per tick and re-read before the next.
- **Never release a daemon the user pinned.** Add an explicit per-server "keep loaded" toggle next to the existing idle-release window in the Local Runtime Servers table; a pinned server is exempt exactly as an explicitly-sized pool is exempt upstream.
- Make it legible: record the reason on the daemon entry and show it in the **Models → LLMs** row ("released at 09:14 — host memory pressure"), so an unexpectedly stopped server is explained rather than mysterious. The existing lazy/on-demand start path already brings it back on next use.
- Keep the decision a pure function of (current state, pressure reading, recent history) so it is testable without putting the machine under real pressure — mirroring how `reapIdleDaemons` is already unit-tested.

**Fix:** `server/lib/managedDaemon.js` (`reapIdleDaemons`, `registerIdleDaemon`, new pressure policy), `server/lib/memoryStats.js` (consumer only), per-server settings in `server/services/settings.js`, `client/src/components/models/ModelsPanel.jsx`.

**Scope:** medium — one new policy function plus a settings toggle and a UI reason string; the reaper, the memory reading, and the lazy-start path all already exist.

atomantic:
## Summary
- Adds memory-pressure-aware eviction pass to `reapIdleDaemons` in `server/lib/managedDaemon.js`:
  - Pure policy function `evaluateMemoryPressurePolicy` evaluating free memory threshold (4GB default), dead-band hysteresis (2GB default), sustained duration (30s default), and calm-down window (60s default).
  - Releases the least recently used unpinned daemon early when host memory is under sustained pressure, rather than waiting out the idle window.
  - Formats human-legible release reason (`"released at HH:MM — host memory pressure"`).
  - Displays release reason warning text in the **Models → LLMs** runtime rows.
- Adds per-runtime "Keep loaded" toggle to exempt daemons from both idle unloading and pressure-based eviction:
  - Settings updated in `localLlmSettingsSchema` for `llama`, `mtplx`, and `slotstream`.
  - Disables idle window input when "Keep loaded" is active.
  - Prevents `--sleep-idle-seconds` from being added to `llama-server` launch arguments when pinned.
  - Fail-safe handles `isPinned()` evaluation errors by treating daemons as pinned to avoid accidental evictions.

Closes #5818

## Test Plan
- Unit tests:
  - `server/lib/managedDaemon.pressure.test.js`: 18 tests covering memory pressure policy, hysteresis, calm-down, LRU ordering, pinned exemption, fail-safe error handling, and reaper integration.
  - `server/services/llamaServerManager.test.js`: verified `--sleep-idle-seconds` omitted when `keepLoaded` is configured.
  - `client/src/components/settings/RuntimeServersCard.test.jsx`: 31 tests covering "Keep loaded" toggle interaction and `releaseReason` rendering.
  - `npm test -- lib/index.test.js`: catalog and barrel export validation.


atomantic:

PR #6267, opened 2026-09-04
fix(github): require authenticated account for repository sync

tzioup:
## Summary

- derive repository ownership from a verified GitHub CLI credential instead of a baked-in account
- add authentication/account status UI and fail closed when authentication or cached ownership does not match
- pin repository and secret mutations to the credential that was verified, with bounded command execution
- preserve per-account repository configuration and avoid deleting cached rows when GitHub's 200-repository limit is reached
- keep healthy account management compact while retaining prominent signed-out and error guidance

## Root cause

The initial GitHub state seeded from the atomantic even on a local fork, and sync used that stored/default owner without requiring a usable authenticated account, or offering an authentication pathway. A signed-out install could therefore list the maintainer's public repositories anonymously, and a fork user's remotes were not visible/identifiable in relation to the user's private Apps and projects within PortOS.

## Verification

- server GitHub, credential-inventory, and import-scope tests: 67 passed
- client GitHub and Credentials UI tests: 13 passed
- full client suite: 10,326 passed, 6 skipped
- client lint: clean
- production client build: passed
- live local acceptance: authenticated account detection, account switching through Shell, and repository sync
- Playwright desktop/mobile checks, including a maximum-length fake login and dropdown viewport containment
- independent Codex review after the final upstream rebase: PASS, no blockers

The broad server suite also completed 39,138 tests with 49 skipped; 13 unrelated environment-sensitive tests failed under the restricted local runner (process inspection/spawn permissions, watcher exhaustion, and a protected home-directory write). The changed server tests and the import-budget gate pass on the final rebased tree.

### DIFF
--- a/client/src/components/settings/CredentialsTab.jsx
+++ b/client/src/components/settings/CredentialsTab.jsx
@@ -71,6 +71,7 @@ export function CredentialsTab() {
       <div className="space-y-3">
         {credentials.map((credential) => {
           const configured = credential.configured === true;
+          const verificationUnavailable = credential.verification === 'unavailable';
           return (
             <div
               key={credential.id}
@@ -82,11 +83,19 @@ export function CredentialsTab() {
                   <p className="text-sm text-gray-400 mt-1">{credential.unlocks}</p>
                 </div>
                 <div className="flex flex-wrap items-center gap-2">
-                  <span className={`inline-flex items-center min-h-[28px] px-2 rounded text-xs ${configured ? 'bg-emerald-500/15 text-emerald-300' : 'bg-port-border/60 text-gray-400'}`}>
-                    {configured ? 'Configured' : 'Not configured'}
+                  <span className={`inline-flex items-center min-h-[28px] px-2 rounded text-xs ${
+                    verificationUnavailable
+                      ? 'bg-amber-500/15 text-amber-300'
+                      : configured
+                        ? 'bg-emerald-500/15 text-emerald-300'
+                        : 'bg-port-border/60 text-gray-400'
+                  }`}>
+                    {verificationUnavailable ? 'Could not verify' : configured ? 'Configured' : 'Not configured'}
                   </span>
                   <span className="inline-flex items-center min-h-[28px] px-2 rounded text-xs bg-port-bg text-gray-400">
-                    {SOURCE_LABEL[credential.source] || SOURCE_LABEL.none}
+                    {verificationUnavailable && credential.source === 'none'
+                      ? 'Status unknown'
+                      : SOURCE_LABEL[credential.source] || SOURCE_LABEL.none}
                   </span>
                   {credential.tier && credential.tier !== 'none' && (
                     <span className="inline-flex items-center min-h-[28px] px-2 rounded text-xs bg-port-bg text-gray-500">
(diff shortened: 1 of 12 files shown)


atomantic (commented):
## Review summary

Ran a 5-lens review (surface scan, security audit, cross-file contract, cross-file tracing, surface quality) against this PR plus my own pass. Since I have write access to this branch (maintainer edits are enabled), I pushed fixes directly rather than leaving inline suggestions — commit 1d5665d.

**Two critical gaps, both fixed:**

1. **`updateRepoFlags` was the one mutation not pinned to the verified credential** — all five lenses independently flagged this. It trusted the cached `githubUser` field instead of re-checking live `gh` auth, so a signed-out or account-mismatched request could still queue `NPM_TOKEN` (or any `managedSecrets` value) onto a repo for a later secret push to act on. Now gated by `requireActiveCachedAccount`, same as `setSecret`/`syncSecretToRepos`/`setRepoArchived`.

2. **A scope-degraded credential's valid listing was treated as authoritative for deletion** — a token that lost `repo` scope still exits 0 with a complete-looking listing of public repos only (not truncated), so every private repo of the same owner — and its `flags`/`managedSecrets` — was silently deleted on the next sync. Cached rows with no user configuration are still pruned when genuinely missing; rows carrying `flags` or `managedSecrets` are now stamped `missingFromRemote` instead of deleted.

**Also fixed:**
- Off-by-one in the truncation check: now requests `REPO_SYNC_LIMIT + 1` so an account with exactly 200 repos isn't misclassified as truncated (and permanently loses pruning).
- Unguarded `JSON.parse` on the `gh repo list` output — now uses `safeJSONParse` like `findPullRequestForBranch` already does, returning a 502 instead of a bare `SyntaxError`.
- `setRepoArchived` no longer fabricates an unpersisted "success" object when the account changes mid-request; it reports the same 409 its sibling checks use.
- `gh auth token --user` has no fallback for older `gh` builds — added one, since installs upgrade `gh` independently of PortOS.
- `GitHub.jsx` now re-derives auth status from the server after every sync instead of patching `githubUser` alone. The old partial patch left `login` stale, which could invert the account-mismatch banner and leave every repo action disabled after a legitimate account switch, and silently dropped a sync's 401/409 without refreshing the UI's stale "authenticated" state.
- `lastRepoSyncTruncated` is now persisted so the 200-repo-limit banner survives a page reload instead of only showing for one render.
- Added `scripts/migrations/341-github-repos-owner-reset.js`: any install whose `data/github-repos.json` still carries the old hardcoded `'atomantic'` default gets it cleared, since `getRepos()` now filters by that field (per AGENTS.md, on-disk semantic changes need a migration).

Test suites for all touched files pass (52 server + 13 client), with new coverage for the fail-closed gate, the scope-loss/tombstone behavior, the exact-limit boundary, and the older-`gh` fallback.

**Left as-is / worth a follow-up, not blocking:**
- `getGitHubAuthStatus` has no TTL cache (unlike `checkGhHealth`'s 60s one), so Settings → Credentials and every mutation spawn their own `gh api user`. Didn't touch this — caching auth status on a fail-closed path has real correctness stakes if done carelessly, and it's a cost issue, not a bug.
- The Credentials tab reports a rejected `GH_TOKEN` as "Not configured" while the GitHub page correctly names it as "rejected" — same underlying status, different fidelity on each page.
- `detectorMustVerify = entry.id === 'github'` in `credentialInventory.js` hardcodes a registry id inline; a future rename of that id would silently revert GitHub to unverified truthiness.
- A few test-quality nits (duplicated child-process factories, a `truncated: true` client fixture with an empty `repos` map, an `execGh` timeout test/comment claiming "redaction" it doesn't perform).

Nice work isolating the account-verification logic and covering the fail-closed paths — the account-mismatch tests in particular are genuinely pinned (they assert the mutation never spawned, not just the status code).

PR #6272, opened 2026-09-04
Make voice fine-tuning lifecycle tests deterministic, and fix the cancelled-run outcome they exposed

### LINKED ISSUE #6268: Make voice fine-tuning lifecycle tests reliable under Windows CI contention
## Observed failure
Windows server shard 3/3 failed in [CI run 33924165423, attempt 2](https://github.com/atomantic/PortOS/actions/runs/33924165423/job/101189861213): `services/voice/fineTuning.test.js > runs fine tuning lifecycle, emits checkpoints, and promotes checkpoint` still observed `running` after its 5-second polling budget. The suite took about 6.1 seconds; sibling jobs were cancelled.

This test launches the real Python runner through `startFineTuningJob`. Other lifecycle tests in the same file use the same 5-second wait. The server runner allows 30-second tests/hooks on Windows specifically because process startup and filesystem work are slower under contention, but these inner waits retain a shorter deadline. The current failure does not establish a production lifecycle defect.

## Decision
Reproduce the Windows CI-shaped run and make lifecycle verification deterministic: drive process events through a controlled child-process fixture for state-machine cases, retaining one boundary test for actual runner wiring. For the real process boundary, use the repository's platform-aware test budget and explicitly drain terminal persistence before teardown. Do not weaken completion/checkpoint assertions or add repeated production sleeps.

Affected files: `server/services/voice/fineTuning.test.js`, the subprocess boundary in `server/services/voice/fineTuning.js`, and shared test helpers if an existing child-process fixture fits.

## Acceptance
- Lifecycle, checkpoint, promotion, cancellation, and terminal-sidecar assertions remain covered.
- Windows full-shard execution reliably completes without orphaned subprocesses or temp-directory cleanup races.
- Process timing and product state-machine regressions remain distinguishable.

atomantic:
## Summary

Windows server shard 3/3 failed on `services/voice/fineTuning.test.js > runs fine tuning lifecycle…` with the job still `running` after a fixed 5-second inner poll. Every lifecycle case launched the real Python runner, so the deadline was measuring interpreter startup under CI contention, not the state machine.

- **State-machine cases now drive a scripted child-process double** — frames, exit code, and node's abort ordering under test control. Lifecycle, checkpoint, promotion, cancellation, sidecar, restart-recovery and eviction assertions are unchanged in strength and no longer depend on process timing. The cancellation case no longer leaves a 2500-step trainer running to be killed at teardown.
- **One boundary case still spawns the real runner** so the scripted frames stay honest against `scripts/qwen3_tts_runner.py`. It takes the repository's Python-shelling budget (`PY_TEST_TIMEOUT_MS`) with a strictly smaller inner poll, so a genuinely stuck run reports the status it observed instead of a bare vitest timeout.
- **Every case drains the job's write chain before teardown** rather than letting the temp-dir removal retry through an in-flight `atomicWrite`.

Making the abort path deterministic surfaced two real defects in the subprocess boundary, both fixed here:

- **A cancelled run persisted as `failed`.** An aborted spawn fires *both* terminal events — `error` with an `AbortError`, then `close(null, 'SIGTERM')` — and the `error` handler overwrote the `cancelled` state that `cancelFineTuningJob` had just written. The operator's next status poll, and the `job.json` sidecar after a restart, both reported `failed`. Both handlers now share one guard: the first terminal event writes the outcome, later ones only finalize.
- **A killed trainer reported "Process exited with code null".** A child killed by a signal has no exit code, so an OOM reap partway through a long run said nothing useful. It now names the signal.

Finally, the service is now tagged in `WINDOWS_RISK_RULES` and its suite added to the Windows contract baseline (`scripts/ci-test-plan.js`). It was untagged, so the Windows shard only saw it on a full-matrix run — which is why the failure landed on `main` rather than on the PR that introduced it.

## Test plan

- `cd server && npm test` — 1958 files / 39366 tests pass.
- `npx vitest run services/voice/fineTuning.test.js` — 9 pass; the 8 scripted cases run in ~27–49 ms each (previously ~5 s of real-process wall time each), and the real-runner boundary case still exercises actual spawn wiring.
- Bypass probe: reverting the `error`-handler guard alone fails `settles a cancelled job as cancelled…` and nothing else, confirming that assertion is not vacuous.
- `npx vitest run routes/voice.test.js services/voice/` — 621 pass, so the route layer sees no behavior change.
- `npx vitest run ../scripts/ci-test-plan.test.js ../scripts/run-ci-tests.test.js` — 40 pass. A diff touching only `server/services/voice/fineTuning.js` now plans `windows: true, mode: related` and selects `fineTuning.test.js`; this PR itself plans a full matrix, so the Windows shard that originally failed runs here.

Closes #6268

### DIFF
--- a/scripts/ci-test-plan.js
+++ b/scripts/ci-test-plan.js
@@ -87,6 +87,13 @@ const WINDOWS_RISK_RULES = [
   /^server\/services\/(?:shell|pm2|appBuilder)\b/,
   /^server\/services\/agentTuiSpawning(?:\.test)?\.js$/,
   /^server\/services\/autonomousJobs\/execution\.shellSpawn/,
+  // The voice fine-tuning service spawns a real Python trainer and settles the
+  // job from its `close`/`error` events, so its assertions are the child-process
+  // contract, not platform-independent logic. Nothing here was Windows-tagged,
+  // so the shard only saw it on a full-matrix run — which is where it failed,
+  // with an interpreter-startup budget mistaken for a state-machine defect
+  // (#6268).
+  /^server\/services\/voice\/fineTuning(?:\.test)?\.js$/,
   /^server\/routes\/apps\//,
   /^server\/routes\/scaffoldVite\.js$/,
 ];
@@ -138,6 +145,7 @@ export const WINDOWS_CONTRACT_TESTS = [
   'server/services/agentTuiSpawning.test.js',
   'server/services/agentImportCycles.test.js',
   'server/services/twinImportCycles.test.js',
+  'server/services/voice/fineTuning.test.js',
 ];
 
 // Contract guards that run on EVERY plan, whatever the impact scope selects.
--- a/server/services/voice/fineTuning.js
+++ b/server/services/voice/fineTuning.js
@@ -44,6 +44,11 @@ const JOB_RECORD_FILE = 'job.json';
 // built from a caller-supplied id.
 const JOB_ID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
+// A user cancel is recorded by `cancelFineTuningJob` and, when the abort reaches
+// the child before the caller returns, by the child's terminal handler. One
+// shape so whichever writes first leaves the same record.
+const CANCELLED_OUTCOME = Object.freeze({ status: 'cancelled', error: 'Cancelled by user' });
+
 const jobRecordPath = (profileId, jobId) =>
   join(profileArtifactDirectory(profileId), 'fine-tune', jobId, JOB_RECORD_FILE);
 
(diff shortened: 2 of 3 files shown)


atomantic:
Work on this branch is complete and pushed; CI is still running. Intentionally left open and **not merged** — the swarm run that produced it was stopped after the issues were reported as also being picked up on another machine, and an unattended merge is not the right call under that collision.

Claim markers on #6268 are deliberately left in place so the branch keeps reading as in-flight rather than inviting duplicate work. Merge once CI is green (the Windows server shard is the one this change targets).

atomantic:
Review-loop verification completed for `82d4a442c387fda35b58eea507f36782f29beecb`.

- Antigravity (`gemini-3.8-flash`, optional, one-round cap): skipped/inconclusive. The `agy` binary is installed, but the supplied reviewer procedure marks it unavailable because it has no enforced read-only mode for public-content review. No substitute reviewer ran and no review verdict is claimed.
- Local verification: 661 tests across 25 voice and CI-planning suites passed; the client production build passed.
- [CI attempt 2](https://github.com/atomantic/PortOS/actions/runs/33928059351/attempts/2): all 12 jobs passed, including all Windows shards, CI Gate, and Full CI Gate. Windows shard 1 previously failed in the existing `detachedSpawn.test.js` PID-file readiness check, then passed on this same commit after one rerun; treated as an intermittent CI failure.
- No review findings or unresolved review threads were present. No additional code changes or commits were needed.

Proceeding with the requested merge commit under the explicit optional-review policy.

Issue #6282, opened 2026-09-05
[ref-watch-phosphene-repeat-failures] Hold repeatedly failing video jobs


atomantic:
From `reference-watch` review of Phosphene (commit `10f822d7`, 2026-09-05).
Reference: https://github.com/mrbizarro/phosphene/commit/10f822d7d12eae698aa128f830d136fc6ece06d2 (MIT). Adopt; clean-room behavior only.

Phosphene stops consuming queued renders when the same failure repeats. PortOS currently archives every failed job in `startLaneJob()`, releases the lane, and lets `drainLoop()` immediately promote the next job: a missing/broken local video runtime can therefore consume an entire queued batch without preserving the remaining work for repair.

Fix: Extend `server/services/mediaJobQueue/index.js#startLaneJob()`, `#drainLoop()`, `#runJob()`, `#persistImpl()`, and `#initMediaJobQueue()` to hold local-video jobs for an affected model/runtime after three consecutive matching failures in that cohort. Resolve the model/runtime through the existing media-model catalog; use a stable failure classification plus normalized cause, never just the generic exit code, as the failure identity. Add a validated explicit resume endpoint in `server/routes/mediaJobs.js`, expose held state through `server/services/mediaJobQueue/sanitizeJob.js`, and wire a resume action through `client/src/services/apiMediaJobs.js` and `client/src/components/media/MediaJobsQueue.jsx#MediaJobsQueue()`.

Chosen behavior:
- Count terminal failures once, including a pre-dispatch throw; cancellations and startup reconciliation are not repeated engine failures. A successful render resets its cohort's streak; a different failure starts a new streak.
- After the third matching failure, keep matching pending jobs queued in their original order, with their original ids, parameters, owners, and staged inputs. Do not automatically retry failed jobs.
- Gate only the affected local video model/runtime cohort. Independent local models, image/audio/training work, cloud jobs, and federated jobs remain eligible for their existing lanes.
- Show the normalized cause and held-job count with an explicit Resume action. Resume clears that hold and streak; it does not repair/install a runtime or start any extra AI work.
- Persist active holds alongside the existing jobs envelope so restarting does not consume the remaining batch. Accept old envelopes with no holds; follow the repository migration/storage contract for this additive local state and retain the list API's array shape.
[...]

PR #6297, opened 2026-09-05
fix: surface the render child's own cause on an ordinary nonzero exit

### LINKED ISSUE #6281: [ref-watch-phosphene-render-error-cause] Explain failed video renders
From `reference-watch` review of Phosphene (commit `10f822d7`, 2026-09-05).
Reference: https://github.com/mrbizarro/phosphene/commit/10f822d7d12eae698aa128f830d136fc6ece06d2 (MIT). Adopt; clean-room behavior only.

Phosphene now includes a useful engine diagnostic when H3 exits unsuccessfully. PortOS has the same current gap: `spawnWatch.js#spawnAndWatchVideo()` retains stderr for the Metal retry, but its nested `handleChildClose()` reduces every ordinary nonzero exit without a missing-module match to `Exit code N`, discarding the actual exception from the queue and UI.

Fix: In `server/services/videoGen/spawnWatch.js#spawnAndWatchVideo()` (the stdout/stderr readers and `handleChildClose()`), retain a bounded diagnostic tail and include the final meaningful exception/cause in the ordinary nonzero-exit reason. Extend `server/services/videoGen/generateVideoHelpers.js` for diagnostic selection if needed, reusing `server/lib/secretText.js#scrubSecretTokens()` for credentials. Keep the existing missing-module repair guidance, signal diagnosis, cancellation handling, prompt-encoding retry, and successful completion-watchdog behavior authoritative. The same concise reason must reach terminal SSE and `videoGenEvents.failed`, then the existing `server/services/mediaJobQueue/index.js#runJob()` error persistence and `MediaJobsQueue.jsx#JobRow()`.

Acceptance:
- An H3/CUDA/MLX process that prints an exception and exits 1 shows that cause in its failed media-job record and UI; terminal messages still carry the exit code.
- Capture an unterminated last line and a cause written to stdout as well as stderr; do not select a progress marker, echoed prompt, or a full traceback as the summary.
- Bound retained bytes as well as line count. Fall back to the exit-code message when no useful diagnostic exists; strip credential-shaped content before persistence or emission.
- Extend the existing process-boundary cases in `server/services/videoGen/local.test.js` with a realistic nonzero-exit traceback and chunked final line. Preserve tests for missing modules, cancellation, signal death, and the one-shot Metal retry.
- Works on macOS and Windows without platform-specific traceback assumptions. No upstream source, prose, prompts, or assets copied.

Estimated scope: small (runner error handling and focused process-boundary coverage).

Bryandero98:
## Summary

Closes #6281.

`spawnWatch.js#handleChildClose()`'s fallback branch reported only a bare `Exit code N` whenever a video-gen child exited nonzero with no death signal and no `ModuleNotFoundError` match — discarding whatever the child actually printed to explain why.

## What this adds

- Two new `createOutputTail()` instances (`stdoutDiagTail`/`stderrDiagTail`) in `spawnAndWatchVideo`, fed only the lines neither the child's own STATUS:/STAGE:/RUNTIME:/SPEEDPROFILE: protocol nor `PYTHON_NOISE_RE` already claimed as noise — the same gate `stdoutReader`/`stderrReader` already use to decide what's worth raw-logging. A progress bar, a dependency warning, or a STAGE marker can therefore never win over the real cause, and each tail's own char budget means this can never surface an unbounded traceback.
- `generateVideoHelpers.js#pickExitDiagnostic()`: picks between the two tails (stderr first — a raised exception normally lands there; a Python traceback also prints its exception message LAST, so the tail's most recent line is almost always that message rather than a stack frame above it) and scrubs credential-shaped tokens via the existing `scrubSecretTokens()` before the reason can reach the job record or the UI.
- The fallback branch now reports `Exit code N: <cause>` when a cause was found, falling back to the bare `Exit code N` exactly as before when neither stream printed anything useful — terminal messages still carry the exit code either way.
- Kept separate from the existing `stderrTail` array, which the Metal-retry sniffer needs UNFILTERED (it greps for the `STAGE:encode-prompt` markers this pair deliberately excludes).

Cross-platform by construction: no traceback-format assumptions, just the last bounded window of otherwise-unrecognized output.

Clean-room implementation from the issue's own description; the referenced Phosphene commit was not opened.

## Test plan

- `cd server && npx vitest run services/videoGen/local.test.js services/videoGen/generateVideoHelpers.test.js` — 359 passed, 8 skipped (unrelated), 0 failed.
- New tests added per the issue's acceptance criteria:
  - `pickExitDiagnostic` unit tests (null when empty, stderr-over-stdout preference, stdout fallback, credential scrubbing) in `generateVideoHelpers.test.js`.
  - A new `generateVideo — ordinary nonzero exit surfaces a cause instead of a bare exit code (#6281)` describe block in `local.test.js`: a realistic multi-line traceback, a final line split across two chunks with no trailing newline (flushed on close), STATUS:/tqdm noise correctly excluded from the reported cause, a stdout-only fallback, and the bare-exit-code fallback when neither stream printed anything.
  - Existing missing-module, cancellation, signal-death, and one-shot Metal-retry process-boundary tests were left untouched and still pass.

### DIFF
--- a/server/services/videoGen/generateVideoHelpers.js
+++ b/server/services/videoGen/generateVideoHelpers.js
@@ -13,6 +13,7 @@ import { broadcastSse } from '../../lib/sseUtils.js';
 import { generateThumbnail, optimizeForStreaming } from '../../lib/ffmpeg.js';
 import { formatBytes } from '../../lib/fileUtils.js';
 import { renderTimingFields } from '../../lib/renderTiming.js';
+import { scrubSecretTokens } from '../../lib/secretText.js';
 import { videoGenEvents } from './events.js';
 
 /**
@@ -393,6 +394,30 @@ export function describeSignalDeath(signal, { fingerprint = null } = {}) {
   const cause = SIGNAL_DEATH_CAUSES[signal] || `Killed by signal ${signal}`;
   const fp = formatRuntimeFingerprint(fingerprint);
   return fp ? `${cause} [runtime: ${fp}]` : cause;
+}
+
+/**
+ * Pick the cause text for an ordinary (no signal, no missing-module match)
+ * nonzero exit — the gap that otherwise reports only `Exit code N` with
+ * nothing else to go on (issue #6281). `stdoutText`/`stderrText` come from
+ * two `createOutputTail()` instances in `spawnWatch.js` fed ONLY the lines
+ * neither the child's own STATUS:/STAGE:/RUNTIME:/SPEEDPROFILE: protocol nor
+ * `PYTHON_NOISE_RE` claimed, so a progress bar, a dependency warning, or a
+ * STAGE marker can never win over the real cause — and each tail already caps
+ * itself by a char budget, so this can never surface an unbounded traceback.
+ * stderr wins over stdout (a raised exception normally lands there); a Python
+ * traceback also prints its exception message LAST, so the tail's most recent
+ * line is almost always that message rather than a stack frame above it.
+ * Returns null when neither tail collected anything, so the caller can fall
+ * back to the bare exit-code message.
+ * @param {object} opts
+ * @param {string} [opts.stdoutText]
+ * @param {string} [opts.stderrText]
+ * @returns {string|null}
+ */
+export function pickExitDiagnostic({ stdoutText = '', stderrText = '' } = {}) {
+  const chosen = (stderrText || stdoutText || '').trim();
+  return chosen ? scrubSecretTokens(chosen) : null;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


Bryandero98:
Closing as redundant. While rebasing this onto an updated `main`, I found that commit `d6a179ff9` (part of the unrelated issue #6282, "hold local video batches after repeated runtime failures") already introduced `server/lib/videoFailure.js` with `createVideoDiagnosticTail()`/`normalizeVideoFailure()`, and `spawnWatch.js#handleChildClose()` now calls `diagnostics.failure(...)` in exactly the branch this PR was patching (previously a bare `Exit code ${code}`).

That implementation independently satisfies #6281's acceptance criteria: it captures both stdout and stderr tails, filters out progress/prompt/JSON noise before picking a summary line, scrubs credentials/paths/prompts/addresses, bounds retained bytes (8192) and lines (last 40), and falls back to the bare exit-code message when nothing useful is found — a superset of what this PR implemented.

No action needed on this PR; see #6281 for the pointer to the resolving commit.

Issue #6343, opened 2026-09-05
Extend execGh consecutive-failure backoff to remaining periodic gh readers


atomantic:
## Background

PR #TBD added an opt-in `backoffKey` option to `execGh()` (`server/services/github.js`) so a caller that polls the same `gh` read on every scheduler tick doesn't re-fire it on the very next tick after a failure. It was applied to two confirmed sites that both hit the same real incident (concurrent `gh` timeouts logged within the same few minutes): `branchReconcile.js`'s `getOpenPrsByHead` (keyed by repo spec) and `updateChecker.js`'s `checkForUpdate` (keyed by a fixed `'update-check'` string).

Other periodic/polling `execGh` callers have the identical "transient failure retried on the very next tick with no backoff" shape and were deliberately left out of that PR to keep it scoped to the confirmed incident. This issue tracks generalizing the fix to them.

## Decision

Apply `backoffKey` to periodic READ call sites only (mirrors the existing design: mutating/one-off calls opt out by not passing a key, since silently skipping those would be a correctness surprise, not a load-shedding win). Key each by something that uniquely identifies the repeated operation (repo spec, host, or a fixed string), matching the pattern in `branchReconcile.js`/`updateChecker.js`.

**Sites to convert (all periodic polling reads):**
- `server/services/prWatcher.js:71` — `gh api user --hostname <host>` (self-login check). Key by `host`.
- `server/services/prWatcher.js:100` — `gh repo view <repoSpec> --json defaultBranchRef`. Key by `repoSpec`.
- `server/services/prWatcher.js:115` and `:281` — the PR-poll reads. Key by `repoSpec`.
- `server/services/issueReconcile.js:326` — `ghList` helper wrapping `execGh(args)`, used for periodic zombie-issue scans. Key by `repoSpec` (available in the enclosing scope).
- `server/services/blockedIssueReconcile.js:81` — same `ghList` shape as above, periodic blocked-issue scan. Key by `repoSpec`.
- `server/services/taskDataInputs.js:128` — already passes `{ cwd, env }`; add `backoffKey` if this call is on a recurring poll path (verify before converting — confirm it's not a one-off user-triggered fetch).

**Sites to leave alone (mutating/one-off — no backoffKey):**
- `issueReconcile.js:683,689` and `blockedIssueReconcile.js:214,217` — `issue comment` / `issue edit` (mutations).
[...]

PR #6345, opened 2026-09-05
fix: back off repeated GitHub CLI reads

atomantic:
## Summary
- apply consecutive-failure backoff to periodic GitHub read paths in PR, issue, blocked-issue, branch, and update reconciliation
- preserve explicit/manual update checks and mutating GitHub calls as immediate operations
- guard per-process backoff outcome settlement across child lifecycle events

Closes #6343

## Validation
- `npm test -- services/github.test.js services/branchReconcile.test.js services/updateChecker.test.js services/prWatcher.test.js services/issueReconcile.test.js services/blockedIssueReconcile.test.js services/taskDataInputs.test.js` — 418 passed
- `npm test` — 39,846 passed, 35 skipped

## Review gate
- Required local `agy --effort medium` reviewer found and the fix was applied for one low-severity child lifecycle finding.
- Follow-up invocation timed out after 5 minutes without a verdict; review status is `review-blocked`. Per delivery policy this PR is left open for cold pickup and is not being merged.

PR #6346, opened 2026-09-05
fix: back off gh's consecutive-failure calls instead of retrying every tick

atomantic:
## Summary
- A `gh` blip let branch-reconcile and updateChecker re-fire the same 60s-timeout `gh pr list` / release-check call for every managed repo on every scheduler tick with no cooldown — logs showed repeated timeouts across multiple repos within a few minutes during a real incident. `ensureForgeReachable`'s cheap health probe doesn't catch this: it can pass while the heavier read still hangs.
- Adds an opt-in `backoffKey` to `execGh()` with exponential backoff per key, cleared on success. Mutating/one-off gh calls are unaffected (no key = no backoff — silently skipping those would be a correctness surprise, not a load-shedding win).
- Wired into branch-reconcile's PR-list read (keyed by repo spec) and updateChecker's release check (keyed by a fixed string, with `backoffMaxMs` sized to exceed its own 30-min scheduler interval — the default 15-min cap would otherwise always expire before the next tick and never actually suppress anything).
- A manual "check for updates now" click always bypasses the backoff (`{ manual: true }`), so an explicit user action is never silently swallowed by a cooldown the background scheduler armed on its own.
- Filed #6343 to extend `backoffKey` to the remaining periodic `gh` readers (prWatcher, issueReconcile, blockedIssueReconcile).

## Test plan
- `cd server && NODE_ENV=test npx vitest run services/github.test.js services/branchReconcile.test.js services/updateChecker.test.js routes/update.test.js` — all green, including new regression tests for the backoff mechanics, the idempotent-settle fix, and the configurable cap.
- Full server suite green except two pre-existing/unrelated items: a known load-flaky test (`eidoverseWorld.runtime.test.js`, passes in isolation) and local submodule pointer drift in `lib/slashdo` (not part of this change).

### DIFF
--- a/server/routes/update.js
+++ b/server/routes/update.js
@@ -44,9 +44,12 @@ router.get('/status', asyncHandler(async (req, res) => {
   res.json({ ...status, activeCosAgents, persistentMindImages });
 }));
 
-// POST /api/update/check — triggers manual check
+// POST /api/update/check — triggers manual check. `manual: true` bypasses the
+// unattended scheduler's gh backoff — a user who explicitly asked for a check
+// must get a real attempt, not a silent rejection from a cooldown the
+// background poller armed on its own.
 router.post('/check', asyncHandler(async (req, res) => {
-  const result = await updateChecker.checkForUpdate();
+  const result = await updateChecker.checkForUpdate({ manual: true });
   res.json(result);
 }));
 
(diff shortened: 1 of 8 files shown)


atomantic:
Superseded by #6345, which already merged an equivalent (and more complete — it also covers prWatcher/issueReconcile/blockedIssueReconcile) fix for this issue via an independently-claimed swarm worker on #6343 while this PR's review was in flight. Filing a small follow-up for the one thing #6345 doesn't have: a configurable backoff cap, since updateChecker's 30-min scheduler interval exceeds the hardcoded 15-min cap and the backoff never actually suppresses a retry there.

PR #6532, opened 2026-09-07
feat: same-origin Eidoverse host for single-port tailcat forwards

atomanticagent:
## Summary

- **Symptom:** A single-port [tailcat](https://github.com/tailscale/tailcat) forward (`127.0.0.1:15555` → remote `:5555`) breaks the Eidoverse iframe. `hostUrlFor` pointed at `hostname:5563`, which is not tunneled; Managed Update–style flows on `:5555` were fine.
- **Why a path prefix alone is not enough:** The eidoverse-worlds client uses many **absolute root** URLs (`WebSocket` → `/ws`, `fetch('/version')`, `/avatars`, `/library/…`, `/upload`, `/defs`, `/animations`, `/whoami`, `/authcfg`, `/geom`, `/thumb`, `/perflog`, `/node_modules/…`, `/shared/…`, …). Those resolve to `${location.origin}/…`, not under a prefix.
- **Design:**
  1. Iframe URL → `${location.origin}/eidoverse-host/?world=…&name=…` (same host+port as the PortOS UI, including a forward port).
  2. Main `:5555` reverse-proxies `/eidoverse-host` → loopback `:8940` with the prefix stripped (relative assets like `./main.js` keep working).
  3. While the on-demand host is active, allowlisted root routes (from eidoverse-worlds `server/routes.ts` matchers) are also proxied — including WebSocket upgrade for `/ws` — mounted **before** the SPA fallback.
  4. `/embed-config` and `/host` terminate on PortOS; same-origin `parentOrigin` preserves the full browser `Host` (e.g. `http://127.0.0.1:15555`). Legacy `:5563` bridge still uses hostname + API port (not the bridge port).
  5. Vite proxies `/eidoverse-host` + root allowlist (with Referer bypass for `/node_modules/` / `/shared/` so Vite’s own deps are not stolen). `/eidoverse-host` added to `SERVER_OWNED_PREFIXES`.
  6. `:5563` still starts with `ensureEidoverseHost()` as an optional/legacy bridge; the iframe prefers the same-origin path.

PortOS-side only — no edits under `data/repos/`.

## Test plan

- [x] `server`: `eidoverseHost.test.js`, `eidoverseProxy.test.js`, `dev-proxy-drift.test.js` (28 passed)
- [x] `client`: `Eidoverse.test.jsx`, `EidoverseGuest.test.jsx` (29 passed)
- [ ] Manual: open Eidoverse via a tailcat forward (`:15555` → `:5555`); confirm iframe loads, `/version` and `/ws` succeed on the forward origin, and frame handshake completes (`parentOrigin` matches the forward origin)


atomanticagent:
Pushed import-budget bump for CI: measured 95,246 → ceiling 95,300 (same-origin Eidoverse proxy at boot). Other jobs were fail-fast cancels from that single failure.

PR #6555, opened 2026-09-07
test(agentCliSpawning): replace fixed post-close sleeps with vi.waitFor (#6552)

### LINKED ISSUE #6552: Replace the fixed post-close waits in agentCliSpawning.test.js with vi.waitFor conditions
## Problem

`spawnDirectly` returns before its child `close` handler runs, so `server/services/agentCliSpawning.test.js` waits a fixed 30–80 ms (`await new Promise((r) => setTimeout(r, 30))`) for the handler's `finally` block in roughly thirteen tests. Those waits are a flake vector under CI load, and their budget silently depends on how much the completion path loads: when the completion dispatch briefly became a lazy import, the first post-close test in the file started failing on a cold worker until a warm-up import was added.

## Decision

Convert each fixed post-close wait into a condition, using the pattern 58 server suites already use: `await vi.waitFor(() => expect(<the mock the test asserts on>).toHaveBeenCalled())` — for example `runSpawnerCompletionCleanup`, `finalizeAgent`, or `activeAgents.has(...)` being false. Keep the short pre-emit `setTimeout(r, 10)` yields that let the spawn settle; only the waits that follow a `fakeProcess.emit('close', …)` change. The TUI suite is already condition-based (`await capturedOnExit(...)` awaits `finish()` itself) and needs nothing.

## Acceptance

- No `setTimeout(r, 30)` / `setTimeout(r, 80)` remains after a `fakeProcess.emit('close', …)` in `agentCliSpawning.test.js`.
- Every assertion is unchanged; the suite passes and its wall time does not regress.

naba-runn:
Closes #6552.

The post-close waits in `agentCliSpawning.test.js` slept a fixed 30–80ms hoping the child's fire-and-forget `close` handler had reached its `finally` block before asserting. That's timing-dependent and flaky under CI load. This replaces those sleeps with `vi.waitFor`, which polls the assertion until it passes instead of guessing a duration.

**Converted (5 sites):** the waits that block on the close handler before assertions on `finalizeAgent`, `cleanupWorktreeFn`, `releaseRetryHold`, `handlePipelineProgression`, and `activeAgents` state.

**Deliberately left as-is:**

- The 10/20/50ms waits that sequence mid-stream `emit()` calls before `close` — these control event ordering, not the finally block, and polling would break them.
- Two waits guarding negative assertions (`handlePipelineProgression.not.toHaveBeenCalled()` and the `runToClose` host-restart cases). `vi.waitFor` would pass on the first poll and stop giving the handler time to (not) run, silently gutting the check — so the fixed delay is doing real work there.

Scoped to the close-handler waits the issue describes; happy to follow up on the output-batch telemetry waits if you'd like those too.

**Verification:** `cd server && npm test -- services/agentCliSpawning.test.js` → 68/68 passing. Only the test file changed.

### DIFF
--- a/server/services/agentCliSpawning.test.js
+++ b/server/services/agentCliSpawning.test.js
@@ -730,17 +730,18 @@ describe('stream error containment', () => {
     });
 
     await expect(spawnDirectly(minimalArgs)).resolves.toBe('agent-test');
-    await new Promise((r) => setTimeout(r, 30));
 
-    expect(finalizeAgent).toHaveBeenCalledWith(expect.objectContaining({
-      agentId: 'agent-test',
-      exitCode: 2,
-      success: false,
-      // The output the child managed to write before exiting is exactly what
-      // explains the failure, so it has to survive the buffering too.
-      outputBuffer: expect.stringContaining('unknown flag --nope'),
-    }));
-    expect(activeAgents.has('agent-test')).toBe(false);
+    await vi.waitFor(() => {
+      expect(finalizeAgent).toHaveBeenCalledWith(expect.objectContaining({
+        agentId: 'agent-test',
+        exitCode: 2,
+        success: false,
+        // The output the child managed to write before exiting is exactly what
+        // explains the failure, so it has to survive the buffering too.
+        outputBuffer: expect.stringContaining('unknown flag --nope'),
+      }));
+      expect(activeAgents.has('agent-test')).toBe(false);
+    });
   });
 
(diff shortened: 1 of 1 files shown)


naba-runn:
Thanks — I see #6552 was already resolved in #6557, so I'll close this as
superseded. Noting for anyone who finds this later: #6557 also converts the
negative-assertion sites by waiting on a positive completion signal and then
asserting the negative, which this PR left as fixed delays. Good pattern to know.

Issue #6547, opened 2026-09-07
Compile standalone video briefs into revisioned scripts and timed shot artifacts


atomantic:
Extend the Creative Director treatment/tool-plan contract for standalone multi-shot productions; do not add an orchestration engine or require Series. Resolve a requested duration range to a displayed exact target within it. Persist stable script/scene/shot/reference IDs, timing and artifact revisions; validate agent output/tool inputs and dependency references server-side.

Use shared Reactor/Grok/video capability limits for per-shot durations/prompts/aspect/starting frames/continuation; never send a 180-second single clip. Feed selected Universe/cast/style through existing canon renderers and catalog references; attach existing Music/voice/Series assets without mutating sources, preserve source context revision and expose deleted-source repair actions. Expose artifacts and ordinary tool summaries rather than private reasoning.

Acceptance: fake provider result for a 1–3 minute standalone short yields a bounded valid timed multi-shot plan; unresolved dependencies block consumers; incompatible backend inputs and deleted sources have clear failures. Produced artifacts link back to creative suite IDs; structured revisions survive reload. Update prompts through version/integrity tooling and add route/workflow tests, no paid rendering.
Areas: creativeDirectorPrompts, creativeDirector plan/treatment validation/projects adapters, planAdvance, sceneRunner, creative tool registry, existing Series scene-media/canon compilers.

Part of #6542

On merge, check this child in #6542. If it is the last open child, comment with the delivered child issues and close #6542; otherwise leave the epic open.

atomantic:
Taking a standalone prerequisite slice first: reject duplicate plan step IDs, missing dependency targets, and cyclic dependency graphs at the shared server schema before persistence or dispatch. This PR will use Refs rather than Closes. Revisioned scripts, bounded shot compilation, source context revisions/repair, prompt integration, and artifact presentation remain on this issue.

atomantic:
Done ✓: PR #6561 implements server-side plan identity/dependency validation and rejection-before-write/dispatch tests (awaiting merge).

Remaining ▢: revisioned scripts and stable scene/shot/reference artifacts; exact target selection and backend-bounded shot compilation; creative-source context/revisions and deleted-source repair; planner prompt integration and artifact presentation. The PR intentionally uses Refs #6547; this issue and its parent checklist must remain open.

atomantic:
[...]

PR #6576, opened 2026-09-07
fix: validate explicitly pinned Video shot drafts

atomantic:
Video treatment saves now reject unsupported durations for explicitly pinned Grok/Reactor backends, oversized Reactor prompts, unsupported Reactor aspect ratios, and first-shot continuation. Validation uses the shared clip contracts and preserves the saved artifact when a draft or prompt edit fails. Legacy Creative Director treatments keep their existing behavior.

Validation: 111 persistence, transform and import-scoping tests passed; git diff --check passed. No paid rendering or real database tests.

## Remaining

This is a partial validation slice. Source resolution/canon grounding and deleted-source repair, planner prompt integration, tool-plan artifact links and artifact UI remain in #6547. Compilation against inherited/local/fal capabilities and full starting-frame compatibility also remain. Explicit pin checks do not establish renderer availability or authorize dispatch.

Refs #6547

### DIFF
--- a/server/services/creativeDirector/projectsFile.test.js
+++ b/server/services/creativeDirector/projectsFile.test.js
@@ -84,6 +84,30 @@ const createVideo = () => file.createProject({
 });
 
 describe('Video treatment artifacts', () => {
+  it('rejects backend-incompatible drafts and edits before replacing a saved artifact', async () => {
+    const p = await createVideo();
+    await file.updateProject(p.id, { renderBackend: { video: { mode: 'reactor' } } });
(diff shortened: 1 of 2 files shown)


atomantic:
Pinned optional Ollama review completed. Its undefined-variable finding is false: applySceneUpdate defines updated from the stored scene and patch before calling validateVideoShot. Its ordering recommendation is also inapplicable: updateScene has no accumulated elapsed time; sorting by order selects the first playback shot. The persisted-write regression passes, including a rejected oversized prompt edit.

PR #6612, opened 2026-09-08
fix: align provider controls with the Claude launch harness

### LINKED ISSUE #6605: Give the cli/tui/api provider-type predicates a pure-leaf home so the client re-exports them and the Claude-harness gate keys on the type
## Problem

The `cli` / `tui` / `api` provider-type predicates have no pure-leaf home on the server, so the same test is restated in six places and the browser still carries a copy:

- `client/src/utils/providerTypes.js` declares `isTuiProvider` / `isCliProvider` / `isApiProvider` / `isProcessProvider` as a "mirror of `isTuiProvider` in `server/services/agentCliSpawning.js`" — a `services/` module, which the browser bundle cannot import. These are the last copies left after the provider predicates were converted to re-exports of `server/lib/providerModels.js` (the PR that closes the c82dcdf3a / #6364 sweep).
- `server/lib` restates `type === 'cli' || type === 'tui'` privately in `providerPrerequisites.js` (`const isProcessProvider`, ~L139), `providerVendors.js` (`isDirectBinaryProvider`, ~L157) and `providerFamilies.js` (~L29), and inline in `grok.js` (~L65) and `kimi.js` (~L90).
- The `services/agentCliSpawning.js` copy (~L300) has one consumer, `services/agentLifecycle.js` (~L440).

Two client-side consequences hang off the missing leaf:

1. `isClaudeCommandProvider` (client) is `commandBasename(command) === 'claude'`, while the server's own forwarding gate (`isLocalBackedClaude` in `lib/cliChildEnv.js`, and `providerFamilies.js` ~L29) is `isProcessProvider(p) && isClaudeCommand(p.command)` — blank command counts as Claude for a process record because the spawners default it to `claude`. For a blank-command `claude-ollama` record the server forwards only `MAX_THINKING_TOKENS`, but `generationControlsFor` (`client/src/utils/providerModels.js`) still offers temperature/top-p the spawn discards, and `ProviderForm.jsx` (~L509) shows the "spawned OpenCode provider" hint for a record that will launch `claude`.
2. `AGENT_HARNESS_PROVIDER_TYPES` (client) claims to be "in lockstep with the api-provider rejection in `services/agentProviderResolution.js`" (~L46), but that is an inline `provider.type === 'api'` with no named allowlist; every consumer uses the list as `.includes(provider?.type)` (`client/src/utils/providerSelection.js` ~L152, `client/src/components/cos/AgentJobProviderFields.jsx` ~L17, ~L49), i.e. `isProcessProvider`.

## Decision

Add one pure leaf, `server/lib/providerTypes.js`, that imports `PROVIDER_TYPES` from `lib/aiToolkit/constants.js` and exports `isCliProvider`, `isTuiProvider`, `isApiProvider`, `isProcessProvider`, plus `isClaudeHarnessProvider(p) = isProcessProvider(p) && isClaudeCommand(p?.command)` (import `isClaudeCommand` from `providerModels.js`; that module must not import the new leaf back — check with `staticImportClosure`). Then:

- Point `providerPrerequisites.js`, `providerVendors.js`, `providerFamilies.js`, `grok.js`, `kimi.js`, and `services/agentCliSpawning.js` / `agentLifecycle.js` at the leaf and delete the private restatements.
- `client/src/utils/providerTypes.js` re-exports the four type predicates and `isClaudeHarnessProvider` (keep the `isClaudeCommandProvider` name as an alias for existing imports), deletes its local declarations, and adds the names to the identity guard's shared-name list in `client/src/utils/providers.test.js`. The `generationControlsFor` fixtures there omit `type` today and need it (a fixture-fidelity fix, not a product regression).
- Delete `AGENT_HARNESS_PROVIDER_TYPES`; its three call sites use `isProcessProvider(provider)`. Have `services/agentProviderResolution.js` (~L46) read the same predicate instead of the inline literal.
- Optional, same class: the pure id/basename predicates in `grok.js` / `kimi.js` (`isGrokCliProvider` / `isGrokTuiProvider` and the kimi twins) are trapped behind those modules' `os`/`fs`/`crypto` imports; move them to the leaf so the client's `isGrokBuildCli` becomes a re-export rather than a composition.

Register the new file in `server/lib/index.js` and `server/lib/README.md` (the barrel/README guard in `server/lib/index.test.js` enforces it) and expect a small bump in `server/lib/importScoping.test.js`'s instantiation count — a zero-dependency leaf is the tolerated shape.

## Acceptance criteria

- `grep -rn "=== 'cli' || .*=== 'tui'\|=== 'tui' || .*=== 'cli'" server/lib server/services` matches only the leaf.
- `client/src/utils/providers.test.js` "shares exactly the re-exported names" lists the type predicates and `isClaudeHarnessProvider`; the identity rows pass.
- `generationControlsFor({ id: 'claude-ollama', type: 'tui', command: '', ollamaBacked: true })` returns the Claude shape (`{ temperature: false, topP: false, thinking: true }`); an `api` record with no command still gets its full controls.
- `cd client && node_modules/.bin/vite build` succeeds; `scripts/client-server-import-purity.test.js` passes.

atomantic:
## Summary

Claude providers with a blank command now show controls and credential guidance for the Claude binary they actually launch. API providers retain their sampling controls.

Share the CLI/TUI/API predicates through a browser-safe server module, retain the existing client Claude alias, and replace the agent-picker allowlist and repeated PortOS process checks. Keep agentLifecycle's mocked service boundary through a re-export. The vendored aiToolkit remains self-contained as required by AGENTS.md; its internal type checks and the trajectory schema's three-type validation remain outside this consolidation.

## Test plan

- Client provider utilities, agent-job fields, and provider form: 323 tests passed, including blank-command Claude controls and credential guidance.
- Server provider, dispatch, lifecycle, and structural checks: 321 tests passed; additional caller coverage: 100 tests passed.
- Rechecked the tracked module catalog, import-scoping, and client/server import-purity guards: 44 tests passed.
- Production Vite build passed; git diff whitespace check passed.

Closes #6605


atomantic:
Pinned tool-free Ollama review returned no findings. The first CI attempt was canceled after the Windows detachedSpawn test timed out waiting for a child PID; that test does not import the changed provider leaf. Retrying the failed/canceled checks on the same commit.

PR #6704, opened 2026-09-09
fix: read embedding models from public status (#6703)

### LINKED ISSUE #6703: [structural-drift-embeddings-installed-models] Read installed embedding choices from the public models field
Audit slice: local LLM status consumers, including Models > Embeddings. Audited 2026-09-09.

## Problem
client/src/components/settings/EmbeddingsTab.jsx:38-44 reads status.ollama.installedModels and status.lmstudio.installedModels. The producer server/services/localLlm.js:717-718 and :733-734 returns models, never installedModels. This client maintains an incompatible copy of the response contract.
The tab is reachable through client/src/pages/Models.jsx:57. Its candidateModels calculation at EmbeddingsTab.jsx:60-71 therefore receives no installed model IDs, and the installed-model-dependent warning at :170 is also disabled.

## Evidence and impact
A read-only Node VM evaluation of the actual setModels projection supplied:
- ollama.models = [{ id: 'example-embed:latest' }]
- lmstudio.models = [{ id: 'example/embedding-model' }]
It returned { ollama: [], lmstudio: [] }. These are synthetic values; no daemon or live data was read.

Installed embedding models outside KNOWN_768_DIM never appear as datalist suggestions, and Refresh cannot discover them. The user must type their IDs manually. Shipped suggestions remain visible, masking the failure; this does not imply all embedding setup is blocked or that model dimensions have been validated.

History: git log -S'installedModels' -- client/src/components/settings/EmbeddingsTab.jsx traces the incorrect consumer field to eb3604e4f. The service's models: ollamaModels contract traces to 47d67e406, and current bodies still disagree. No EmbeddingsTab component test exists; client/src/pages/Models.test.jsx:19 mocks the entire tab.

Producer: localLlm.getStatus. Transport: server/routes/localLlm.js /status and client/src/services/apiLocalLlm.js getLocalLlmStatus. Affected consumer: EmbeddingsTab.refreshModels -> candidateModels -> embeddings-model-options. Other consumers (useLocalModels, useReviewerModelOptions, useLocalLlmStatus, LocalLlmPlayground) already read models.

## Decided fix
Change the two accesses in client/src/components/settings/EmbeddingsTab.jsx to the existing models API field. Do not introduce an installedModels alias on the server. Add a rendered component regression in client/src/components/settings/EmbeddingsTab.test.jsx using the actual public response shape for both backends, with an installed embedding ID outside the shipped suggestions.
Keep known model suggestions, deduplication, manual entry, provider switching, and the existing 768-dimensional storage contract unchanged. There is no persisted-format change or migration obligation.

## Acceptance criteria
- [ ] An installed embedding model outside KNOWN_768_DIM appears in the selected backend's datalist.
- [ ] Refresh updates the choices from models for both Ollama and LM Studio.
- [ ] Empty model arrays still retain shipped suggestions and manual entry.
- [ ] A focused rendered test fails against the current installedModels accesses and passes after the fix.

Scope: small.
Dispatch: model:light because the fix is two known field accesses plus a self-contained regression; effort:low because the producer contract and failure are unambiguous.

atomantic:
Use the public local LLM status `models` arrays when populating embedding choices, and add rendered regression coverage for both backends, refresh behavior, empty results, known suggestions, and manual model entry.

Tests: `npm test -- --run src/components/settings/EmbeddingsTab.test.jsx` (passed); `npx biome lint --error-on-warnings src/components/settings/EmbeddingsTab.jsx src/components/settings/EmbeddingsTab.test.jsx` (passed). The regression test fails against the former `installedModels` reads (exit 1) and passes with the `models` fix.

Closes #6703

### DIFF
--- a/client/src/components/settings/EmbeddingsTab.jsx
+++ b/client/src/components/settings/EmbeddingsTab.jsx
@@ -39,8 +39,8 @@ export default function EmbeddingsTab() {
     setModelsLoading(true);
     const status = await getLocalLlmStatus({ silent: true }).catch(() => null);
     setModels({
-      ollama: (status?.ollama?.installedModels || []).map((m) => m.id || m.name),
-      lmstudio: (status?.lmstudio?.installedModels || []).map((m) => m.id || m.name),
+      ollama: (status?.ollama?.models || []).map((m) => m.id || m.name),
+      lmstudio: (status?.lmstudio?.models || []).map((m) => m.id || m.name),
     });
     setModelsLoading(false);
   };
--- a/client/src/components/settings/EmbeddingsTab.test.jsx
+++ b/client/src/components/settings/EmbeddingsTab.test.jsx
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from 'vitest';
-import { render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { MemoryRouter } from 'react-router';
 
 vi.mock('../../services/apiSystem', () => ({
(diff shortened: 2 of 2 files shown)


atomantic:
Optional local Claude review was inconclusive because the reviewer produced no verdict. The focused client test could not start in this fresh worktree because dependencies are not installed (`vitest: command not found`).

PR #6772, opened 2026-09-09
fix: identify Grok Build providers as grok-4.6

atomantic:
## Summary
Grok Build providers stored `grok-configured-default`, which leaked into planner labels instead of identifying the model. Pin the shipped CLI/TUI providers to `grok-4.6`, matching the default reported by `grok models`, and migrate legacy placeholder fields while preserving custom selections and argv. Keep legacy sentinel support for compatibility.

## Test plan
- Migration persistence, custom-selection preservation, idempotence, and concrete CLI/TUI model invocation.
- Grok, CLI argument, and TUI handshake suites: 330 tests passed.
- Verified the live CLI reports `grok-4.6`; updated both installed provider records and renamed the existing GitHub planner label to preserve issue associations.


atomantic:
Merge is blocked by a reproducible failure outside this change in Windows CI. Both attempts of run 34418019083 failed `server/lib/detachedSpawn.test.js` → `streams stdout and stderr, then closes with the exit code`, reporting `detached spawn produced no PID within 10000ms` from `detachedSpawn.js:574`. The workflow then canceled the other jobs, leaving CI Gate and Full CI Gate red.

The Grok migration and CLI/TUI suites pass locally (330 tests). No detached-spawn implementation was changed, and fetching main found no newer fix for those files. A full CI rerun reproduced the failure; resolving it requires Windows process-launch diagnosis, which this macOS environment cannot validate. Leaving this PR open rather than bypassing the required checks.

The requested live corrections are complete: both Grok Build providers now select `grok-4.6`, and all 12 affected issues were read back with `planner:grok-4.6` after renaming the existing label.

PR #6786, opened 2026-09-10
refactor: name scheduled prompt compatibility cases

### LINKED ISSUE #6765: [cognitive-load-prompt-upgrade-cases] Name the stored-prompt upgrade cases instead of a four-flag nest
Slice audited: CoS scheduled-task cadence (taskSchedule + Schedule tab + Workflow timeline). Date: 2026-09-09.

## Problem

A stored scheduled-task prompt is four fields that only mean something together: `prompt`, `promptVersion`, `promptCustomized`, `promptSource`. The reader who changes one assignment in `readSchedule` (`server/services/taskScheduleStore.js:255-333`) or in `updateTaskInterval` (`server/services/taskSchedule.js:205-223`) must hold this table, which is currently a ~80-line comment compensating for the code:

| Situation | `promptCustomized` | `promptSource` | `promptVersion` | Next `PROMPT_VERSIONS` bump |
|---|---|---|---|---|
| No prompt stored | (cleared) | forced `null` | current | installs the default |
| Body === current default, version missing | false | — | stamped current | no rewrite |
| Body matches a *retired* shipped hash, version missing | false | — | **reset to 1** | upgrades |
| Body matches no shipped hash, version missing | true | `'legacy-inferred'` | current | **must not** overwrite |
| `promptCustomized` + shipped hash + source ≠ `'user'` | cleared | — | 1 if body ≠ current | self-heal, then upgrade |
| `promptCustomized` + shipped hash + `promptSource === 'user'` | **kept** | `'user'` | kept | **must not** overwrite (#5432 paste-of-retired-default) |
| Uncustomized and `storedVersion < PROMPT_VERSIONS` | false | — | bumped | body replaced with current default |
| Write of current default through Settings | false | `null` | — | not a pin (`taskSchedule.js:214-218`: the editor prefills the textarea) |
| Write of any other body (including a retired shipped one) | true | `'user'` | — | pinned |

That is the distribution-model auto-upgrade contract. It cannot be stated in one sentence today, and the comments at `taskScheduleStore.js:290-315` and `taskSchedule.js:205-218` exist *because* the field names do not carry it.

## Impact

Recurring maintenance with a data-loss class of failure, already paid for:

- #5432 (closed): a user-pasted retired default was indistinguishable from a legacy mis-flag, so the next bump overwrote chosen text. That is why `promptSource` exists.
- AGENTS.md (Prompt templates): six migrations re-implemented this loop against the wrong path (`data/task-schedule.json` instead of `data/cos/task-schedule.json`) before the store's `loadSchedule` was documented as the only upgrade site. `scripts/migrations/promptBumpMigrations.guard.test.js` exists so a seventh copy cannot ship.
- `git log` on `taskScheduleStore.js`: frozen-prompt unstick (`83945be30`, `3dc8d6dd0`), provenance (`e1c729f0c` / #5432), then hash-not-bodies (`4c9406a30` / #6480`). Each was a one-line-looking assignment that was wrong without the table.

The next `PROMPT_VERSIONS` bump author (a routine PortOS change) still has to reconstruct the table from comments. A wrong `promptCustomized = true` freezes every install of that type off upgrades; a wrong self-heal overwrites a user pin.

## Fix (decided)

Keep the four persisted fields (they are the on-disk contract). Extract the decision so a bump author reads a named case, not a nested if:

1. `reconcileStoredPrompt(config, taskType)` in `server/services/taskPromptDefaults/` (the data leaf already owns `promptMatchesShippedDefault` / `PROMPT_VERSIONS` / `DEFAULT_TASK_PROMPTS`). Returns `{ ...fields, changed }`. `readSchedule`'s loop calls it and ORs `changed` into `needsSave`.
2. `stampPromptWrite(prompt, taskType)` for the write side in `updateTaskInterval` — empty → null; current default → `{ promptCustomized: false, promptSource: null }`; anything else → `{ promptCustomized: true, promptSource: 'user' }`.
3. Replace the compensating paragraphs with a one-sentence contract on each helper: *reconcile* "non-`'user'` stored prompts that still match a shipped default upgrade; `'user'` pins never do"; *stamp* "a Settings save pins only when the body is not the current default."
4. Keep the existing `loadSchedule` / `updateTaskInterval` tests as the boundary (they already enumerate #5432). Add a table-driven unit test on `reconcileStoredPrompt` that *is* the table above, so a future field edit fails on the case, not on a comment.

Do not add a fifth flag. Do not move retired bodies back into the tree (#6480). Do not add another prompt-bump migration that writes the schedule file (the guard test forbids it).

## Acceptance criteria

- [ ] `readSchedule` no longer inlines the upgrade/self-heal/version-inference branches; it calls `reconcileStoredPrompt`.
- [ ] `updateTaskInterval` no longer inlines the pin rule; it calls `stampPromptWrite`.
- [ ] Every row of the table above is a named example in the new unit test (including #5432 user-pin of a retired hash, and "re-save of current default does not pin").
- [ ] Existing `taskSchedule.test.js` promptSource / self-heal / auto-upgrade cases still pass with no assertion changes.
- [ ] A `PROMPT_VERSIONS` bump plus `node scripts/regen-prompt-integrity-snapshot.js` still auto-upgrades an uncustomized install and still leaves a `'user'` pin alone.

Scope: medium

model:heavy — this is the distribution-model compatibility contract; a wrong case overwrites user prompts or freezes upgrades across independently-updated installs. effort:medium — once the table is the test, the extract is localized to the store, `updateTaskInterval`, and the new helper. (Not both-medium: capability is the hard axis, budget is ordinary.)

atomantic:
## Summary

Scheduled prompt upgrades now use named reconciliation cases and a shared Settings-write rule. Retired defaults still upgrade, inferred customizations still self-heal, and explicit user pins remain protected. The four persisted fields and shipped prompt bodies are unchanged.

Added named compatibility examples for legacy inference, retired hashes, user pins, empty writes, and re-saving the current default. The extraction preserves existing legacy edge behavior rather than treating the issue's abbreviated table as a storage migration.

## Test plan

- 352 tests passed across `taskSchedule.test.js`, `taskPromptDefaults.test.js`, `taskPromptDefaults/storedPrompt.test.js`, and `promptBumpMigrations.guard.test.js`.
- Existing boundary assertions unchanged; new cases also check input immutability and stable repeat reconciliation.

Closes #6765

### DIFF
--- a/server/services/taskPromptDefaults.js
+++ b/server/services/taskPromptDefaults.js
@@ -30,3 +30,4 @@
 export { DEFAULT_TASK_PROMPTS } from './taskPromptDefaults/prompts.js';
 export { PROMPT_VERSIONS, REFERENCE_WATCH_AUDITED_VERSION } from './taskPromptDefaults/versions.js';
 export { promptMatchesShippedDefault } from './taskPromptDefaults/shippedPrompts.js';
+export { reconcileStoredPrompt, stampPromptWrite } from './taskPromptDefaults/storedPrompt.js';
--- a/server/services/taskPromptDefaults/storedPrompt.js
+++ b/server/services/taskPromptDefaults/storedPrompt.js
@@ -1,0 +1,59 @@
+import { DEFAULT_TASK_PROMPTS } from './prompts.js';
+import { PROMPT_VERSIONS } from './versions.js';
+import { promptMatchesShippedDefault } from './shippedPrompts.js';
+
+/** Reconcile shipped defaults and inferred customizations without overwriting explicit user pins. */
+export function reconcileStoredPrompt(config, taskType) {
+  const currentPrompt = DEFAULT_TASK_PROMPTS[taskType];
+  const currentVersion = PROMPT_VERSIONS[taskType] || 1;
+  const fields = Object.fromEntries(
+    ['prompt', 'promptVersion', 'promptCustomized', 'promptSource']
+      .filter(key => Object.hasOwn(config, key))
+      .map(key => [key, config[key]])
+  );
+  const finish = () => ({
+    ...fields,
+    changed: Object.keys(fields).some(key => fields[key] !== config[key])
+  });
+
+  // Missing body: install the current default and clear stale provenance.
+  if (!fields.prompt && currentPrompt) {
+    fields.prompt = currentPrompt;
+    fields.promptVersion = currentVersion;
+    if (fields.promptSource) fields.promptSource = null;
+    return finish();
+  }
+
+  const isShippedDefault = promptMatchesShippedDefault(fields.prompt, taskType);
+  const isCurrentDefault = fields.prompt === currentPrompt;
+
+  // Legacy unversioned body: recognize defaults before inferring a customization.
+  if (fields.prompt && fields.promptVersion === undefined && currentPrompt) {
+    fields.promptVersion = isCurrentDefault || !isShippedDefault ? currentVersion : 1;
+    if (!isShippedDefault) {
+      fields.promptCustomized = true;
+      fields.promptSource = 'legacy-inferred';
+    }
+  }
+
+  // Inferred pin of a shipped body: repair both the flag and a misleading version.
+  if (fields.promptSource !== 'user' && fields.promptCustomized && isShippedDefault) {
+    fields.promptCustomized = false;
+    if (!isCurrentDefault) fields.promptVersion = 1;
+  }
+
+  // Unpinned older version: replace the body with this release's default.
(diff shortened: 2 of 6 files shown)


atomantic:
The serialized merge queue left this PR open because required CI Gate failed. The server suite reports services/taskScheduleModules.test.js, taskSchedule module boundaries > keeps prompt-version migration data in the persisted store seam: storeSource is expected to contain PROMPT_VERSIONS and promptMatchesShippedDefault, while this extraction delegates to reconcileStoredPrompt. The failure is reproducible from the changed module boundary, so no flaky-check rerun was used. The branch, worktree, assignee, and in-progress marker remain intact for follow-up.

PR #6788, opened 2026-09-10
fix: recover share-bucket export cache writes (#6777)

### LINKED ISSUE #6777: [runtime-safety-exporter-cache-tail-poison] One failed blob-index write permanently breaks later exports to that bucket
Audit slice: peer federation / share-bucket (`server/services/sharing/exporter.js`). Date: 2026-09-09.

## Problem

`withAssetHashCache` serializes blob-index writes per bucket path, but the tail is a raw `.then(successOnly)` with no rejection recovery:

```103:118:server/services/sharing/exporter.js
const cacheWriteTails = new Map();

async function withAssetHashCache(bucketPath, fn) {
  const cache = await loadAssetHashCache(bucketPath);
  const initialKeys = Object.keys(cache).length;
  const result = await fn(cache);
  if (Object.keys(cache).length !== initialKeys) {
    const prevTail = cacheWriteTails.get(bucketPath) || Promise.resolve();
    const tail = prevTail.then(async () => {
      const onDisk = await loadAssetHashCache(bucketPath);
      await atomicWrite(bucketBlobIndexPath(bucketPath), { ...onDisk, ...cache });
    });
    cacheWriteTails.set(bucketPath, tail);
    await tail;
  }
  return result;
}
```

`prevTail.then(async () => { ... })` does **not** run the writer if `prevTail` is rejected — the new promise rejects immediately with the old reason. The rejected promise stays in `cacheWriteTails` for the process lifetime.

The rest of the tree already documents this trap. `createFileWriteQueue` (`server/lib/fileWriteQueue.js:33-36`) chains `tail.then(fn, fn)` and silences the stored tail specifically so "a rejection from poisoning subsequent waiters." `peerSyncShared.js:82` and `dashboardLayouts.js:431-435` do the same. This exporter tail copied the Map and not the recovery.

## Impact

**Runtime — a transient blob-index write failure permanently fails later exports to that bucket until PM2 restart.**

Share buckets are meant to live on Drive/Dropbox folders (`watcher.js` is built around cloud-sync delivering files). Those clients lock files during upload. Sequence:

1. `exportSeries` / `exportUniverse` copies assets (that part already finished — `fn(cache)` runs before the index write).
2. `atomicWrite` of the blob index throws (file locked, ENOSPC, EACCES). `await tail` rejects, the export errors.
3. The user retries, or a later recordEvents debounce re-exports. `cacheWriteTails.get(bucketPath)` is still the rejected promise.
4. The new tail skips the writer entirely and rejects with the **original** error, even though Drive has released the file.

So one transient lock turns into "every subsequent share to this bucket fails" until the server process restarts. Asset copies may already be on disk; only the index write is poisoned, but the thrown error fails the whole export (subscription `lastManifestId` is not stamped, inbox/share UI reports failure).

## Fix

Delete the hand-rolled `cacheWriteTails` Map. Drive the index write through `createKeyCachedQueue()` from `server/lib/createKeyCachedQueue.js`, keyed by `bucketPath`. That helper already:

- runs `work` on both fulfil **and** reject of the prior link,
- self-prunes the key when the tail settles, so a rejected write does not stick around.

Keep the reload-and-merge (`loadAssetHashCache` then `{ ...onDisk, ...cache }`) inside the queued work so two parallel `exportSeries` against the same bucket still accumulate entries.

Add a unit test that stubs the first `atomicWrite` of the blob index to reject, then asserts a second `exportSeries` (or a direct `withAssetHashCache` via the export path) still persists the new cache keys. There is currently no exporter test covering this tail.

`model:light` — single function, existing helper is the drop-in. `effort:low` — the recovery pattern is already written down next to the queue helper; the test is one reject-then-retry case.

Scope: small.

## Acceptance criteria

- [ ] `withAssetHashCache` no longer stores a rejected promise as the per-bucket tail; a failed index write does not prevent the next export from writing.
- [ ] Concurrent exports to the same bucket still merge cache entries (reload-and-spread preserved).
- [ ] A test: first blob-index `atomicWrite` rejects, second export still writes the index and resolves.
- [ ] Existing exporter/subscription export tests still pass.

atomantic:
## Summary
- Replace the exporter blob-index tail map with the shared self-pruning keyed queue.
- Preserve per-bucket reload-and-merge behavior for concurrent exports.
- Add a transient index-write failure regression test proving the next export succeeds.

## Tests
- `server/node_modules/.bin/vitest run lib/createKeyCachedQueue.test.js services/sharing/integration.test.js` (66 passed)
- `git diff --check`

Closes #6777

### DIFF
--- a/server/services/sharing/exporter.js
+++ b/server/services/sharing/exporter.js
@@ -19,6 +19,7 @@ import { join, basename } from 'path';
 import { readFile, stat } from 'fs/promises';
 import { existsSync } from 'fs';
 import { PATHS, ensureDir, atomicWrite, readJSONFile, sha256File, copyFileGuarded } from '../../lib/fileUtils.js';
+import { createKeyCachedQueue } from '../../lib/createKeyCachedQueue.js';
 import { getOrComputeImageSha256 } from '../../lib/assetHash.js';
 import { isPlainObject } from '../../lib/objects.js';
 import { getBucket, ensureBucketLayout, bucketBlobsDir, bucketBlobPath, bucketBlobSidecarPath, bucketBlobIndexPath, bucketRecordsDir, bucketRecordPath, imageSidecarName, isHexHash } from './buckets.js';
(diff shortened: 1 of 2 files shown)


atomantic:

Issue #6831, opened 2026-09-10
videoGen/fal.js's post-finalize catch never forces past the idempotency guard (unlike grok/reactor)


atomantic:
## Problem

`videoGen/fal.js`'s `runFalVideo` has the same latent gap that #6830 just fixed for the three image backends, but it was left out of that fix's scope (it only forced the three image post-exit catches).

`finalizeGeneratedVideo` (`server/services/videoGen/generateVideoHelpers.js:748`) stamps `job.status = 'complete'` BEFORE its own async tail (`optimizeForStreaming`, `generateThumbnail`, the serialized history write). `runFalVideo`'s success path already clears `activeRequests`/`activeJobs` and then awaits `finalizeGeneratedVideo` (`server/services/videoGen/fal.js:267-269`) inside the function's outer `try`. If that await throws — e.g. `optimizeForStreaming`/`generateThumbnail` failing, or the history write rejecting — the throw lands in the catch at `fal.js:271-276`, which calls the shared finalizer WITHOUT `force: true`:

```js
} catch (err) {
  // NOTE: not forced, unlike videoGen/grok.js's post-exit catch — see #6831.
  finalizeJobFailure(job, jobId, null, `fal.ai video generation failed: ${err?.message || err}`);
}
```

Because `job.status` is already `'complete'` by that point, the finalizer's idempotency guard makes this a no-op: no `failed` event, no SSE error frame, no `closeJobAfterDelay`. The job sits reading `'complete'` with no terminal frame ever delivered to the client or the media queue — the exact window `videoGen/grok.js`'s post-exit catch already forces past (fa3796650), and the one #6830 fixed for `imageGen/{agy,codex,grok}.js`.

`videoGen/reactor.js` does not have this gap — its single catch-all always passes `{ force: true }` already.

## Fix (decided)

Add `{ force: true }` to the one catch-all call in `runFalVideo` (`server/services/videoGen/fal.js`, currently around line 276), matching `reactor.js`'s catch and the comment already in `videoGen/grok.js`:

```js
} catch (err) {
  finalizeJobFailure(job, jobId, null, `fal.ai video generation failed: ${err?.message || err}`, { force: true });
}
```

Update the `NOTE:` comment above that call site (added by #6830) to reflect the fix instead of the gap.

## Test plan

[...]

PR #6832, opened 2026-09-10
Share one job-failure finalizer across the six media backends

### LINKED ISSUE #6830: [simplify-media-backend-finalize-error] Share one job failure finalizer across the six media backends
**Slice audited:** media generation backends (`server/services/imageGen/{agy,codex,grok}.js`, `server/services/videoGen/{grok,fal,reactor}.js`) — dead-code and duplication audit, 2026-09-10.

## Problem

Six backends each define a private `finalizeError` that does the same eight things — idempotency guard, clear the active process or request slot, stamp `job.status = 'error'`, drop from `activeJobs`, log, `broadcastSse` an `error` frame, emit `failed`, `closeJobAfterDelay`:

- `imageGen/agy.js:410-419`, `imageGen/codex.js:462-474`, `imageGen/grok.js:442-452`
- `videoGen/grok.js:312-324`, `videoGen/fal.js:278-287`, `videoGen/reactor.js:334-343`

They have drifted. Review fix fa3796650 (2026-07-22, "forced terminal on finalize throw") added a `{ force }` option to `videoGen/grok.js` so a post-exit handler that throws AFTER stamping `complete` but BEFORE emitting `completed` still emits a terminal `failed` (`videoGen/grok.js:298` passes `force: true`). The two video backends added in September copied that variant (`fal.js:276`, `reactor.js:334`). The three image backends never got it: their post-exit catches at `imageGen/grok.js:434`, `imageGen/agy.js:405` and `imageGen/codex.js:453` call the un-forced finalizer, which is a no-op once `job.status === 'complete'` — and `imageGen/grok.js:424` stamps `complete` before the `broadcastSse` and `imageGenEvents.emit('completed')` calls at `:429-430`.

## Impact

Copy-paste drift with a latent gap: an image render whose completion handler throws inside that window — a subscriber's `res.write` failing inside `broadcastSse`, or any throw the catch at `imageGen/grok.js:433` was written to contain — leaves the job stamped `complete` with neither a `completed` nor a `failed` event, so the media job queue and the SSE client never see a terminal frame for it. The video side proved the window is real; that is why fa3796650 exists. Beyond the gap, every finalizer change is a six-file edit, and the September backends show new copies are still being minted instead of shared.

## Fix (decided)

1. Add `createJobFailureFinalizer({ jobs, activeJobs, activeSlots, label, events, failedPayload })` to `server/lib/sseUtils.js` beside `broadcastSse` and `closeJobAfterDelay`. It returns `(job, jobId, slotOwner, reason, { force } = {})` implementing the shared body: return when `!force` and the job is already terminal; delete the slot when `slotOwner == null || activeSlots.get(jobId) === slotOwner`; stamp `error`; `activeJobs.delete`; log with `label`; `broadcastSse` the error frame; `events.emit('failed', failedPayload(jobId, reason))`; `closeJobAfterDelay`.
2. Replace the six private definitions with the factory: `activeSlots` is `activeProcs` for the CLI backends and `activeRequests` for fal and reactor; `failedPayload` supplies the `mode` field the image event bus carries.
3. Pass `{ force: true }` from the three image post-exit catches, matching `videoGen/grok.js:298`, so the image path gets the fix.
4. Tests: one `sseUtils.test.js` block covering idempotency, `force` and the stale-owner guard; one regression in `imageGen/grok.test.js` where `broadcastSse` throws after `complete` and a `failed` event is still emitted.

## Acceptance criteria

- [ ] `grep -rn "^const finalizeError" server/services/imageGen server/services/videoGen` returns nothing; all six backends use the shared factory.
- [ ] The three image backends emit `failed` when their post-exit handler throws after `job.status = 'complete'` (regression test).
- [ ] Existing backend tests under `imageGen/` and `videoGen/` pass without changing their assertions on event payloads.
- [ ] `server/lib/README.md` documents the factory and `server/lib/index.test.js` passes.

Scope: medium.

Dispatch: `model:medium` + `effort:medium` — six call sites with three small variation axes (slot map, event bus, force); routine, but each site needs its axis chosen deliberately rather than pattern-replaced.

atomantic:
## Summary

- Add `createJobFailureFinalizer` to `server/lib/sseUtils.js` — the shared "this job died" finalizer (idempotency guard with a `force` escape hatch, slot-owner clear, SSE error frame, `failed` event, `closeJobAfterDelay`) behind all six media-generation backends.
- Migrate `imageGen/{agy,codex,grok}.js` and `videoGen/{grok,fal,reactor}.js` off their private, drifted `finalizeError` copies onto the shared factory.
- Fix the drift the audit found: the three image backends never got the `force` option the video backends picked up in fa3796650, so a throw landing after `job.status` is already stamped `'complete'` (the success-path SSE/event tail failing) was silently swallowed as a no-op instead of delivering a terminal `failed` event. Now all three image post-exit catches pass `{ force: true }`, matching `videoGen/grok.js`.
- Document the factory in `server/lib/README.md`.
- `videoGen/fal.js` has one remaining catch-all call site with this same latent gap; left unforced here since the issue's decided fix named only the three image catches — filed #6831 to close it separately.

## Test plan

- `sseUtils.test.js` — new coverage for the factory: default idempotency, `force` override, stale-slot-owner guard (a late finalize from a superseded run doesn't evict a newer job's slot), unconditional clear when `slotOwner` is `null`, and a custom `failedPayload`.
- `imageGen/grok.test.js` — new regression test: a subscriber's `res.write` throwing on the terminal `complete` SSE frame (after `job.status` is already `'complete'`) still emits `failed`. Verified this fails against the pre-fix code in a clean detached worktree at the base commit, and passes with the fix.
- Full existing suites for all six backends plus `sseUtils.test.js` and `lib/index.test.js` pass unchanged (165 tests, including two new `.canceled` unit tests added during self-review).

Closes #6830


atomantic:
Orchestrator merge review (swarm run, reviewer `opencode~opt~max=1` re-run tool-free at effort high on the full branch diff): 4 findings, none actionable here.

- `force` also bypassing the `error` guard, the owner-checked slot delete still stamping `status='error'`, and `closeJobAfterDelay` being skipped if the SSE broadcast or `failed` emit throws — all three are the pre-existing finalizer semantics from `videoGen/grok.js` (fa3796650) and the image backends, reproduced verbatim by the factory. This PR's contract is behavior-preserving apart from the deliberate `force` fix, so they stay as they were. The stale-owner case is also unreachable today: every backend `set`s its slot exactly once per job id.
- The un-forced catch-all in `videoGen/fal.js` is tracked in #6831.

Also verified the new `imageGen/grok.test.js` regression fails at the base commit (1 failed / 29 passed) and passes on the branch.

PR #6840, opened 2026-09-10
refactor: move description fold ownership to cosTaskStore (#6838)

### LINKED ISSUE #6838: [module-hygiene-description-fold-owner] Let the task store own the multi-line description fold the generator repeats twice
**Slice audited:** server service hubs ranked by import reach × 90-day churn; this item is the follow-up the 2026-09-08 module-hygiene run recorded against `server/services/cosTaskGenerator.js` (the CoS task intake seam, `cosTaskStore.js` ↔ `cosTaskGenerator.js`), verified 2026-09-10.

## Problem

The rule "a multi-line task description becomes `metadata.prompt` and the description collapses to its first line" (#4153) has an owner: `addTask` in `server/services/cosTaskStore.js:356-388` applies it on **both** the raw and the constructed branch ("the raw path is how the generator … queue their pre-built tasks", `cosTaskStore.js:357-359`) and encodes the one precedence rule that matters — a producer-written `metadata.prompt` wins over the inference (`cosTaskStore.js:382-384`).

`cosTaskGenerator.js` re-implements the same fold twice, immediately before handing the task to that store:

- `prepareQueuedImprovementTask` at `cosTaskGenerator.js:1376-1385`, called at `:1424` right before `persistTask(task, 'internal', { raw: true, … })` at `:1426`;
- an inline copy at `cosTaskGenerator.js:1588-1592`, right before `addTask(task, 'internal', { raw: true, … })` at `:1594`, under a 25-line comment that ends: "Keep the queue-path normalization visible here as well as in the install-wide helper … the queue contract is source-checked by the scheduler tests."

That last sentence is the whole reason the copy exists: `server/services/cos.test.js:1671-1677` reads the generator function's **source text** and asserts it contains `metadata.prompt = <x>.description` and `.description = firstLine(`. The test pins the *text* of a duplicate rather than the *behavior* of the owner, so the duplicate cannot be deleted without the test failing, and the test would still pass if the store's fold were broken.

The two copies also differ from the owner on precedence — the generator overwrites `metadata.prompt` unconditionally, the store preserves a producer prompt. No generator path sets a `metadata.prompt` before either copy today (`grep -n '^\s*prompt:' cosTaskGenerator.js` is empty), so the difference is latent, not live — but it is the kind of fork the #4153 contract was written to prevent.

## Impact

- Three implementations of one persisted-shape rule; a change to the fold (e.g. trimming trailing blank lines, a different first-line rule) must land in three places, and `cos.test.js` will keep passing when only the generator's copy is updated.
- A source-grep test on a 3,000-line module is the vacuous-assertion pattern the repo's test strategy (`AGENTS.md`, "Test Strategy") tells reviewers to remove: it mirrors implementation text and proves nothing about the persisted record.

## Reuse search

- `server/lib/cosTaskPrompt.js` (`getTaskPrompt`, `splitTaskPromptFields`) is the read-side contract; `cosTaskStore.js#addTask` is the write-side owner; `cosTaskIntake.js:20-21` documents that "the multi-line description → `metadata.prompt` classification stay[s] in `cosTaskStore.js`". Nothing else needs a fold helper — the fix is deletion, not a new shared function.
- Prior work: #4153 (the prompt/context split and the store-side fold), #6604 (moved the request→metadata mapping into `cosTaskIntake.js` and replaced the generator's source-grep guards with a parity test — the same conversion this item asks for on the remaining grep).

## Fix (decided)

1. Delete `prepareQueuedImprovementTask` (`cosTaskGenerator.js:1376-1385`, its call at `:1424`) and the inline fold at `:1588-1592` plus its comment block; the `firstLine` import at `:43` goes if nothing else uses it. The store's fold on the raw path is what those two `addTask`/`persistTask` calls already receive.
2. Replace the two source-text assertions in `cos.test.js:1671-1677` with a boundary test at the store: `addTask({ …, description: 'headline\nbody line 2' , metadata: {} }, 'internal', { raw: true })` persists `description === 'headline'` and `metadata.prompt` equal to the full body, and a task that already carries `metadata.prompt` keeps it (the producer-wins rule). Put it in `cosTaskStore.test.js` next to the existing `firstLine` cases (`cosTaskStore.test.js:148-157`) if no equivalent case exists there yet; keep the neighbouring `getTaskDescription` assertion in `cos.test.js` untouched.
3. Confirm the on-demand dispatch path (which emits the in-memory task on `task:ready` without persisting) still renders the full body: `getTaskPrompt` falls back to the description, and `agentPromptBuilder.js` renders `task.description` in full — no fold is needed there.

**Scope:** small

## Acceptance criteria

- [ ] `grep -n 'metadata.prompt = \|firstLine(task.description)' server/services/cosTaskGenerator.js` returns nothing; `prepareQueuedImprovementTask` no longer exists.
- [ ] `cos.test.js` contains no `toMatch(/metadata\.prompt` or `toMatch(/\.description\s*=\s*firstLine` source assertions; a store-level test proves the raw-path fold and the producer-prompt precedence, and fails when the fold in `cosTaskStore.js:377-388` is removed.
- [ ] `cosTaskGenerator.test.js` and `cos.test.js` still pass; a generated install-wide task queued through the real `addTask` reads back from `COS-TASKS.md` with a single-line description and the full prompt in `metadata.prompt`.
- [ ] `cd server && npm test` green.

Dispatch: `model:light` (two deletions and one test rewrite at a known boundary) / `effort:medium` (the replacement test must be mutation-checked against the store's fold, and the on-demand path confirmed).

atomantic:
## Summary

- Remove two duplicate implementations of the multi-line description fold from `cosTaskGenerator.js` — the store owns this rule (#4153)
- Replace vacuous source-text grep assertions in `cos.test.js` with a behavioral boundary test at the store
- Update generator test to expect unfolded description (fold now happens in store, not generator)

## Test plan

- `cosTaskStore.test.js`: 170 tests pass (added 1 new boundary test for producer-prompt-wins rule)
- `cosTaskGenerator.test.js`: 154 tests pass (updated expectations to match new fold ownership)
- `cos.test.js`: 151 tests pass (removed source grep, kept getTaskDescription assertion)
- Full server suite: 2133 test files passed, 42721 tests passed
- Mutation check: new test fails when fold disabled in cosTaskStore.js:376, passes when restored
- OpenCode review: clean, no findings, 65227ms

## Remaining scope

None.

Closes #6838

### DIFF
--- a/server/services/cos.test.js
+++ b/server/services/cos.test.js
@@ -1658,24 +1658,6 @@ describe('cos.js source — priority + capacity invariants', () => {
       'queue path must NOT use getTaskDescription (one-line stub bypasses prompt enrichment)'
     ).not.toMatch(/getTaskDescription\s*\(/);
 
-    // The generator returns a multi-line `description` (the full Phase 1–7
-    // prompt template). COS-TASKS.md serialization interpolates the whole
-    // description onto a single `- [ ]` line and the parser only matches the
-    // first line, so persisting a multi-line description corrupts the file
-    // AND truncates the prompt on the next `dequeueNextTask` re-read. The
-    // queue path must move the body to `metadata.prompt` (which IS
-    // newline-escaped) so the agent prompt builder reconstitutes it on
-    // dispatch. `prompt`, not `context` — the two were split in #4153 so a
-    // multi-thousand-character agent payload is distinguishable from the
-    // one-line human note. Pin both halves of the split.
-    expect(
-      fnBody,
-      'queue path must move multi-line description body to metadata.prompt (survives markdown round-trip)'
-    ).toMatch(/metadata\.prompt\s*=\s*\w+\.description/);
-    expect(
-      fnBody,
-      'queue path must collapse description to a single line via firstLine()'
-    ).toMatch(/\.description\s*=\s*firstLine\(/);
 
     // appActivity helpers must come from the file-level static import (line ~23),
     // NOT a dynamic `await import('./appActivity.js')` *inside* the per-app
--- a/server/services/cosTaskGenerator.js
+++ b/server/services/cosTaskGenerator.js
@@ -40,7 +40,7 @@ import { remainingActionBudget } from '../lib/domainBudgets.js';
 import { getDomainBudgetStatus } from './domainUsage.js';
 import { pendingCosActionReservations } from './cosAdmissionReservations.js';
 import { cosEvents, emitLog } from './cosEvents.js';
-import { addTask, updateTask, getAllTasks, getCosTasks, firstLine } from './cosTaskStore.js';
+import { addTask, updateTask, getAllTasks, getCosTasks } from './cosTaskStore.js';
 import { PRIORITY_VALUES } from '../lib/taskParser.js';
 import { recordDecision, DECISION_TYPES } from './decisionLog.js';
 import { isAppOnCooldown, markAppReviewCooldown, bindAppReviewAgent, markIdleReviewStarted, getNextAppForReview, loadAppActivity, isAppActivityOnCooldown } from './appActivity.js';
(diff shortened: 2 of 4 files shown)


atomantic:
Orchestrator re-review before merge (tool-free `opencode` pass at effort high on the branch diff, base 358f77750):

- **Finding 1 — pass `newTask` to `recordDeferredPerpetualDispatch`: rejected.** That function uses the task object only as a lookup KEY (`deferredPerpetualSignatures.get(task)`), so it needs the original object identity the generator registered; it never reads the description or prompt. Switching to the store's new object would break the perpetual-dispatch signature lookup.
- **Finding 2 — add an end-to-end fold test through `persistTask`: rejected.** `persistTask` defaults to `addTask` itself, and `cosTaskStore.test.js` already round-trips a raw multi-line description through the markdown file (`moves a raw multiline description into metadata.prompt before markdown persistence`). The mutation probe (fold neutralised in the store) fails four store tests including the new producer-wins case, so the boundary is guarded at its owner.

Suites re-run on the branch: 475 tests across `cosTaskStore`, `cosTaskGenerator`, `cos` pass. Merging.

Issue #6869, opened 2026-09-10
[cognitive-load-pr-ownership-stamp-name] The persisted ownsPrWorkflow stamp means "rendered the inline PR section", not "agent opens its own PR" — and every reader assumes the latter


atomantic:
**Slice audited:** the CoS agent PR-ownership contract — `server/lib/slashdoInvocation.js`, `server/services/promptSections/completion.js`, `server/services/agentLifecycle.js`, `server/lib/agentRegistrationRecord.js`, `server/services/agentTuiSpawning.js`, `server/services/agentCliSpawning.js`, `server/services/agentCompletionCleanup.js`, `server/services/agentWorktreeCleanup.js`, `server/lib/prDisposition.js`, `server/lib/mergeGateContract.js`. Cognitive-load audit, 2026-09-10.

## Problem

One identifier, `ownsPrWorkflow`, is defined three ways, and every layer that reads it assumes a fourth meaning: "the agent opens its own PR".

1. `agentOwnsPrWorkflow({ providerType, leanMode })` (`server/lib/slashdoInvocation.js:514-517`) — "a real coding harness that is not lean". True for a Claude TUI or CLI.
2. The persisted stamp `metadata.ownsPrWorkflow` (`server/services/agentLifecycle.js:645-652`, written by `server/lib/agentRegistrationRecord.js:121`) — `inlinePrLifecycleSection(...) !== null`. That predicate returns `null` whenever `canTypeSlashCommands(...)` is true (`server/services/promptSections/completion.js:894`), so the stamp is **false for every slashdo-capable host** — the Claude TUI and Claude CLI that run `/do:pr` and open, review and merge their own PR. It answers "did the prompt render the manual `git`/`gh` PR section", not "does the agent open its own PR".
3. `resolveOwnsPrWorkflow({ persisted, ... })` (`slashdoInvocation.js:538-541`) — the stamp when present, else `canTypeSlashCommands(...)`. Inside the harness set the fallback is the *complement* of the stamp: a legacy Claude record resolves `true`, a stamped Claude record resolves `false`.

Every consumer reads the name as meaning 4:

- `resolvePrCreation`'s parameter doc says "`agentOwnsPr` — the prompt told the agent to open it" (`server/lib/prDisposition.js:100`) and does `if (!agentOwnsPr) return PR_CREATION.ALWAYS` (`:109`).
[...]

PR #6876, opened 2026-09-10
completion sections: branch on the resolved mode, not re-derived slashdo flags

### LINKED ISSUE #6873: [cognitive-load-slashdo-capability-names] One fact, four names (canRunSlashCommands / canTypeSlash / hasSlashdo / slashdoFree), and the completion sections re-decide a mode the resolver already chose
**Slice audited:** the agent prompt completion contract — `server/services/agentPromptBuilder.js`, `server/services/promptSections/completion.js`, `server/lib/agentCompletionMode.js`. Cognitive-load audit, 2026-09-10. Independent of the PR-ownership stamp finding filed alongside it (that one is about what gets persisted for cleanup; this one is about the prompt builder's own names).

## Problem

Whether the session can type a `/do:*` command is computed once per prompt path — `canRunSlashCommands` on the full path (`server/services/agentPromptBuilder.js:500`), `canTypeSlash` on the light path (`:998`) — and then re-expressed as two more booleans: `hasSlashdo = !isTui && canTypeSlash` (`:1003`) and the negated `tuiSlashdoFree = isTui && !canTypeSlash` (`:1010`), passed on as `slashdoFree` (`:1124`). The full path derives `slashdoFree` two different ways ten lines apart (`:563` `isTui && !canRunSlashCommands`, `:577` `!canRunSlashCommands`), with a seven-line comment (`:568-574`) explaining why one of them is unreachable.

`resolveCompletionMode` (`server/lib/agentCompletionMode.js:111-112`) already decides `TUI_SLASHDO_FREE` vs `TUI` from the same inputs, and its header says "call sites branch on the returned key and never re-test the flags themselves" (`:14-16`). Yet `buildTuiCompletionSection` receives `slashdoFree` and re-tests it (`completion.js:613-620`), and `buildCliCompletionSection` receives `hasSlashdo` and re-tests it against `worktreeInfo`/`willOpenPR` (`:968-1006`) to rediscover `WORKTREE_NO_PUSH` / `PORTOS_MERGES` / `COMMIT_AND_PUSH`, which the light path had already resolved before calling it (`agentPromptBuilder.js:1160-1162`). The mode's doc comment lies for the slashdo CLI as a result: `WORKTREE_NO_PUSH` says "commit only, the system pushes and opens it" (`agentCompletionMode.js:82`) while the section that mode renders tells a slashdo CLI to run `/do:pr` (`completion.js:971-1004`).

**What a reader must hold in their head:** that `slashdoFree === false` means `hasSlashdo` only when `isTui`; that `hasSlashdo` is never true for a TUI; that the `:563` and `:577` derivations differ on purpose; and that the section builders' branch order must mirror the resolver's rule order by hand.

## Impact

This is the drift class #3114 fixed — "three inline provider-id allowlists that had already drifted apart … a codex TUI used to be told to run `/do:pr`, which it cannot" — and `:568-574` records that one arm "previously passed no `slashdoFree` at all, which is how gates like it drift". #6616 centralized the decision but left the section builders re-deriving it from the raw flags, so the next drift lands in the builders, where no snapshot names it.

## Fix (decided)

Pass the mode, not the flags.

- `buildTuiCompletionSection({ mode, ... })` takes `mode` (`TUI` | `TUI_SLASHDO_FREE`) instead of `slashdoFree`; `if (slashdoFree)` at `:616` becomes `if (mode === COMPLETION_MODES.TUI_SLASHDO_FREE)`.
- `buildCliCompletionSection({ mode, canTypeSlashCommands, ... })` branches on `mode` first: `WORKTREE_NO_PUSH` with `canTypeSlashCommands` → the `/do:pr` block (`:971-1004`), `PORTOS_MERGES` with `canTypeSlashCommands` → the auto-merge commit block (`:1006-1017`), then the inline-PR and plain blocks as today.
- `worktreeCommitGuidance` and `buildLightTaskContextSections` take the same `mode` + `canTypeSlashCommands`.
- Delete `hasSlashdo`, `tuiSlashdoFree` and both `slashdoFree` derivations in `agentPromptBuilder.js`; one name, `canTypeSlashCommands`, on both paths.
- Fix the `WORKTREE_NO_PUSH` comment: "commit only; `/do:pr` when the host can type it, otherwise the system pushes and opens it".

The rendered prompt text is unchanged; the byte-for-byte prompt snapshots in `agentPromptBuilder.test.js` and `completion.test.js` are the proof.

Scope: small.

Dispatch: `model:medium` and `effort:medium` — a parameter swap across three files guarded by existing byte-for-byte prompt snapshots; neither axis needs more, and neither is mechanical enough for `light`/`low` because the CLI section's branch order has to be matched to the resolver's rule order.

## Acceptance criteria

- [ ] `slashdoFree`, `tuiSlashdoFree` and `hasSlashdo` no longer exist under `server/`.
- [ ] `buildTuiCompletionSection` and `buildCliCompletionSection` branch on `COMPLETION_MODES` keys, not on capability flags.
- [ ] Prompt snapshot tests pass unchanged.
- [ ] The `WORKTREE_NO_PUSH` comment matches what a slashdo CLI is told.

atomantic:
## Summary

- `buildTuiCompletionSection` and `buildCliCompletionSection` (`server/services/promptSections/completion.js`) now branch on the resolver's `COMPLETION_MODES` key plus one `canTypeSlashCommands` capability flag, instead of each re-deriving its own `hasSlashdo` / `tuiSlashdoFree` / `slashdoFree` boolean.
- `worktreeCommitGuidance` and the light path's `buildLightTaskContextSections` take the same `mode` + `canTypeSlashCommands` pair.
- The light prompt path now resolves `completionMode` once (moved earlier in `buildLightContextSections`) and reuses it, instead of computing it a second time further down.
- Fixed the `WORKTREE_NO_PUSH` mode's doc comment in `server/lib/agentCompletionMode.js` — it said the system always pushes and opens the PR, but the CLI section this mode renders tells a slashdo-capable host to run `/do:pr` itself.
- Removed the `slashdoFree` / `hasSlashdo` / `tuiSlashdoFree` identifiers from code, comments, and test titles (none remain under `server/`).
- `ownsPrWorkflow` is deliberately untouched — a separate sibling issue (#6869) owns renaming/re-deriving it.

## Test plan

- `cd server && ./node_modules/.bin/vitest run services/agentPromptBuilder.test.js lib/agentCompletionMode.test.js lib/slashdoInvocation.test.js services/agentCompletionCleanup.test.js` → 399/399 passed, including every byte-for-byte prompt assertion (no snapshot regeneration needed).
- Also ran the other importers of the touched modules (`cosValidation`, `importScoping`, `reviewerConfig`, `slashdoCatalog`, `agentCliSpawning`, `agentLifecycle`, `agentTuiSpawning`, `cosTaskIntake`, `taskTemplates`, `promptSections/reviewLifecycle`) → 488/488 passed.
- Baseline probe: copied the updated `agentPromptBuilder.test.js` onto the pre-change commit and ran it there — exactly one test failed (`slashdo-free TUI bullet points at the commit + PortOS handoff, not a /do:* command`), confirming the test edit is coupled to the production change rather than vacuous. The other two edited call sites already agreed with `mode` and passed at both revisions.
- Local reviewer (opencode, effort medium): `clean`, 0 findings.

Closes #6873

### DIFF
--- a/server/lib/agentCompletionMode.js
+++ b/server/lib/agentCompletionMode.js
@@ -71,7 +71,7 @@ export const COMPLETION_MODES = Object.freeze({
   TUI: 'tui',
   /** Worktree with no PR: commit only, PortOS merges the branch back. */
   PORTOS_MERGES: 'portos-merges',
-  /** Worktree with a PR: commit only, the system pushes and opens it. */
+  /** Worktree with a PR: commit only; `/do:pr` when the host can type it, otherwise the system pushes and opens it. */
   WORKTREE_NO_PUSH: 'worktree-no-push',
   /** Plain checkout: commit and push. */
   COMMIT_AND_PUSH: 'commit-and-push',
--- a/server/lib/slashdoInvocation.js
+++ b/server/lib/slashdoInvocation.js
@@ -379,8 +379,9 @@ export function slashdoSkillName(command) {
 
 /**
  * Which invocation shape a provider gets — the single home for the
- * provider→slashdo-shape decision (`hasSlashdo` / `tuiSlashdoFree` in
- * `agentPromptBuilder.js` derive from this rather than re-deriving it).
+ * provider→slashdo-shape decision (`agentPromptBuilder.js`'s
+ * `canTypeSlashCommands` capability flag, and the completion `mode` it feeds,
+ * derive from this rather than re-deriving it).
  *
  * Detection reuses the shared provider predicates, so a path-configured or
  * renamed binary is recognised. An unidentified provider falls through to
@@ -447,11 +448,12 @@ export function resolveSlashdoStyle({
  * gets `SKILL` because it skips command discovery entirely.
  *
  * This is the single home for the completion-workflow gates in
- * `agentPromptBuilder.js` (formerly three inline provider-id allowlists:
- * `hasSlashdo`, `tuiSlashdoFree`, and the guideline-bullet `slashdoFree`). It
- * uses the `assumeClaudeWhenUnknown` posture because those gates describe a
- * session the spawners are about to launch, and every spawner resolves a blank
- * command to `claude`.
+ * `agentPromptBuilder.js` (formerly three inline provider-id allowlists, and
+ * later a trio of per-shape capability flags the completion sections
+ * re-derived; both collapsed to this predicate plus the resolved completion
+ * `mode` — see #6873). It uses the `assumeClaudeWhenUnknown` posture because
+ * those gates describe a session the spawners are about to launch, and every
+ * spawner resolves a blank command to `claude`.
  *
  * `assumeClaudeWhenUnknown` defaults to `true` here (unlike `resolveSlashdoStyle`)
  * because the callers are describing a CLI/TUI session about to be spawned. An
--- a/server/services/agentCompletionCleanup.test.js
+++ b/server/services/agentCompletionCleanup.test.js
@@ -145,7 +145,7 @@ describe('handlePipelineProgression', () => {
 
 // #3114 — `agentOwnsPR` decides whether PortOS skips its own push+PR because the
 // agent was told to run `/do:pr` itself. It MUST derive from the same
-// `canTypeSlashCommands` predicate the prompt's `hasSlashdo` gate used: when the
+// `canTypeSlashCommands` predicate the prompt's completion-mode gate used: when the
 // two disagree, PortOS fires `gh pr create` on a branch that already has a PR
 // ("a pull request already exists" preserves the worktree as a false-positive
 // failure), or conversely never opens the PR the agent was told not to open.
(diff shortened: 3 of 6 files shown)


atomantic:
Orchestrator re-review (opencode, effort high) returned three findings; none applied, for the record:

1. `worktreeCommitGuidance` falling through when a higher-precedence mode (read-only / no-code / tool-free / sentinel / review follow-up) coincides with a slashdo-capable CLI worktree and `openPR` — that is the issue's design (branch on the resolved mode), and in those shapes the old text contradicted the completion section actually rendered. Every coherent shape is byte-identical, as the snapshot suites pin.
2. `PORTOS_MERGES` with `willOpenPR` true cannot be produced: `portosMergesBranchOnExit` is `Boolean(worktreeInfo) && !willOpenPR`, and the light path resolves the mode from that same value right before dispatch.
3. Keeping `slashdoFree` / `hasSlashdo` aliases contradicts the acceptance criterion that removes them; every in-repo call site was updated (`git grep` is empty).

PR #7036, opened 2026-09-11
fix: cancel stale AgentDetail loads (#7000)

### LINKED ISSUE #7000: [react-lifecycle-agent-detail-route-race] Cancel in-flight personality and account fetches on agentId switch in AgentDetail
### UI Lifecycle & State Audit: AgentDetail Personality and Platform Account Route Race

**Audited Slice:** Agent Persona Details and Platform Account Navigation (`client/src/components/agents/AgentDetail.jsx`) — 2026-09-11

### Problem
In `client/src/components/agents/AgentDetail.jsx:25-42`, both `fetchAgent` (`api.getAgentPersonality(agentId)`) and `api.getPlatformAccounts(agentId)` run upon `agentId` changes without cancellation tokens or active ID checks:
```javascript
  const fetchAgent = useCallback(async () => {
    const data = await api.getAgentPersonality(agentId).catch(() => null);
    if (!data) {
      setNotFound(true);
      setLoading(false);
      return;
    }
    setAgent(data);
    setLoading(false);
  }, [agentId]);

  useEffect(() => {
    fetchAgent();
  }, [fetchAgent]);

  useEffect(() => {
    api.getPlatformAccounts(agentId).then(setPlatformAccounts).catch(() => {});
  }, [agentId]);
```

### Impact
1. **Wrong Personality State:** When switching between agents in `/agents/:agentId`, if the response for agent 1 finishes after agent 2, `agent` is overwritten with agent 1's personality.
2. **Tab Availability Drift:** `visibleTabs` (lines 54-60) filters sub-tabs (`ToolsTab` for Moltbook, `WorldTab` for Moltworld) based on `platformAccounts`. An out-of-order resolution applies agent 1's platform accounts to agent 2, hiding or showing tabs inappropriately.
3. **Action Misdirection:** Clicking `handleToggle` calls `api.toggleAgentPersonality(agent.id, !agent.enabled)`, which toggles the stale agent rather than the agent indicated by the URL.

### Fix
In `client/src/components/agents/AgentDetail.jsx`:
1. Use an active flag (`let cancelled = false; ... return () => { cancelled = true; };`) or sequence counter in `fetchAgent` and the `getPlatformAccounts` effect.
2. Reset `agent` and `platformAccounts` when `agentId` changes so stale state does not persist during transitions.

### Dispatch Justification
- `model:light`: Confined to a single component (`client/src/components/agents/AgentDetail.jsx`).
- `effort:low`: Standard effect cancellation for route-dependent queries.

### Acceptance Criteria
- [ ] Navigating between agent IDs never leaves the personality or platform accounts populated from the previous agent.
- [ ] Tab availability matches the active agent's platform registrations.
- [ ] Toggling an agent's enabled state always operates on the agent specified in the URL.

Scope: small

atomantic:
## Summary
- Cancel stale personality and platform-account updates when the AgentDetail route changes.
- Clear the previous agent and account state during transitions so tabs and actions stay scoped to the URL agent.
- Add regression coverage for out-of-order responses, transition clearing, and active-agent toggling.

## Test plan
- `npm test -- --run src/components/agents/AgentDetail.test.jsx`
- `node_modules/.bin/biome lint --error-on-warnings src/components/agents/AgentDetail.jsx src/components/agents/AgentDetail.test.jsx`
- Full client suite: 933 files passed; one unrelated `a11yConventions.test.js` case timed out while probing unavailable `localhost:3000`.

Closes #7000

### DIFF
--- a/client/src/components/agents/AgentDetail.jsx
+++ b/client/src/components/agents/AgentDetail.jsx
@@ -13,6 +13,11 @@ import SchedulesTab from './tabs/SchedulesTab';
 import ActivityTab from './tabs/ActivityTab';
 
 export default function AgentDetail() {
+  const { agentId } = useParams();
+  return <AgentDetailContent key={agentId} />;
+}
+
+function AgentDetailContent() {
   const { agentId, tab } = useParams();
   const navigate = useNavigate();
   const activeTab = tab || 'overview';
--- a/client/src/components/agents/AgentDetail.jsx
+++ b/client/src/components/agents/AgentDetail.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback, useMemo } from 'react';
+import { useState, useEffect, useCallback, useMemo, useRef } from 'react';
 import { Link, useParams, useNavigate } from 'react-router';
 import { ArrowLeft } from 'lucide-react';
 import BrailleSpinner from '../BrailleSpinner';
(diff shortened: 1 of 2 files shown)


atomantic:

atomantic:
Ran Claude once with medium effort, plan mode, only Read/Glob/Grep, and MCP/hooks/plugins/network tools disabled. The tree was unchanged by the reviewer. Status: no-verdict (optional, non-blocking): the output included prose before its structured findings. The configured one-round budget is exhausted.

Validated the substantive transition finding and fixed it in c6d9baa40 by keying the detail state to the URL agent. Also reproduced and fixed the directly related delayed-toggle race that invalidated the next agent request and left its loading indicator stuck. Both new regression assertions fail against the original PR implementation; all 3 focused tests and lint pass with the fix.

Excluded the future callback-argument nit (current callers pass no arguments) and the proposed unmount test (asserting absent DOM after unmount does not establish stale-update suppression).

Local build cannot resolve the declared @scalar/api-reference-react package because it is absent from the shared primary node_modules. No dependencies were installed or shared files modified. Awaiting the current-head CI client build before merge.

Issue #7164, opened 2026-09-12
[resilience-restore-manifest-read-failure] Refuse database restore when an existing integrity manifest is unreadable


atomantic:
Audited 2026-09-12. Slice: database restore integrity checks, HTTP route, and Settings restore preview.
Scope: small
Severity: 8/10.

## Problem
server/services/backup.js:924 reads manifest.json with readJSONFile(manifestPath, null). A malformed or unreadable existing file becomes null, exactly like a missing legacy manifest; :926 then skips the hash comparison. restorePostgres reports an OK dry-run at :938 and can launch destructive psql replay at :959. The dump includes DROP/recreate statements; a transaction prevents partial SQL application but does not establish that the chosen bytes are intact.

Trigger: damage manifest.json on a snapshot (truncated JSON or filesystem read failure). Today both preview and execution proceed without integrity verification. If the dump has also suffered syntactically valid corruption, it can replace the live DB without the expected hash check. The Settings caller at client/src/components/settings/BackupTab.jsx:200 trusts the OK preview and opens confirmation.

Existing tests at server/services/backup.test.js:896 and :908 explicitly demonstrate this fail-open behavior and passed during the audit. They came from closed #3918, a test-coverage task; this issue intentionally changes the failure contract, not duplicate test coverage. Missing legacy manifests remain supported.

## Fix
Use the existing strict JSON read outcome from server/lib/jsonIo.js in server/services/backup.js. Allow ENOENT legacy fallback, but return a stable failed/manifest_unreadable result for an existing malformed/unreadable manifest before preview success or psql spawn. Preserve the existing behavior for a readable legacy manifest with no dump hash. Update BackupTab.jsx to explain that verification metadata could not be read and to choose another snapshot or repair the backup media before retrying. Update docs/BACKUP.md and the existing corruption tests.

## Acceptance criteria
- [ ] Malformed JSON and non-ENOENT manifest read errors return failed/manifest_unreadable in both dry-run and real restore.
- [ ] No psql subprocess starts after either failure.
- [ ] Absent legacy manifests and readable legacy manifests without a dump hash still work.
- [ ] Matching and mismatching hash behavior remains covered.
- [ ] The UI does not open confirmation for the failure and provides an actionable explanation.
- [ ] Tests use synthetic files/mocked subprocesses, never the live database.

Dispatch: model:medium because the service contract, UI message and existing tests change together; effort:medium because the strict-read boundary and legacy exceptions need explicit verification, while the fix reuses an established helper.

PR #7191, opened 2026-09-12
fix: stage client builds before publishing served index (#7182)

### LINKED ISSUE #7182: [console-client-build-clears-live-dist] Stage client builds before publishing the served index
Audited 2026-09-12. Slice: client build publication and built-client HTTP serving. Severity: 6/10. Scope: medium.

## Problem and runtime evidence
The real server log recorded `Server error [GET /]: ENOENT ... stat <repo>/client/dist/index.html` and an identical error for `/system-resources/overview` on 2026-09-11 at 23:15:52 UTC. Those historical lines prove the missing served index, but do not identify the remover.

The current normal build path independently reproduces that condition. Root `package.json:17` delegates to `client/package.json:12`, which runs `vite build`. `client/vite.config.js:214` leaves output-directory clearing at Vite's default. Vite clears dist before writing the replacement bundle. Meanwhile `server/services/assetMounts.js:183` serves that directory and `server/index.js:454` through line 483 reads its index for page requests. After the freshness interval, `server/lib/buildId.js:75` replaces the cached snapshot with the dev/null snapshot when index.html is missing, and the sendFile fallback fails on the same missing path.

Temporary reproduction outside the checkout used the installed Vite with its default output behavior, a tiny synthetic client, and a copy of the unchanged build-ID module feeding the same Express fallback. An observer hook held the render phase past the existing one-second freshness interval: the index was absent, a page request returned HTTP 404, and middleware received ENOENT/stat. Before building the response was 200; after building the index existed again. No live rebuild was performed.

## Impact
A manual `npm run build` against a running built-UI server interrupts page navigation and removes existing hashed chunks during publication. A build that fails after clearing dist can leave the UI unavailable until repaired. All built-UI routes share this failure, although ordinary navigation was healthy during this audit. The standard self-update path stops processes before its build; this proposal addresses the supported direct build command while a server is running, not that separate shutdown interval.

## Chosen fix
Add a build-publication wrapper under `scripts/` and route the normal and analysis build commands in `client/package.json` through it. Build into a unique staging directory on the same filesystem. Validate the staged index and its local referenced assets before publishing. Copy staged assets into the served tree before replacing index.html via a temporary sibling and rename; retain previous hashed assets so in-flight pages keep working. Publish the index last, only after successful completion. Failed builds must leave the previously served index/assets intact. Clean abandoned staging directories without deleting the active bundle; do not use emptyOutDir:false alone, cache missing HTML, or suppress the error as a substitute. Keep analyzer output and existing build stamps working. Adjust `client/vite.config.js` only where needed to respect the staging destination.

Dispatch: model:medium for routine build-script/config integration; effort:high for publication ordering, failure injection, and cross-platform validation.

## Acceptance criteria
- [ ] A temporary Express fixture serving a previous build returns usable HTML and referenced assets throughout a second build and publication.
- [ ] Inject a build failure after rendering begins; the previous complete build remains served and no ENOENT page error occurs.
- [ ] The replacement index is exposed only after all its local asset references are available; previous hashed chunks remain readable for existing pages.
- [ ] Normal and analysis builds preserve build identity and expected output locations on supported platforms.
- [ ] Tests run entirely in temporary fixtures; do not rebuild or remove the live checkout's dist during validation.

atomantic:
## Summary

- Build the client in a unique staging directory and validate `index.html` plus every same-origin boot asset before publication.
- Publish changed files through atomic sibling renames, keep byte-identical and previous hashed chunks intact, then replace the served index last.
- Route normal and analysis builds through the publisher, including staged analyzer output and cleanup for failed or abandoned stages.

## Test plan

- `server/node_modules/.bin/vitest run scripts/publish-client-build.test.js scripts/direct-invocation-drift.test.js scripts/clientBuildEnv.test.js`
- Production-mode analysis build into an isolated temporary output directory; verified 40 local index references, analyzer report output, and embedded branch build identity.
- `node --check scripts/publish-client-build.js`
- `git diff --check`

Closes #7182

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ node_modules/
 # Build output
 dist/
 build/
+/client/.dist-stage-*/
 
 # Runtime data (app registry, history, runs, migration state). Override the
 # directory-level `data/` rule from the global ~/.gitignore so git descends
--- a/client/package.json
+++ b/client/package.json
@@ -9,8 +9,8 @@
   },
   "scripts": {
     "dev": "vite --host 0.0.0.0 --port 5554",
-    "build": "vite build",
-    "build:analyze": "ANALYZE=true vite build",
+    "build": "node ../scripts/publish-client-build.js",
+    "build:analyze": "node ../scripts/publish-client-build.js --analyze",
     "lint": "biome lint --error-on-warnings src",
     "preview": "vite preview",
     "test": "vitest run",
--- a/client/vite.config.js
+++ b/client/vite.config.js
@@ -15,6 +15,7 @@ import {
 
 const ANALYZE_BUNDLE = process.env.ANALYZE === 'true';
 const CONFIG_DIR = import.meta.dirname;
+const BUILD_OUT_DIR = process.env.PORTOS_CLIENT_BUILD_OUT_DIR || resolve(CONFIG_DIR, 'dist');
 
 const rootPkg = JSON.parse(readFileSync(resolve(CONFIG_DIR, '../package.json'), 'utf-8'));
 
@@ -152,7 +153,7 @@ export default defineConfig(({ command, mode }) => {
     plugins: [
       react(),
       ANALYZE_BUNDLE && visualizer({
-        filename: 'dist/bundle-report.html',
+        filename: resolve(BUILD_OUT_DIR, 'bundle-report.html'),
         gzipSize: true,
         brotliSize: true,
         template: 'treemap',
--- a/scripts/publish-client-build.js
+++ b/scripts/publish-client-build.js
@@ -1,0 +1,214 @@
+/**
+ * Build the client away from the served dist directory, then publish it with
+ * index.html last. The old index and its content-hashed assets stay usable
+ * throughout a build, including when Vite fails after it starts rendering.
+ */
+import { spawn } from 'node:child_process';
+import { randomUUID } from 'node:crypto';
+import {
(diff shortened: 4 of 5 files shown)


atomantic:
This PR is intentionally left open because required CI failed on Windows. The failing test is `server/services/backup.test.js` → `restorePostgres` → `dry-run reports size/tableCount without spawning psql` in [CI run 34703587166](https://github.com/atomantic/PortOS/actions/runs/34703587166/job/103579541761).

The failure is deterministic and predates this build-publication diff. An isolated reproduction of the Windows path found that the earlier snapshot-listing test leaves `fs.readdir` mocked with Dirent objects. The strict manifest reader added in #7164 reaches the Windows swap-file check, which expects string directory names and throws `entry.toLowerCase is not a function`; restore then reports `manifest_unreadable` instead of the expected legacy no-manifest result. Resetting that test's directory-listing mock is needed before the full Windows gate can pass. No flake rerun was attempted because the failure was reproduced.

The build publication change passed its focused tests and isolated production analysis build. Its claim branch, worktree, and issue markers remain available for continuation; it will not be merged with this CI failure.

After syncing this branch to the completed batch, required CI failed on the same Windows restore test again at head `2bd3d4827fb515cf7093d5d002423e37038bbb70`: [final CI run 34704316595](https://github.com/atomantic/PortOS/actions/runs/34704316595/job/103581528539). The required CI Gate is failed.

PR #7218, opened 2026-09-12
fix: abort the database transaction when dump streaming fails (#7213)

### LINKED ISSUE #7213: [data-safety-import-read-error-commit] Abort the database transaction when dump streaming fails
Audited 2026-09-12 at e0ddd2f41. Slice: database backend-copy import streaming, stream failure lifecycle, SQL transaction semantics, route consumers and existing tests. Severity: 8/10. Scope: small.

## Problem and upgrade scenario
An install moving between native and Docker databases uses POST /api/database/sync to copy its active records into the other backend before switching. server/routes/database.js:24 calls syncDatabase; server/services/dbAdmin.js:485 streams the exported dump through importDumpFile. Exports contain --clean DROP/recreate statements (scripts/db.sh:565).

If reading the dump fails after complete SQL statements have already reached psql, server/services/dbAdmin.js:396 calls finish(1). finish at :343-353 destroys the source but then calls psql.stdin.end() and resolves immediately. It never terminates psql on that error. A normal EOF is successful completion of the SQL script from psql's perspective; ON_ERROR_STOP only catches SQL errors, not the upstream filesystem failure. A complete valid prefix can therefore commit even though PortOS reports failure.

## Evidence and impact
Executed the actual importDumpFile body in an isolated VM with EventEmitter source and real PassThrough child streams. Injected a complete synthetic DROP statement followed by a source-read error. Observed reportedExitCode=1, childReceivedNormalEOF=true, childKilled=false. No live DB, dump or records were read or changed.

PostgreSQL 17 startup.c treats noninteractive stdin as an implicit file at lines 195-202 and commits a successfully parsed single-transaction script at lines 399-407: https://github.com/postgres/postgres/blob/REL_17_STABLE/src/bin/psql/startup.c . Thus lack of an explicit -f is NOT itself a finding. The failure is converting an upstream error into normal end-of-script after destructive SQL has already been forwarded.

An error inside an unfinished COPY may cause rollback, but an error between complete statements can commit dropped or partly repopulated tables in the target, destroying its previous recovery copy. The active source remains intact; do not describe this as proven loss of both databases.

Existing server/routes/database.test.js:109 mocks successful psql exit on stdin finish. Its read-failure case at :218 opens a nonexistent file and checks only a nonzero result; it never covers failure after a valid destructive prefix or verifies rollback-oriented child shutdown. History 339b55c76 preserved this implementation during route-to-service extraction.

## Decided fix
Separate successful EOF from abort in server/services/dbAdmin.js. On source read failure, stop pumping and terminate psql without sending normal EOF; wait for confirmed child close before returning the failure, with bounded escalation if termination stalls. Preserve the original read error in the returned diagnostic and keep completion idempotent across read error, child error, close and timeout. Keep normal byte-transparent filtering and backpressure behavior unchanged.

Extend server/routes/database.test.js with an injected mid-stream read error after a complete SQL prefix, asserting no normal EOF, child termination, and settlement only after close. Add a gated temporary test-database integration case for preservation of preexisting target rows if practical under the repository's test:db rules; never run against the real portos database.

## Acceptance criteria
- [ ] A read error after complete destructive SQL triggers child abort rather than successful script EOF.
- [ ] The caller does not report completion while the failed import child can still write or commit.
- [ ] Existing target records survive the interrupted transaction in a safely isolated DB integration check.
- [ ] Diagnostics retain the source failure, and duplicate lifecycle events settle once.
- [ ] Successful imports, SQL failures, timeout, byte preservation and backpressure remain covered without production sleeps or live data access.

Dispatch: model:medium for a bounded child-process lifecycle repair; effort:high because event ordering, stream shutdown and transaction rollback must be reasoned about together.

LouisDeconinck:
Closes #7213

## What

`importDumpFile` (`server/services/dbAdmin.js`) streams the exported dump into psql running `-v ON_ERROR_STOP=1 --single-transaction`. On a dump **read** failure, the old `finish(1)` called `psql.stdin.end()` — a normal EOF. To psql that is a successfully completed script, so the single transaction **committed** whatever complete statements had already been forwarded — including the export's leading `--clean` DROPs — destroying the target's recovery copy while the sync reported failure.

## Fix

Read failure now takes a distinct abort path instead of `finish()`:

- Pumping stops and stdin is left **open** — no end-of-script EOF reaches psql.
- The child is terminated (`SIGTERM`), so its open transaction rolls back when the connection dies.
- The promise settles only on the child's confirmed `close`, so the caller never sees completion while psql could still write or commit.
- Bounded escalation if termination stalls: `SIGKILL` after a 2s grace, then the failure resolves after a further 2s bound so a wedged child can't hang the request.
- A kill failure surfacing as child `'error'` mid-abort escalates instead of resolving early; read error / child error / close / timeout all settle once via the existing `finished` guard.
- The triggering read error is now carried in the returned `stderr` diagnostic (previously it was only logged, so callers could surface a bare "Import failed" with no cause).

Normal filtering, backpressure, the success EOF path, SQL-error and timeout handling are unchanged.

## Tests

- `server/routes/database.test.js`: the dump source is now injectable (`createReadStream` wrapped) so tests can fail a read **after** a complete SQL prefix. New cases assert no `stdin.end()`/normal EOF, child termination, settlement only after the confirmed close, SIGTERM→SIGKILL escalation under fake timers, and a mid-abort child error settling once with the read error preserved.
- New gated `server/services/dbAdmin.db.test.js` (added to `DB_TEST_INCLUDE`): a real psql imports a synthetic dump that delivers `DROP/CREATE/INSERT/pg_sleep` then fails the read — verified against a live Postgres that the aborted transaction rolls back and preexisting target rows survive.
- Verified the new suite **fails** on the pre-fix implementation.

`npm test --prefix server`: 43,692 tests green. `npm run test:db` (against a `pgvector/pgvector:pg17` container): 360 tests green including the new suite.

### DIFF
--- a/server/routes/database.test.js
+++ b/server/routes/database.test.js
@@ -318,6 +318,35 @@ describe('importDumpFile (no-shell streaming import)', () => {
     }
   });
 
+  it('bounds escalation when both termination signals synchronously emit errors', async () => {
+    vi.useFakeTimers();
+    try {
+      const child = makeFakePsql({ closeOnKill: false });
+      const failKill = () => {
+        child.emit('error', Object.assign(new Error('kill EPERM'), { code: 'EPERM' }));
+        return false;
+      };
+      // Two consecutive failures reproduce Node's synchronous error delivery;
+      // the fallback prevents a regressed implementation overflowing the test runner.
+      child.kill.mockImplementation(() => false)
+        .mockImplementationOnce(failKill).mockImplementationOnce(failKill);
+      spawn.mockReturnValue(child);
+      const stdinEnd = vi.spyOn(child.stdin, 'end');
+      createReadStream.mockImplementationOnce(() => makeFailingSource('SELECT 1;\n'));
+
+      const promise = importDumpFile('/fake/dump.sql', '5561', {});
+      await vi.advanceTimersByTimeAsync(10_000);
+      const result = await promise;
+      expect(child.kill.mock.calls.map(([signal]) => signal)).toEqual(['SIGTERM', 'SIGKILL']);
+      expect(stdinEnd).not.toHaveBeenCalled();
+      expect(result.exitCode).toBe(1);
+      expect(result.stderr).toContain('dump read failed mid-stream');
+      expect(vi.getTimerCount()).toBe(0);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
   it('keeps waiting for close when the child errors mid-abort and settles once', async () => {
     const child = makeFakePsql({ closeOnKill: false });
     spawn.mockReturnValue(child);
--- a/server/routes/database.test.js
+++ b/server/routes/database.test.js
@@ -33,6 +33,17 @@ vi.mock('../lib/db.js', () => ({
   query: vi.fn(async () => ({ rows: [] })),
 }));
 
+// Wrap fs so a test can substitute a dump source that fails mid-stream — the
+// mid-import read error the abort path exists for. Everything else delegates
+// to the real module (the suite writes real temp dumps).
+vi.mock('fs', async (importOriginal) => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    createReadStream: vi.fn(actual.createReadStream),
+  };
+});
+
 // Mock child_process.execFile + spawn at the module level.
 vi.mock('../lib/childProcess.js', async (importOriginal) => {
   const actual = await importOriginal();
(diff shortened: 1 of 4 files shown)


atomantic:
Reviewed with /do:review; disposition: apply (fork permits maintainer edits). Pushed 43bad77134b819cd1aa70b808941b5ad28ec74a3 with all three verified findings resolved.

- **Critical:** `server/services/dbAdmin.js:400` — make SIGKILL escalation one-shot and install its guard/deadline before signalling. Node emits kill errors synchronously; repeated EPERM previously recursively overflowed the stack and crashed the server. The new regression failed before the fix and passes afterward.
- **Improvement:** `server/services/dbAdmin.db.test.js:65` — replace the 300ms startup assumption with a bounded psql execution marker after the destructive SQL. The real database test now proves the prefix ran before the read failure and verifies the original row survives rollback.
- **Improvement:** `server/services/dbAdmin.js:323` — correct completion claims: settlement waits for close OR the abort deadline; deadline expiry does not confirm termination. This also qualifies the stronger wording in the original PR description.

Host self-review covered all four changed files. Selected lifecycle review for abort/kill/settlement ordering and quality review for rollback coverage and completion claims. Both focused reviewers verified their findings were resolved; no findings remain. No external CLI reviewers were configured.

Validation: 43,704 server tests passed (36 skipped); 360 database tests passed against portos_test; production build passed. Focused lifecycle/DB-guard tests passed, and an additional reproduction using Node's real kill implementation confirmed only two attempts under persistent EPERM. No merge requested or performed.

_Generated by /do:review_

Issue #7238, opened 2026-09-12
[resilience-ollama-app-destructive-replace] Stage and roll back the macOS Ollama.app upgrade instead of deleting first


atomantic:
**Slice audited:** local-backend install/upgrade paths — `server/services/localLlm.js`, `server/services/ollamaManager.js`, `server/services/providerRuntimeInstaller.js` — 2026-09-12.

## Problem

`upgradeOllamaMacApp` deletes the user's working Ollama bundle **before** it knows the replacement can land, and swallows the delete's own failure:

```js
// server/services/localLlm.js:423-429
emit('Installing /Applications/Ollama.app…')
// rm the old bundle first — `mv` can't merge with an existing directory on macOS.
await rm(appPath, { recursive: true, force: true }).catch(() => {})
const move = await runStreaming('mv', [extractedApp, appPath], emit, 60_000)
if (!move.success) {
  await rm(tmpDir, { recursive: true, force: true }).catch(() => {})
  return { success: false, error: `Could not install ${appPath}: ${move.error}. ...` }
}
```

Two reachable failures:

1. **`rm` succeeds, `mv` fails** (the 60 s timeout, `ENOSPC`, an MDM/SIP policy on `/Applications`, or the exact permission problem the error message itself anticipates). The working install is already gone, and the failure branch then deletes `tmpDir` too — so the freshly downloaded replacement is destroyed in the same step. The user is left with **no Ollama at all** and no recovery path but a manual reinstall.
2. **`rm` fails and is swallowed** (a straggler process holding a file open past the `pkill`/1.5 s settle at `localLlm.js:412-419`, or a bundle owned by another account). `mv <src> <existing-dir>` on macOS then moves the source *inside* the target, producing `/Applications/Ollama.app/Ollama.app`, and **exits 0**. `move.success` is true, so the function strips quarantine on the outer path, launches it, and returns `{ success: true, note: '... The new binary is now serving requests.' }` — while the user actually has a corrupted, still-old bundle. A partially-completed recursive `rm` makes this worse: a half-deleted bundle with the new one nested inside it.

The success path also never verifies that the version moved. `before` is read at `localLlm.js:404` and reported, but `after` (`localLlm.js:444`) is never compared to it. The Homebrew sibling path in the same file **does** verify — `localLlm.js:537-543` compares `target`/`before`/`after` with `compareSemver` and returns `success: false` when the running version did not advance. The .app path, which is the one that destroys state, has the weaker check.

## Impact

[...]

Issue #7226, opened 2026-09-12
[data-safety-restore-content-comparison] Restore differing file bytes even when size and timestamp match


atomantic:
Audited 2026-09-12 at b03fb7df0. Slice: snapshot creation, file restore, database restore, routes, consumers, tests and upgrade history. Severity: 7/10. Scope: small.

## Problem and upgrade scenario
An install recovering from an upgrade, or restoring a snapshot from another machine, can have a damaged or unwanted live file whose byte length and modification time match the snapshot. `server/services/backup.js:354` invokes rsync in archive mode without content comparison. `restoreSnapshot` verifies the SOURCE manifest at `server/services/backup.js:1141`, but its transfer flags at line 1143 do not force rsync to compare destination bytes. The transfer at line 1153 therefore skips a different destination file when size and mtime match, then reports successful restoration.

This is distinct from #7167: that fixed source integrity, whereas this defect occurs with a completely valid source and valid manifest. It also affects legacy snapshots without a manifest. Timestamp-preserving copies, corruption without metadata changes, and equal-length edits within timestamp resolution make the condition reachable.

## Reproduction and impact
Using real rsync with the exact base flags and two temporary synthetic directories, set snapshot record.json to {"value":"old"} and live record.json to {"value":"new"}, with equal mtimes. Current flags exited 0 and left the live value as new. Repeating with --checksum exited 0 and restored old. No live files were read or changed.

A restore can silently retain the very bytes the user was trying to recover; preview can also omit the file. Existing backup service and route suites pass (178 tests), but mock transfer completion and do not pin this destination-content case.

## Decided fix
Add --checksum to the restore-specific flags in `server/services/backup.js` for both preview and execution. Retain source-manifest verification, literal subdirectory filtering, and legacy compatibility. Keep the flag local to restore rather than changing snapshot creation policy. Update the matching argv assertions in `server/services/backup.test.js`, and add one temporary-directory boundary regression using real rsync through the restore workflow. Document the content comparison in `docs/BACKUP.md`.

## Acceptance criteria
- [ ] A valid snapshot and a different live file with equal size and mtime produce a preview change and restore the snapshot bytes.
- [ ] Equal-content files remain safely skippable.
- [ ] The same content comparison applies to full and selective restore, including a genuinely absent legacy manifest.
- [ ] Manifest mismatch still refuses the operation before live writes; out-of-scope files remain unchanged.
[...]

PR #7247, opened 2026-09-12
fix: stop a failed audio mux from marking a finished CD stitch as failed

### LINKED ISSUE #7237: [resilience-audiomux-inplace-rename] Give the audio-mux in-place replacements the same rollback ffmpeg.js already has
**Slice audited:** in-place media file replacement — `server/services/pipeline/audioMux.js`, `server/lib/ffmpeg.js`, `server/services/creativeDirector/stitchRunner.js` — 2026-09-12.

## Problem

All four audio-mux entry points replace the stitched episode in place with a bare, uncaught `rename`:

- `server/services/pipeline/audioMux.js:149` (`muxMusicBed`)
- `server/services/pipeline/audioMux.js:299` (`muxVoLines`)
- `server/services/pipeline/audioMux.js:495` (`muxCueBed`)
- `server/services/pipeline/audioMux.js:528` (`muxStripAudio`)

Each is `await rename(tmpOut, inputVideoPath);` with no try/catch and no rollback. `inputVideoPath` is verified to exist at the top of every one of these functions, so the destination always exists.

`server/lib/ffmpeg.js:480` already solves exactly this, and its own comment says why it was extracted:

> *"On Windows, fs.rename fails when the destination already exists — but a simple unlink-first would destroy the original if the subsequent rename failed... Extracted because three callers need this identical rollback, and a data-loss-sensitive path with three copies is one that eventually gets fixed in only two of them."*

`installEncodedVideo` is module-private to `ffmpeg.js`; `audioMux.js` never adopted it and has no `IS_WIN`/platform handling at all.

Separately, the functions' own contracts are broken by the throw. Each documents *"Returns `{ ok: false, reason }` on failure so the caller keeps the prior output"* (e.g. `audioMux.js:507-512`), and every caller in `server/services/creativeDirector/stitchRunner.js:179, 193, 206, 213, 228, 240, 258` branches on `result.ok` and degrades gracefully. A thrown `rename` bypasses all of that.

## Impact

**On Windows, every non-trivial CD stitch fails permanently.** `fs.rename` onto an existing destination fails there, so the mux throws out of `maybeMuxPipelineAudio` — which `stitchRunner.js:122` awaits inside the stitch `try` — and lands in the catch at `stitchRunner.js:139-143`, which sets `status: 'failed'` with the rename's error text. The consequences:

- `finalVideoId` is never recorded and `status: 'complete'` is never written (`stitchRunner.js:127-131`), so a successfully rendered and stitched episode is orphaned on disk and the project reads as Failed.
- The project is never added to its collection (`stitchRunner.js:134-136`).
- The temp file is only unlinked on the *ffmpeg-failure* branch (`audioMux.js:144, 296, 492, 525`), so each attempt leaves a `<episode>.muxing.<uuid>.mp4` / `.cuemux.` / `.silent.` sibling in the videos directory forever.
- The user-facing `failureReason` is a raw `EPERM: operation not permitted, rename ...` rather than "music bed skipped".

On POSIX the same throw is reachable whenever `rename` fails for a non-clobber reason (EACCES, ENOSPC, EBUSY on a file another process holds open) — with the identical "stitch marked failed, episode orphaned" outcome, and with no backup of the original to fall back to.

## Fix (decided)

Reuse the existing, already-tested rollback helper rather than writing a fifth variant.

1. `server/lib/ffmpeg.js` — export `installEncodedVideo` (currently a module-private `const` at `:480`), add it to the `server/lib/index.js` barrel and a row in `server/lib/README.md` per the Module Organization rule.
2. `server/services/pipeline/audioMux.js` — replace all four bare `rename(tmpOut, inputVideoPath)` calls with `installEncodedVideo(tmpOut, inputVideoPath, '<label>')` and return its `{ ok: false, reason }` straight through, so the documented graceful-degradation contract holds on every platform. Labels: `music bed`, `VO mux`, `cue bed`, `silent strip`.
3. Drop the now-unused direct `rename` import from `audioMux.js` if nothing else uses it.
4. Tests in `server/services/pipeline/audioMux.test.js`: with the install step stubbed to fail, each of the four functions resolves `{ ok: false, reason }` (never throws), the original `inputVideoPath` bytes are unchanged, and no `*.muxing.*` / `*.cuemux.*` / `*.silent.*` temp file is left behind.

## Acceptance criteria

- [ ] `installEncodedVideo` is exported from `server/lib/ffmpeg.js`, re-exported from the `server/lib/index.js` barrel, and has a row in `server/lib/README.md`.
- [ ] All four `rename(tmpOut, inputVideoPath)` sites in `server/services/pipeline/audioMux.js` go through it.
- [ ] A failing install returns `{ ok: false, reason }` from `muxMusicBed`, `muxVoLines`, `muxCueBed` and `muxStripAudio` — no throw escapes into `stitchRunner.maybeMuxPipelineAudio`.
- [ ] On a simulated install failure the CD stitch still reaches `status: 'complete'` with `finalVideoId` set, logging the "skipped" warning path the callers already have.
- [ ] The original episode file is byte-identical after a failed mux, and no `*.muxing.*` / `*.cuemux.*` / `*.silent.*` temp file remains.
- [ ] `server/lib/ffmpeg.js`'s three existing `installEncodedVideo` callers are unchanged in behaviour (existing tests still pass).

**Scope:** small

`model:light` / `effort:medium` — the change is mechanical once the existing helper is exported, but it touches four sites plus the barrel/README contract and needs failure-path tests, so it is more than a one-line edit.

atomantic:
## Summary

All four audio-mux entry points in `server/services/pipeline/audioMux.js` replaced the stitched episode with a bare, uncaught `rename(tmpOut, inputVideoPath)`. Each function verifies `inputVideoPath` exists first, so the destination is always present — which on Windows makes that rename **always** fail.

The throw escaped `stitchRunner.maybeMuxPipelineAudio` into the stitch's catch, so a successfully rendered and stitched episode was written off as `status: 'failed'`: never recorded as `finalVideoId`, never added to its collection, orphaned on disk beside a permanently stranded `.muxing.` / `.vomux.` / `.cuemux.` / `.silent.` temp file, with a raw `EPERM: ... rename` as the user-facing reason. On POSIX the same throw is reachable on EACCES/ENOSPC/EBUSY.

It also broke each function's own documented contract — all four promise `{ ok: false, reason }` so the caller keeps the prior output, and every call site in `stitchRunner.js` already branches on `result.ok` to degrade gracefully.

- Export `installEncodedVideo` from `server/lib/ffmpeg.js` (move the original aside, install, restore on any failure) instead of writing a fifth rollback variant, and add it to the `server/lib/README.md` row. The `server/lib/index.js` barrel already re-exports the module wholesale.
- Route all four mux sites through it, returning its `{ ok: false, reason }` straight through.
- Drop the now-unused `rename` import.

Its own comment predicted this: *"a data-loss-sensitive path with three copies is one that eventually gets fixed in only two of them."*

## Tests

- `server/lib/ffmpeg.test.js` — `installEncodedVideo` had **no** direct coverage despite being the module's one destructive helper. Adds its contract: installs over an existing target, installs onto a missing target, keeps the original byte-identical when the install fails, and removes the temp encode. Failures are injected against the real filesystem (no stubbing of the helper under test) and both injections fail identically on POSIX and Windows.
- `server/services/pipeline/audioMux.test.js` — one case per entry point: with the install driven to a real failure (fake ffmpeg exits 0 without producing its output), each resolves `{ ok: false, reason }` rather than throwing, the episode is byte-identical, and nothing is stranded beside it.
- `server/services/creativeDirector/stitchRunner.test.js` — new; `stitchRunner` had no tests. A failed in-place install still reaches `status: 'complete'` with `finalVideoId` set and the episode added to its collection; the companion case proves the mux really installs when it can, so the failure branch is real.

## Test plan

```
cd server && npm test
```

Verified locally: the touched suites are green (90 tests). Mutation-probed every new assertion — reverting to the bare `rename` fails all four audioMux cases and reproduces the exact reported symptom in the stitchRunner test (`status: 'failed'` carrying the raw rename errno); removing the rollback's `unlink` fails the temp-cleanup test. Four pre-existing failures in `services/creativeDirector/` (`localBackend`, `planAdvance`, `recovery`) reproduce on a clean `origin/main` worktree and pass in isolation — unrelated load flakes.

Closes #7237


atomantic:
### Two extra commits: unblocking the full Windows suite

This branch changes `server/lib/ffmpeg.js`, which is barrel-exported and widely imported, so the CI impact planner reports `full: true — targeted test set exceeded safety cap`. **Full mode is the only mode that runs the whole suite on the Windows runner**, and it had accumulated two latent failures from PRs that merged today under targeted mode and never reached Windows. Neither is related to #7237; both are fixed here because they block this PR (and every future full-mode PR):

- **`backup.test.js`** — the one case that shells out to a **real** rsync (added in #7226) had no gate, so it failed `spawn rsync ENOENT` on the Windows runner, which ships no rsync. Now gated on probing `resolveRsyncBinary()` rather than on `process.platform`, so a Windows box with MSYS rsync still gets the coverage and a Linux box without it skips instead of failing.
- **`localLlm.test.js`** — the macOS Ollama.app swap suite (added in #7238) pins `process.platform` to darwin so it runs everywhere, but its fake `pathExists` matched the extracted bundle with `endsWith('/Ollama.app')`. The code under test builds that path with `path.join`, which spells it `\Ollama.app` on Windows, so the fake reported "no Ollama.app in the archive" and every assertion in the block collapsed onto that one wrong error. Now matches either separator — the suite keeps running on every runner rather than being skipped.

A third failure, `peerSync.test.js > forcePushRecord > pushes even when existing sub lastPushedHash matches`, is a timing race the test's own comment already documents ("racy in slower CI"); its `vi.waitFor` expired once on the Windows runner and passed on re-run. Left alone rather than edited on a guess — worth a follow-up if it recurs.

PR #7267, opened 2026-09-13
fix(a11y): give the four hand-rolled tablists TabPills' roving tabindex

### LINKED ISSUE #7244: [a11y-handrolled-tablists-bypass-tabpills] Four hand-rolled tablists never got the roving tabindex and arrow keys TabPills has
### Audited slice
- Slice: every hand-rolled `role="tablist"` in `client/src/` that does not go through the shared `TabPills` primitive — `client/src/components/messages/InboxTab.jsx`, `client/src/pages/VideoTimelineEditor.jsx`, `client/src/pages/VoiceCallHost.jsx`, `client/src/pages/MoodBoardDetail.jsx`
- Date: 2026-09-12
- Verified live against the running instance over CDP (`/messages`, `/settings`, `/goals`, `/media`), not from source alone.

### Problem

#6910 added roving tabindex and arrow-key handling to `client/src/components/ui/TabPills.jsx` (`handleTabKeyDown` at lines 72-96, `tabIndex={active ? 0 : -1}` at lines 145/197). Four surfaces roll their own tab bar instead and never got it:

- `client/src/components/messages/InboxTab.jsx:533-545` — `role="tablist"` / `role="tab"` / `aria-selected`, no `tabIndex`, no `onKeyDown`
- `client/src/pages/VideoTimelineEditor.jsx:717-723` — same
- `client/src/pages/VoiceCallHost.jsx:323-337` — same
- `client/src/pages/MoodBoardDetail.jsx:452-465` — same

This also violates the documented convention in `client/src/AGENTS.md`: *"The tab bar is `client/src/components/ui/TabPills.jsx` … **reuse it, never roll a new tab bar.**"*

Live evidence — the defect and the correct implementation sit **on the same page**, `/messages`, so the drift is directly comparable:

```
## TABLISTS: 2
  "Messages sections"  tabs=7  tabindex-1=6  <-- TabPills: correct roving tabindex
  "Triage filters"     tabs=6  tabindex-1=0  <-- InboxTab: all 6 tabs in the tab order
```

For contrast, every other tablist observed live is a `TabPills` instance and is correct: `/settings` "Settings sections" `tabs=15 tabindex-1=14`, `/goals` "Goals views" `tabs=2 tabindex-1=1`, `/media` "Media Gen sections" `tabs=7 tabindex-1=6`.

### Impact

Because `role="tab"` is present, assistive technology switches to the composite-widget model and announces these controls as "tab, 1 of 6". A screen-reader or keyboard user then presses Left/Right — the interaction the role promises — and nothing happens; the only way to change the filter is to Tab through every remaining tab. On `/messages` that is 6 extra tab stops between the filter bar and the message list, on every visit, versus 1 for the correctly-implemented `TabPills` bar immediately above it.

So the announced role and the actual keyboard behaviour disagree (WCAG 4.1.2 Name, Role, Value), and the WAI-ARIA Tabs pattern's required arrow-key navigation is absent. `VoiceCallHost.jsx:323-337` additionally has `disabled` tabs, which a roving-tabindex implementation has to skip — `TabPills` already handles this via `enabledTabIndexes` (line 76).

No guard catches this: `client/src/a11yConventions.test.js` has no `role="tablist"` rule, and roving tabindex is asserted only inside `TabPills.test.jsx`. That is why #6910's fix did not reach these four.

### Fix

Decision: **replace all four with `TabPills`** rather than hand-adding roving tabindex four more times. All four are plain label(+icon)(+count) tab bars, which is exactly `TabPills`' shape — `InboxTab`'s tabs carry a count badge, which `TabPills` already supports via `count` on the tab descriptor, and `VoiceCallHost`'s `disabled` state is supported too. This removes the drift source instead of duplicating the fix.

1. `client/src/components/messages/InboxTab.jsx:533-558` — render `<TabPills>` with `TRIAGE_TABS` mapped to `{ id, label, icon, count }`; keep the existing `overflow-x-auto` phone behaviour via `TabPills`' own scrolling strip.
2. `client/src/pages/VideoTimelineEditor.jsx:717-730`, `client/src/pages/VoiceCallHost.jsx:323-345`, `client/src/pages/MoodBoardDetail.jsx:452-470` — same substitution; pass `disabled` on the VoiceCallHost descriptors.
3. **Add the missing tree-wide guard** to `client/src/a11yConventions.test.js`: fail when a file writes a literal `role="tablist"` outside `components/ui/TabPills.jsx`. This is the backstop #6910 lacked.

Scope: medium

### Acceptance criteria
- [ ] No `role="tablist"` literal remains in `client/src/` outside `components/ui/TabPills.jsx`.
- [ ] On `/messages`, the "Triage filters" bar reports exactly one tab with `tabIndex === 0` and the rest `-1`.
- [ ] Left/Right/Home/End move between triage filters and change the active filter; Tab moves out of the tab bar in one press.
- [ ] The `disabled` tabs on `/voice/call-host` are skipped by arrow-key navigation rather than being focused.
- [ ] Existing behaviour is unchanged for mouse users on all four surfaces (active styling, counts, mobile horizontal scroll).
- [ ] A new rule in `client/src/a11yConventions.test.js` fails on a newly hand-rolled `role="tablist"` and passes on the current tree.

LouisDeconinck:
Closes #7244

## Problem

Four surfaces rendered `role="tablist"` by hand, so they promised composite-widget keyboard behavior they didn't deliver: every tab sat in the normal tab order (no roving `tabindex`) and Arrow/Home/End did nothing. The issue's decision was to migrate them to the shared `TabPills` primitive rather than duplicate roving-tabindex logic four more times.

## Changes

- `components/messages/InboxTab.jsx` — triage filters now render `TabPills` (underline variant); labels, icons, count badges, URL-synced active state, and horizontal-scroll behavior preserved.
- `pages/VideoTimelineEditor.jsx` — clip library now renders `TabPills variant="pills" size="xs"`.
- `pages/VoiceCallHost.jsx` — Call/Capture mode switch now renders `TabPills`; `disabled` descriptors keep the locked-when-attached behavior and drop disabled tabs from the arrow-key order.
- `pages/MoodBoardDetail.jsx` — Image/Note item-type switch now renders `TabPills variant="pills" size="sm"`.
- `a11yConventions.test.js` — new tree-wide convention: only `components/ui/TabPills.jsx` may write a literal `role="tablist"`. The check lexes real JSX opening tags (masked comments, dynamic role expressions, and `<TabPills>` itself don't trip it), and probe assertions pin both detection and the single exemption.

After this change, `TabPills.jsx` is the only file under `client/src/` containing a `role="tablist"` literal.

## Regression coverage

Per-surface tests added asserting the borrowed contract: exactly one `tabindex=0` per bar, ArrowRight movement (with focus following), Home/End on the inbox triage bar, and VoiceCallHost's attached state disabling both tabs.

## Verification

- `vitest run` on the four touched suites + `a11yConventions` + related Messages/skeleton suites: **109 tests, all passing** (8 files).
- `biome lint` on the 9 changed files: clean.
- No server-side changes; no visual changes beyond adopting the standard primitive styling.


atomantic (approved):
✅ **Approved**

Approved: Migrates four hand-rolled tablists to the shared `TabPills` component to improve accessibility and consistency. The change also adds a tree-wide convention to prevent new hand-rolled tablists. All tests pass, and the a11y conventions test has been updated to enforce this.

**Scope:** a11y improvement: migrated four tablists to TabPills

#### Test evidence
- ✅ `vitest run` — 109 tests passed.
- ✅ `biome lint` — No lint errors found in the 9 changed files.

- The four surfaces now render `TabPills`.
- The `a11yConventions` test correctly identifies hand-rolled tablists.
- Arrow keys and Home/End navigate the tablists correctly.
- The disabled state of the tabs functions as expected.
</details>

PR #7273, opened 2026-09-13
fix(a11y): drop role=menu from popovers that never implement arrow-key focus

### LINKED ISSUE #7265: [a11y-handrolled-menus-break-the-role-menu-contract] Seven hand-rolled popovers claim role=menu without arrow-key focus management, one of them wrapping a textbox and a form
Slice audited: every `role="menu"` container in `client/src` (7 sites, 5 files) plus the two that implement the pattern correctly. Audited 2026-09-13, source review.

## Problem

The WAI-ARIA menu pattern is a contract: a `role="menu"` promises that the container manages focus among its `menuitem` children (arrow keys move between items, Home/End jump, Escape closes and returns focus to the trigger), and that the menu owns **only** `menuitem` / `menuitemcheckbox` / `menuitemradio` / `group` / `separator`. Screen readers act on that promise — NVDA and JAWS suppress the virtual/browse cursor inside a `role="menu"` and switch to focus mode, where arrow keys are the *only* way to read and traverse the contents.

PortOS has two implementations that keep the promise:

- `client/src/components/ui/OverflowMenu.jsx:121` — full roving focus, ArrowUp/ArrowDown cycling, explicit Tab handling, Escape-to-trigger.
- `client/src/components/ThemeSwitcher.jsx:108` — arrow keys and Tab handling.

And one place where the repo already made the right call in a code comment:

- `client/src/components/shell/ShellProviderLauncher.jsx:142` — *"Deliberately NOT `role="menu"`/`"menuitem"`: those roles promise arrow-key …"*

Four other popovers claim the role without implementing any of it. `rg ArrowDown` returns **zero hits** in each of these files:

| Site | What it is |
|---|---|
| `client/src/components/media/CollectionPickerShell.jsx:198` | collection picker popover — see below, the worst case |
| `client/src/pages/Catalog.jsx:882` | "Remix into…" dropdown (`<ul role="menu">` → `<li role="none">` → `<button role="menuitem">`) |
| `client/src/pages/Catalog.jsx:918` | "Add to…" dropdown, same shape |
| `client/src/components/universeBuilder/UniverseCategoryEditor.jsx:201` | assign-to dropdown |
| `client/src/components/universeBuilder/UniverseCategoryEditor.jsx:239` | generate dropdown |
| `client/src/components/universeBuilder/UniverseCategoryEditor.jsx:509` | target-kind picker |
| `client/src/components/Layout.jsx:963` | collapsed-sidebar section flyout (`role="menu"` + `role="menuitem"` links) |

`CollectionPickerShell` additionally violates the *content* half of the contract. Its `role="menu"` container (`:198`) owns:

- a search `<input type="text">` with `autoFocus` (`:216-225`) — so initial focus lands on an element the menu role says cannot be there,
- a `<form>` with a text input and a submit button for creating a new collection (`:239-257`),
- and, depending on the consumer, `role="menuitem"` buttons (`media/BulkTargetPicker.jsx:35`) or `role="menuitemcheckbox"` buttons (`media/AddToCollectionMenu.jsx:63`, `media/PinToMoodBoardMenu.jsx:134`).

A textbox and a form are not permitted owned elements of `menu`. AT that implements the pattern will not expose them dependably.

## Impact

A screen-reader user who opens any of these seven popovers is dropped into focus mode by the menu role and then finds arrow keys do nothing — the announced "menu, N items" cannot be traversed the way the role told the user it could. Tab still moves between the items (they are real `<button>`/`<a>` elements), but only by leaving the promised interaction model and guessing; nothing in the UI says so.

For `CollectionPickerShell` the outcome is worse: the search field it focuses on open, and the "create a new collection" form, sit inside a container whose role declares they do not exist. Adding an item to a collection is the primary action on every media surface that uses these three menus, so this is a routinely-hit path.

## Fix

Follow the precedent the repo already wrote down at `ShellProviderLauncher.jsx:142` — **stop claiming a contract these popovers do not implement**, rather than hand-rolling roving focus four more times:

1. `client/src/components/media/CollectionPickerShell.jsx` — drop `role="menu"` from the container (`:198`); give it `aria-label={title}` and leave it a plain popover `<div>`. Drop `role="menuitem"` at `media/BulkTargetPicker.jsx:35` and `role="menuitemcheckbox"` at `media/AddToCollectionMenu.jsx:63` and `media/PinToMoodBoardMenu.jsx:134`; the checkbox-ish rows keep their state with `aria-pressed` on a plain `<button>`, which needs no owner contract. Change any `aria-haspopup="menu"` on the triggers to `aria-haspopup="dialog"` or drop it; keep `aria-expanded`.
2. `client/src/pages/Catalog.jsx:882,918` and `client/src/components/universeBuilder/UniverseCategoryEditor.jsx:201,239,509` — drop `role="menu"` / `role="none"` / `role="menuitem"`; keep the `<ul>`/`<li>`/`<button>` markup, which already conveys "a list of actions", and keep `aria-expanded` on the trigger. Add `aria-label` to each list naming the action set.
3. `client/src/components/Layout.jsx:963` — the flyout is a list of navigation links, not a command menu. Replace `role="menu"`/`role="menuitem"` with `<nav aria-label="{section} pages">` wrapping a `<ul>`/`<li>` of the existing `<a>`/`<Link>` elements.
4. Where a popover genuinely wants menu semantics in future, route it through `ui/OverflowMenu.jsx` rather than re-deriving them.
5. Add a case to `client/src/a11yConventions.test.js` in the style of its existing scanners: a `role="menu"` in a file with no `ArrowDown`/`ArrowUp` key handling fails, with an allowlist seeded to exactly the two compliant implementations. This is what stops the eighth copy landing.

## Acceptance criteria

- [ ] No `role="menu"` / `role="menuitem"` / `role="menuitemcheckbox"` / `role="none"` remains in `CollectionPickerShell.jsx`, `BulkTargetPicker.jsx`, `AddToCollectionMenu.jsx`, `PinToMoodBoardMenu.jsx`, `Catalog.jsx`, `UniverseCategoryEditor.jsx`, or `Layout.jsx`.
- [ ] Each converted popover carries an `aria-label` naming what it contains, and each trigger keeps `aria-expanded` with a correct `aria-haspopup` (or none).
- [ ] The collection-picker rows expose their selected state via `aria-pressed`, and existing tests for toggling a collection still pass.
- [ ] `OverflowMenu` and `ThemeSwitcher` are untouched.
- [ ] `a11yConventions.test.js` gains a guard that fails a new `role="menu"` without arrow-key handling, allowlisting only `OverflowMenu.jsx` and `ThemeSwitcher.jsx`, and the guard is proved by a bypass probe (a fixture that should fail actually fails).

## Why these labels

`model:light` — the design call is made here; execution is a well-specified attribute removal plus one structural swap in `Layout.jsx`, all in files a single agent can hold at once. `effort:medium` — seven sites across five files, each needing its replacement semantics checked against the popover's actual contents, plus a new conventions guard with a bypass probe; not a blind sed.

Scope: medium

atomantic:
## Summary
- Seven hand-rolled popovers across five files claimed `role="menu"`/`role="menuitem"`/`role="menuitemcheckbox"`/`role="none"` without implementing arrow-key focus management — the WAI-ARIA menu contract that NVDA/JAWS depend on to switch into focus mode. `CollectionPickerShell` additionally nested a search `<input>` and a create `<form>` inside its `role="menu"` container, which the menu role does not permit as owned elements.
- Dropped the mismatched roles rather than hand-rolling roving focus a fourth time, following the precedent already noted at `ShellProviderLauncher.jsx:142`:
  - `CollectionPickerShell.jsx` → plain popover with `aria-label={title}`; its two checkbox-style consumers (`AddToCollectionMenu.jsx`, `PinToMoodBoardMenu.jsx`) move from `role="menuitemcheckbox"` + `aria-checked` to a plain button + `aria-pressed`, and their triggers' `aria-haspopup="menu"` becomes `aria-haspopup="dialog"`.
  - `Catalog.jsx`'s "Remix into…" and "Add to universe/series…" popovers, and `UniverseCategoryEditor.jsx`'s three pickers (assign-to-trunk, generate-variations, promote-to-canon), drop the roles, add `aria-label`, and change their triggers' `aria-haspopup="menu"` to `"true"`.
  - `Layout.jsx`'s collapsed-sidebar flyout becomes a `<nav aria-label>` wrapping a `<ul>/<li>` of the existing links, instead of a `role="menu"` div with `role="menuitem"` links.
  - `OverflowMenu.jsx` and `ThemeSwitcher.jsx` — the two implementations that actually wire ArrowUp/ArrowDown — are untouched.
- `a11yConventions.test.js` gains a tree-wide guard (`handRolledMenus`/`menuOffenders`) that fails any `role="menu"` in a file with no `ArrowDown`/`ArrowUp` handling, allowlisting only the two compliant files, proved by a bypass probe against both a failing and a passing fixture.

## Test plan
- `client/node_modules/.bin/vitest run src/a11yConventions.test.js` — new guard passes; the one pre-existing failure (`keyboard-activates every clickable non-interactive element`, a load-sensitive 5s timeout) passes in isolation and is unrelated to this change.
- `client/node_modules/.bin/vitest run` (full client suite) — 949/954 files pass; the 5 failing tests are all pre-existing timeout flakes under full-suite load (confirmed passing individually), none touch the files this PR changes.
- Updated test assertions in `AddToCollectionMenu.test.jsx`, `PinToMoodBoardMenu.test.jsx`, `Catalog.test.jsx`, and `UniverseBuilder.test.jsx` to query the new roles/attributes — all pass.
- `biome lint --error-on-warnings` on every changed file — clean.

Closes #7265


atomantic:
Closing in favor of #7272, which merged first and is more thorough — it also fixes two same-class strays this PR didn't catch (`pages/Shell.jsx`, `dashboard/LayoutPicker.jsx`) and uses `role="group"` + real `<ul>/<li>` list markup in `CollectionPickerShell`, plus a broader regression guard covering the whole menu-role family instead of just `role="menu"`.

PR #7278, opened 2026-09-13
fix(a11y): release Shift+Tab from the Shell terminal and enable screen-reader mode

### LINKED ISSUE #7262: [a11y-shell-terminal-keyboard-trap] Shell terminal traps keyboard focus and renders all output aria-hidden
Slice audited: `client/src/pages/Shell.jsx` + `client/src/hooks/useShellSession.js` (the only `new Terminal()` in the tree). Audited 2026-09-13, source + inspection of the pinned `@xterm/xterm@6.0.0` bundle in `client/node_modules`.

## Problem

The shell terminal is constructed with default options and no host key handler:

- `client/src/hooks/useShellSession.js:259-266` — `new Terminal({ cursorBlink, cursorStyle, fontSize, fontFamily, theme, scrollback, allowProposedApi })`. No `screenReaderMode`, and `term.attachCustomKeyEventHandler(...)` is never called anywhere in the repo (`rg attachCustomKeyEventHandler client/src` → 0 hits).
- The terminal host element is `client/src/pages/Shell.jsx:305`; xterm's helper `<textarea>` inside it is in the tab order, so keyboard focus reaches the terminal normally.

Two consequences, both verified against the shipped bundle rather than inferred:

**1. Complete keyboard trap (WCAG 2.1.2 Keyboard Trap, Level A).** In `@xterm/xterm@6.0.0`'s `_keyDown`:

- `_customKeyEventHandler` is the documented first-line escape hatch — it is the only way a host can let a key through. PortOS never installs one.
- Tab (`keyCode 9`) resolves to `{ key: C0.HT, cancel: true }`, and `cancel` routes to `this.cancel(e, true)` → `e.preventDefault(); e.stopPropagation()`.
- Shift+Tab resolves to `{ key: ESC + "[Z" }` without `cancel`, but then hits the tail branch `!this.optionsService.rawOptions.screenReaderMode || e.altKey || e.ctrlKey ? this.cancel(e, !0) : ...`. Because `screenReaderMode` is false in PortOS, **Shift+Tab is also `preventDefault()`ed.**
- Escape resolves to `C0.ESC` with `cancel: true` and is likewise swallowed.

So a keyboard-only user who tabs into the terminal cannot tab forward out of it, cannot shift-tab back out of it, and cannot press Escape out of it. Every remaining control on the page — the session tabs, the "cd to app folder" picker (`Shell.jsx:271`), and in fullscreen the "Exit fullscreen" button (`Shell.jsx:319`) — becomes unreachable without a mouse. The only recovery is a pointer click or a full page reload.

**2. Terminal output is hidden from assistive technology (WCAG 1.3.1 / 4.1.3).** In the same bundle:

- `this.options.screenReaderMode && (this._accessibilityManager.value = ...AccessibilityManager...)` — the `.xterm-accessibility` / `.xterm-accessibility-tree` live region is created **only** when `screenReaderMode: true`.
- The DOM renderer's row container is created with `setAttribute("aria-hidden", "true")`.

With `screenReaderMode` off, every rendered row is explicitly `aria-hidden` and no live region exists, so a screen-reader user focusing the terminal hears an empty textarea and nothing else. Command output, prompts, and agent TUI frames are entirely inaudible.

## Impact

A keyboard-only or screen-reader user is fully locked out of the Shell page — one of PortOS's core surfaces, and the surface CoS agent sessions are driven from. The trap is worse than a missing feature: it strands the user on a page they cannot leave with the keyboard, which is a Level A conformance failure for the *whole* page, not just the widget.

## Fix

In `client/src/hooks/useShellSession.js`, at the `new Terminal({...})` site:

1. Pass `screenReaderMode: true`. This instantiates xterm's `AccessibilityManager` so terminal output lands in a live region instead of an `aria-hidden` row container. (It also flips the `_keyDown` tail branch so xterm stops blanket-cancelling resolved keys.)
2. Install `term.attachCustomKeyEventHandler(e => ...)` that returns `false` (do not handle; let the browser act) for the focus-escape gesture. Decision: **Shift+Tab always escapes** (it is the conventional "leave this widget" gesture, and a backtab is far less useful to a shell than a forward Tab, which must stay bound for completion). Plain Tab keeps going to the PTY for shell completion.
3. Announce the escape gesture where the terminal is introduced — add it to `client/src/components/KeyboardHelp.jsx` and as visible helper text near the terminal in `client/src/pages/Shell.jsx`, so the affordance is discoverable rather than folklore (WCAG 2.1.2 requires either no trap or a documented exit).
4. Make sure the custom handler defers to the shared global-key predicate in `client/src/lib/a11yKeyboard.js` rather than adding a second, divergent guard (root `AGENTS.md`: "a new guard goes THERE, never at one caller").
5. Sanity-check `client/src/lib/terminalDictation.js` and `client/src/components/shell/TerminalHotKeys.jsx` still behave once `screenReaderMode` changes the cancel path.

## Acceptance criteria

- [ ] `useShellSession.js` constructs the terminal with `screenReaderMode: true`.
- [ ] `term.attachCustomKeyEventHandler` is installed and returns `false` for Shift+Tab, so focus leaves the terminal for the previous tabbable element.
- [ ] A test in `client/src/hooks/useShellSession.test.js` (or a new one) asserts the handler returns `false` for `{ key: 'Tab', shiftKey: true }` and `true` for `{ key: 'Tab', shiftKey: false }`.
- [ ] A test asserts the terminal is constructed with `screenReaderMode: true` (guarding against a future options rewrite silently dropping it).
- [ ] The escape gesture is listed in `KeyboardHelp.jsx` and rendered as visible text on the Shell page.
- [ ] Manual check with a screen reader: focusing the terminal and running a command announces the output (see "help wanted" below).

## Why these labels

`model:heavy` — the fix is small in lines but hinges on xterm's key-dispatch semantics, the interaction between `screenReaderMode` and the cancel path, and PortOS's shared global-hotkey predicate; getting the wrong key released breaks shell completion or re-opens the trap. `effort:high` — one focused change, but each step needs the bundle behavior re-checked rather than assumed.

`help wanted` — step 6 needs validation on real assistive technology that this repo's CI cannot provide: NVDA on Windows and VoiceOver on macOS handle xterm's live region differently (xterm's own docs caveat NVDA specifically), and a terminal's continuous output is the hardest case for announcement pacing. Input from people who actually drive a terminal by screen reader would materially change whether `screenReaderMode` alone is sufficient or whether the announcement needs throttling.

Scope: small

LouisDeconinck:
Closes #7262

## Problem

The Shell page's xterm terminal was a complete keyboard trap (WCAG 2.1.2, Level A): its helper `<textarea>` sits in the tab order, but `_keyDown` `preventDefault`s Tab, Shift+Tab and Escape alike, so a keyboard user who tabbed in could never leave — the session tabs, the "cd to app" picker, and the fullscreen Exit button all became unreachable without a pointer. And with `screenReaderMode` unset, every rendered row is `aria-hidden` with no live region: a screen-reader user focusing the terminal heard an empty textarea and nothing else.

## Changes

- `hooks/useShellSession.js` — `new Terminal({...})` now passes `screenReaderMode: true`, which instantiates xterm's `AccessibilityManager` so output lands in the `.xterm-accessibility` live region, and relaxes `_keyDown`'s blanket cancel on resolved keys.
- `term.attachCustomKeyEventHandler((e) => !isFocusEscapeKey(e))` is installed right after `open()`. The issue's chosen gesture is **Shift+Tab always escapes**: the `false` return makes `_keyDown` bail before it cancels the event, so the browser performs the backtab and focus lands on the previous tabbable element — and unlike bare `screenReaderMode` (which releases the key but still fires `ESC[Z` into the PTY — zsh reads that as reverse completion), the early return sends nothing. Plain Tab stays claimed for shell completion; Escape stays with the TUI.
- `lib/a11yKeyboard.js` — new `isFocusEscapeKey(event)` predicate, living where the shared stand-down guards do rather than inlined at the caller. Chorded variants are not excluded: a Ctrl/Alt+Shift+Tab the browser or OS owns must pass through too.
- `components/KeyboardHelp.jsx` — new "Shell" section listing `Shift+Tab → Move focus out of the terminal`.
- `pages/Shell.jsx` — visible helper text under the terminal (`Shift+Tab moves focus out of the terminal`, `Kbd size="sm"`), so the exit is discoverable rather than folklore.
- `lib/README.md` — the `a11yKeyboard.js` row now documents `isFocusEscapeKey`.

Adjacent fix surfaced by the new tests: the init effect's `requestAnimationFrame` fit is now cancelled on cleanup — an unmount inside the frame window ran `fitTerminal` on a disposed terminal (`_renderService.dimensions` throws post-dispose) and produced unhandled errors in the suite.

## Sanity checks (issue step 5)

- `lib/terminalDictation.js` — already SR-mode-aware by design (`if (terminal.options?.screenReaderMode) return`, pinned by its "stays out of the way in screen-reader mode" test). Note the upstream trade-off that comes with the option: in SR mode xterm's `_inputEvent` doesn't forward `insertText`, so OS dictation insertions land in the helper textarea without reaching the PTY — the same trade-off VS Code accepts.
- `components/shell/TerminalHotKeys.jsx` — button strip emitting synthetic input over the socket; nothing in it touches the key-dispatch path.
- Global key handlers are safe by construction: SR-mode keydowns are no longer `stopPropagation`ed, but they all originate on the helper `<textarea>`, so `shouldIgnoreGlobalKey`'s `isEditableTarget` guard already stands them down.

## Regression coverage

`useShellSession.test.jsx` gains a `terminal accessibility` block that mounts a real `Terminal` against a real host `<div>` (the existing socket-level tests never attach `terminalRef`): `options.screenReaderMode === true` plus the `.xterm-accessibility .live-region` existing in the DOM; the attached handler returning `false` for `{ key: 'Tab', shiftKey: true }` and `true` for plain Tab and Escape; and a real `keydown` dispatch on `term.textarea` proving Shift+Tab is left uncancelled with **zero** `shell:input` emitted, while plain Tab is cancelled and reaches the PTY as `\t`. `a11yKeyboard.test.js` covers the predicate's backtab-only, chord-inclusive contract.

## Verification

- `vitest run src/hooks/useShellSession.test.jsx src/lib/a11yKeyboard.test.js` — **66 tests, all passing** (36 + 30).
- Full client suite (`npm test` in `client/`): 954 files / 11,663 tests — 2 unrelated flaky `act()` failures in `meatspace/post` tests that pass in isolation and touch nothing in this diff.
- `biome lint --error-on-warnings` on all 6 changed source/test files: clean.
- Not verified: real assistive technology (the issue's `help wanted` callout — NVDA/VoiceOver behavior with xterm's live region still warrants a manual pass).


atomantic (approved):
✅ **Approved**

Approve: The fix addresses a critical keyboard trap in the shell terminal, restoring accessibility for keyboard and screen reader users. The changes are well-scoped and include appropriate tests and documentation updates.

**Scope:** Accessibility fix for the shell terminal keyboard trap.

#### Test evidence
- ✅ `vitest run src/hooks/useShellSession.test.jsx src/lib/a11yKeyboard.test.js` — 66 tests passed.

- terminal constructs with screenReaderMode: true
- xterm live region exists
- Shift+Tab releases keyboard focus
</details>

PR #7305, opened 2026-09-13
fix: scale the restore rsync idle deadline to the largest in-scope file

### LINKED ISSUE #7302: rsync's checksum digest of one very large file can still outrun the backup idle watchdog
## Problem

#7299 fixed the common case: `restoreSnapshot` passes `--checksum`, and `-ii` now makes rsync emit one itemize line per file it digests, so a scan of a large *unchanged tree* keeps `runRsync`'s idle watchdog (`BACKUP_PROCESS_IDLE_TIMEOUT_MS`, 10 minutes, `server/services/backup.js:35`) fed.

That heartbeat is **per file**. A single file whose digest alone exceeds the deadline still emits nothing for its whole duration:

- rsync reports an entry only *after* it finishes deciding about it. Neither `-ii` nor `--info=progress2` reports progress *inside* the generator's checksum read — `--progress` reports within-file bytes during a **transfer**, not during a digest.
- `--checksum` reads both copies, so the wall time is (snapshot bytes + live bytes) / destination throughput for that one file.
- `data/` is media-bearing (video, model weights). A ~15 GB file on a network- or iCloud-backed destination at ~25 MB/s is ~20 minutes of silence — twice the deadline.

The symptom is identical to #7299: `watchBackupProcess` calls `startTermination('idle')`, SIGKILLs rsync, and a healthy snapshot's preview fails with `BACKUP_PROCESS_TIMEOUT`. A non-dry-run restore fails the same way, with the added "some files may already have been overwritten" partial-restore path.

## Why it wasn't rolled into #7299

It is a genuinely different condition (one huge file vs. many files) with a different fix shape. #7299's fix is a flag; this one needs the watchdog itself to change, and the obvious change — a longer deadline — makes a genuinely hung rsync take proportionally longer to detect on **every** restore, which is the trade #7299 deliberately declined.

## Fix

Decide between, and implement, one of:

1. **Observe the child instead of only its stdout.** `watchBackupProcess` already has the precedent: it accepts `progressPath` and polls a growing file for the otherwise-silent `pg_dump`. The analogue here is polling the rsync child's I/O counters (macOS `proc_pid_rusage` / Linux `/proc/<pid>/io`) and treating advancing read bytes as progress. Strongest signal — it distinguishes "reading hard" from "wedged" — but needs a per-platform read with a graceful no-op fallback.
2. **Scale the deadline to the largest file in scope.** `verifySnapshotFiles` already walks the selected scope immediately beforehand, so the largest file's size is available for free. Derive a floor from it (bytes / a conservative worst-case throughput) and pass it to `runRsync` as an idle-timeout override. Keeps the tight default for small trees; only a tree that actually contains a huge file relaxes it.

Prefer (1) if the platform read is clean; (2) is the sound fallback and is strictly better than raising the constant globally, because it relaxes only where a long silence is explainable.

Either way `runRsync` and `watchBackupProcess` need an idle-timeout (or progress-source) parameter rather than always reading the module constant.

## Acceptance criteria

- A restore whose rsync stays silent past the deadline while genuinely making progress is **not** terminated.
- A restore whose rsync is genuinely wedged is still terminated, and for a small tree within roughly the current 10 minutes.
- The backup path's watchdog behavior is unchanged (it passes neither `--checksum` nor `-ii`).
- Covered by a focused contract test with injected/fake time and a fake child — no real sleeps, per the repo's test-strategy rule for timeout state machines.

## Files

- `server/services/backup.js` — `BACKUP_PROCESS_IDLE_TIMEOUT_MS`, `watchBackupProcess`, `runRsync`, `restoreSnapshot`, `verifySnapshotFiles`
- `server/services/backup.test.js`

## Dispatch rationale

`model:heavy` because the call is a process-lifecycle/liveness-detection design decision — picking a progress signal that cannot be faked by a hung child — not multi-file plumbing. `effort:medium` because once that signal is chosen the edit is small and confined to one service.

Refs #7299

LouisDeconinck:
Closes #7302

## Problem

`--checksum` digests both copies of every same-size file in scope, and `-ii` only heartbeats once rsync finishes *deciding* about a file — a single file whose digest outlasts `BACKUP_PROCESS_IDLE_TIMEOUT_MS` (10 min) emits nothing for its whole duration, so `watchBackupProcess` SIGKILLs a healthy restore. ~15 GB at ~25 MB/s ≈ 20 min of silence — twice the deadline. Hits preview and real restores identically.

## Fix — option (2) from the issue

The platform read for option (1) isn't clean on the primary target: macOS has no `/proc`, and `proc_pid_rusage` needs libproc FFI with no native-free read from Node — the "graceful fallback" would no-op exactly where the iCloud/network-destination case bites. The size-aware floor is deterministic on every platform.

- `snapshotManifestFilePaths` → `snapshotScopeInventory` now also returns `largestFileBytes` — free, since the preflight already `stat`s every in-scope file. OS-metadata files are excluded (rsync never sees them).
- `verifySnapshotFiles` returns `largestFileBytes` in both the verified and manifest-absent results; the legacy path walks best-effort and falls back to the default deadline when the scope can't be surveyed — never a new failure mode.
- `restoreIdleTimeoutMs` floors the deadline at `2 × largestFileBytes / RESTORE_DIGEST_WORST_CASE_BPS` (10 MiB/s, ~2.5× headroom under the observed ~25 MB/s healthy iCloud speed). Both copies because `--checksum` reads snapshot **and** live bytes; files whose sizes differ skip the digest entirely and go straight to transfer, where `--progress` already heartbeats.
- `runRsync`/`watchBackupProcess` take an `idleTimeoutMs` parameter instead of reading the module constant, and the timeout error reports the effective duration rather than a hardcoded "10 minutes".
- Backup path passes no override — watchdog behavior unchanged; the 4-hour wall backstop is unchanged.

## Acceptance

- ✅ Silent-but-progressing rsync past the deadline is not terminated (floor covers the digest).
- ✅ A genuinely wedged rsync still terminates — small trees at ~10 min, large-file scopes at the derived floor.
- ✅ Backup watchdog unchanged (no override passed).
- ✅ Focused contract tests with fake timers + a fake child — no real sleeps.

## Tests

`server/services/backup.test.js`: +3 tests — silent rsync survives past the default while a 40 GiB file digests but still dies at the scaled ~136.5-min floor; a small-file scope still dies at 10 min; a manifest-absent legacy snapshot gets the same floor. Full file: 152/152 pass. `npm test --prefix server` (44k tests) shows only unrelated load-contention timeouts in ~30 files (`toolRegistry`, `planAdvance`, `peerSync`, etc.); all backup-touching suites pass standalone.


atomantic (approved):
✅ **Approved**

The change correctly addresses the issue of rsync's idle timeout being triggered by large file checksum calculations. It scales the timeout based on the largest file in scope, preventing premature termination of restores for large files while still ensuring a timely exit for genuinely hung processes. Tests cover the scaling behavior and ensure the default timeout remains effective for small trees.

**Scope:** backup service: adjust rsync idle timeout to largest in-scope file

#### Test evidence
- ✅ `npm test --prefix server` — All backup-touching suites pass standalone. Added contract tests with fake timers demonstrating the scaled timeout behavior.

- The largest file size is correctly calculated during manifest verification.
- The restore idle timeout is dynamically adjusted based on the calculated largest file size.
- The timeout error message accurately reflects the effective timeout duration.
</details>

PR #7369, opened 2026-09-14
fix: preserve existing forge label metadata

### LINKED ISSUE #7365: Filing an issue overwrites the user's existing repo labels, because gh label create is passed --force
`forgeLabelCreateArgs` (`server/lib/forgeIssueCli.js:19-20`) passes `--force` to `gh label create`. On GitHub that is not "create if missing" — it **overwrites an existing label's color and description**.

So when the Persistent Mind (or the Layered Intelligence forge filer) files an issue carrying a common category label, it silently rewrites that label repo-wide in the user's own repository.

## Failure scenario

A user has customized `bug` in their repo — a specific color, a description their team relies on. The mind files one issue labeled `bug`. `gh label create bug --color <slashdo default> --description <slashdo default> --force` runs first, and the user's color and description are gone, on every issue already carrying that label. Nothing in the flow told them it would happen, and nothing records the previous values.

The `glab` arm has no `--force`, so the two CLIs already disagree about this — which is the kind of divergence `forgeIssueCli.js` exists to remove.

## Fix

Drop `--force` and treat "label already exists" as success (the command exits non-zero with an already-exists message; swallow that specific case, as the ad-hoc `2>/dev/null || true` callers effectively do). An existing label is the user's, and the value of a matching color is far below the cost of overwriting their metadata.

Add a test asserting `--force` is not in the emitted argument list and that an already-exists exit is not treated as a failure.

Found in the v2.66.0 release review.

faizanxdev:
Fixes #7365

- Remove `--force` from GitHub label creation.
- Preserve existing repository label metadata.
- Accept only the expected duplicate-label response.
- Add regression tests.


atomantic (approved):
✅ **Approved**

The PR removes the `--force` flag from GitHub label creation, preventing accidental overwriting of user-configured label metadata. It also adds regression tests to verify the fix. This is a good change to improve user experience.

**Scope:** Fixes GitHub label creation idempotency.

#### Test evidence
- ✅ `npm test` — All tests passed, including the new regression tests.

- The `--force` flag is removed from the `forgeLabelCreateArgs` function.
- The test suite now includes a test to ensure that attempting to create an existing label does not result in an error.
</details>

PR #7425, opened 2026-09-16
Add a Subscriptions page under Models to manage AI plans

### LINKED ISSUE #7403: Add a Subscriptions page under Models to enable/disable plans and track usage, cost, and spend
## Problem / Goal
The Models section has no single place to answer "which AI subscriptions do I pay for, are they on, and what are they costing/saving me?" Plan prices live in a cost editor buried inside the Usage report (`SubscriptionSavingsCard` on `/devtools/usage`), provider enable/disable lives on the Providers page (`/ai`), and quota/spend lives across `GET /api/usage` and `GET /api/usage/providers`. We want one **Subscriptions** page under Models where the user can enable/disable each subscription, see its usage / cost / spend, and keep the plan definition current (e.g. Claude Max 5x vs 20x, ChatGPT plan tier, and the monthly price). This is PortOS-side metadata and tracking for features only — it never touches downstream provider subscription/billing management.

## Context
Current behavior, grounded in code:
- Subscription identity is the provider **family** (`server/lib/providerFamilies.js`: `claude`, `codex`, `agy`, `grok`; `familyForProvider` maps a provider config to a family, local-runtime-backed providers map to none).
- "Enabled" today is per-provider (`provider.enabled`), with families derived via `resolveEnabledFamilies` (`server/services/providerUsage.js:767`). There is no family-level on/off — a subscription is "off" only when all its providers are disabled.
- Plan prices are per-install config in `data/settings.json` under `subscriptionCosts` (`server/services/subscriptionCosts.js`), written via `PUT /api/usage/subscriptions` (partial patch; `null` clears) and via the `subscriptionCosts` slice of `PUT /api/settings`. Only a monthly USD number is stored — there is **no plan-tier field** (no "5x vs 20x" label).
- The savings/cost math is `buildSubscriptionSavings` (`server/lib/subscriptionSavings.js`) served as the `subscriptionSavings` block on `GET /api/usage`; the editor is `client/src/components/usage/SubscriptionSavingsCard.jsx` (price inputs + per-family period/API/savings figures). Quota cards come from `GET /api/usage/providers`.
- Models section shape: `client/src/pages/Models.jsx` (`TAB_CONTENT`, `?tab` route param) + `ModelsTabsHeader` (tabs from `getSectionNavTabs('Models')` in `server/lib/navManifest.js`). Usage (`/devtools/usage`) and Quota Burn (`/devtools/quota-burn`) are Models-section entries with legacy paths that render the Models tab bar.
- Prior art: `docs/plans/2026-08-21-models-navigation.md` is the design record for what belongs in Models; the `portos-add-page` skill covers the `<Route>` + `NAV_COMMANDS` + sidebar + ⌘K/voice wiring for a new page.

## Proposed approach
1. Add a **Subscriptions** destination in the Models section (new tab/route such as `/models/subscriptions`, registered in `App.jsx`, `server/lib/navManifest.js` as `nav.models.subscriptions` section `Models`, sidebar, and `navPresentation.js` icon) following the `portos-add-page` checklist so it is reachable from ⌘K and voice.
2. Render one card/row per family from `PROVIDER_FAMILIES`, combining: enabled state (from `resolveEnabledFamilies` / `GET /api/usage/providers`), quota/usage + spend (existing `GET /api/usage` report + `subscriptionSavings` block), and the monthly price editor (reuse `SubscriptionSavingsCard` logic or extract it; writes go through the existing `PUT /api/usage/subscriptions` patch semantics).
3. Family-level enable/disable: define it as toggling `enabled` on that family's CLI/TUI provider records (the same flag `resolveEnabledFamilies` reads), keeping priced-but-disabled families visible with their stored price (the `resolveSubscriptionFamilies` extras rule in `subscriptionCosts.js:102` exists precisely so a disabled plan's price is never silently dropped — preserve it).
4. Plan definition: add a way to record the plan tier alongside the price (e.g. "Claude Max 5x / 20x"). Decide during implementation whether that is a new local-settings field next to `subscriptionCosts` (with a `scripts/migrations/` entry if the on-disk shape changes) or display-only labeling; do not overload the numeric price field.
5. Keep the page metadata-only: toggling a subscription here flips PortOS-side enablement/tracking, never calls a provider billing API.

## Acceptance criteria
- [ ] Models section has a Subscriptions page reachable by URL, sidebar, ⌘K, and voice nav
- [ ] Each subscription family (claude, codex, agy, grok) shows enabled/disabled state and can be toggled, backed by the existing provider `enabled` flag
- [ ] Same page shows per-family usage / cost / spend (quota status + savings block figures) and lets the user update the monthly plan price (existing `PUT /api/usage/subscriptions` semantics: omitted keeps, `null` clears)
- [ ] User can record/update which plan tier they are on (e.g. 5x vs 20x) alongside the price
- [ ] Disabled-but-priced families stay visible with their price intact (no silent price loss on toggle)
- [ ] Page states explicitly that it manages PortOS tracking only, not provider billing
- [ ] Tests: route/nav registration (navManifest parity), toggle behavior incl. disabled-price retention, price/tier save path; docs index updated if a new guide page is added

## Out of scope
- Any integration with actual provider billing/subscription management (Stripe/customer portal/plan purchase/cancel at Anthropic/OpenAI/Google/xAI) — PortOS never touches those
- Changing quota-scrape mechanics (`providerUsage.js` fetches) or the savings math (`subscriptionSavings.js`); the page only surfaces them
- Fleet-wide subscription sharing semantics beyond what `PUT /api/usage/fleet-billing` already does
- Migrating the Usage (`/devtools/usage`) or Quota Burn pages themselves; reuse their components, don't rebuild them

atomantic:
## Summary

`/models/subscriptions` — one row per paid AI plan, answering "which subscriptions do I pay for, are they on, and what are they costing me?" in one place. That was previously spread across three: enable/disable lived per provider on the Providers page, the price lived in an editor buried inside the Usage report, and quota lived on a third card.

Each row carries:

- **Family-level enable/disable**, defined as flipping `enabled` across that family's provider records — the same flag `resolveEnabledFamilies` reads, so the toggle and the savings block cannot disagree. Writes are sequential: each is a read-merge-write of the same `providers.json`.
- **Plan tier** ("Max 20x", "Pro") in a new `subscriptionPlanTiers` settings map — a separate key, not an overload of the numeric price. It falls back to the plan name PortOS already scrapes out of the CLI, so the Usage pill and this pill can't show different answers for one plan; the stored value is the override, which is what a scrape reading `unknown` needs.
- **Price editing** through the existing `PUT /api/usage/subscriptions`, which now takes independent optional `costs` and `tiers` slices.
- **Spend** from the report's savings block and **quota meters** from the provider scrape.

A priced family keeps its row when every provider is off, or when it has no providers at all, so switching a plan off for a month can't silently lose what it costs. The page states in its own copy that it manages PortOS-side tracking only and never touches vendor billing.

Reachable by URL, sidebar, ⌘K and voice (`nav.models.subscriptions`).

### Rolled in

The review pass replaced everything the new page had duplicated with the Usage page's own components — several had already drifted at birth:

- `ProviderQuotaBody` — the quota state ladder. The copy had inverted two branches and dropped the `metrics[]` case, the branch that exists for a backend whose quota can't be queried at all.
- The money cells, price input and "not priced" fallback now come from `SubscriptionSavingsCard`, which already carried a comment about having defined them once for this reason.
- `UsageMeter` draws through `ui/ProgressBar`, gaining the `role="progressbar"` + `aria-valuenow` trio the hand-rolled track never had.
- Report windows move to `lib/usagePeriods.js` — the two copies had disagreed on the default, so a `?period`-less URL meant 7 days on one page and 30 on the other.
- `resolveEnabledFamilies` derives membership from `familyForProvider` instead of keeping a second copy of the matcher and the six local-runtime exclusion flags.
- The absent-vs-cleared merge contract prices and tiers share lives once in `lib/familySettingsMap.js`.

A period click no longer respawns the multi-second CLI quota scrape for figures that don't move with the window; a tier-only edit no longer rebuilds the cost report and fleet breakdown; price and tier save in one request, and both write handlers return the refreshed rows, so a save or toggle costs one round trip.

No migration: `subscriptionPlanTiers` is additive and absent normalizes to `{}`.

## Test plan

- `cd server && npm test` — 44,453 passing
- `cd client && npm test` — 11,911 passing, 973/973 files
- `cd client && npm run lint` — clean
- `cd client && npm run build` — clean
- New coverage: nav/route registration (via the Models tab-completeness guard and navManifest shape tests), family toggle including disabled-price retention, the price/tier save path, the two efficiency regressions above, and `ProviderQuotaBody`/`UsageMeter` behavior.

Closes #7403


atomantic:

PR #7428, opened 2026-09-16
fix: keep an undeclared local model id out of the hosted comparison rules

### LINKED ISSUE #7423: Local model ids without a declared benchmarkModel still fall through to the hosted name rules
After #PR (the `benchmarkModel` declaration), a local install id that its catalog entry does NOT declare still reaches the textual normalization in `server/lib/comparisonModelScope.js`, which was written for hosted provider ids and reads a GGUF repo name wrong.

## Evidence

```
$ node --input-type=module -e "import { catalogSlugForProviderModel } from \"./server/lib/comparisonModelScope.js\"; console.log(catalogSlugForProviderModel(\"lmstudio-community/LFM2.5-1.2B-Thinking-GGUF\"))"
lfm2.5-1.2b
```

`-Thinking` is model identity here, not a reasoning-mode suffix, so the id resolves to the name of the models NON-thinking sibling. `SUFFIXES` also contains `reasoning`, `max`, `high` and `instruct`, each of which appears inside real GGUF repo names for the same reason.

## Impact

Inert in the shipped seed, which carries no `lfm2.5-*` rows — but reachable today by a supported action. `POST /api/providers/comparison/sync-aa` pulls the full public index (~640 models), after which such a name resolves and the wrong models score, price and throughput plot under the locally installed model, with no in-chart signal. This is the cross-endpoint transfer `docs/MODEL-COMPARISON.md` forbids.

## Chosen fix

Do not route a local-backend id through the hosted rules at all. `catalogSlugForProviderModel` already consults `LOCAL_INSTALL_IDS` first; extend that to recognize an id as *local-shaped* (present in the catalog under either backends normalization, regardless of whether its entry declares a `benchmarkModel`) and return `""` when the entry declares none, instead of falling through.

That keeps the declared path exact and makes the undeclared path silent, which is the same contract the entry comment already states. Widening declarations is a separate, per-entry review and should not be bundled here.

## Scope

- `server/lib/comparisonModelScope.js` — the fall-through branch.
- `server/lib/comparisonModelScope.test.js` — assert an undeclared local entry resolves to `""` (the existing "gives a local build no benchmark identity unless its entry declares one" test currently asserts only `not.toBe(<wrong name>)`; tighten it to `toBe("")`).

## Out of scope

Declaring `benchmarkModel` for more entries, and the 32 catalog entries whose two backend ids currently normalize differently — both are only reachable once an entry opts in, and the catalog-wide guard in `scripts/prune-model-comparison-seed.test.js` already covers every entry that does.

atomantic:
## Summary

`catalogSlugForProviderModel()` only routed a local install id through the local-catalog lookup when its `localLlmCatalog.js` entry declared a `benchmarkModel`. An entry with no declared benchmark (e.g. an LM Studio GGUF repo whose id contains `-Thinking`, `-Reasoning`, `-High`, etc.) fell through to the textual normalization rules written for hosted provider ids — those rules strip suffixes like `-thinking`/`-reasoning`/`-high` as if they were reasoning-mode markers, when for a GGUF repo name they're part of the model's identity. That let an undeclared local id resolve to a *different* model's benchmark name once a full Artificial Analysis sync (`POST /api/providers/comparison/sync-aa`) populated the catalog with a same-stemmed hosted row — silently mixing that model's price/throughput into the locally-installed model's chart, with no in-chart signal (the cross-endpoint transfer `docs/MODEL-COMPARISON.md` forbids).

## Fix

- `isLocalCatalogId()` recognizes any id the local-LLM catalog matches at all — declared or not — on either backend, and short-circuits `catalogSlugForProviderModel` before it ever reaches the hosted rules. An undeclared entry now resolves to `''` (no slug) instead of a wrong one.
- The LM Studio side of that match is gated on a real local-build marker (`gguf`/`mlx`/a known quantization tag) actually present in the incoming id. Without that gate, the loose vendor-stripped LM Studio normalization would misclassify a bare hosted `vendor/model-name` id that happens to share a model name with an installed `vendor/Model-Name-GGUF` build (e.g. `google/gemma-4-31b-it` vs. the catalog's `lmstudio-community/gemma-4-31B-it-GGUF`) — confirmed as a real regression against the existing test suite before adding the gate.
- The Ollama side needs no such gate: Ollama tags are already an exact normalized match against the catalog's own ids, with no vendor-stripping ambiguity.

## Test plan

- Tightened the existing "gives a local build no benchmark identity unless its entry declares one" test from `not.toBe(<wrong name>)` to `toBe('')`, and added the issue's exact repro (`lmstudio-community/LFM2.5-1.2B-Thinking-GGUF`).
- `cd server && ./node_modules/.bin/vitest run lib/comparisonModelScope.test.js` — 13/13 pass.
- Full server suite (`cd server && vitest run`) — 44,493 passed, 26 skipped, 0 failed.
- Manually verified the issue's repro now returns `''`, declared local ids (`ornith:35b`, `qwen3-coder:30b`) still resolve correctly, and hosted ids (`us.anthropic.claude-sonnet-5`, `google/gemma-4-31b-it`) are unaffected.

Closes #7423


atomantic:

Issue #7402, opened 2026-09-15
Quality checks runner provider selector omits capable process providers with no subscription family (e.g. OpenCode MTPLX TUI)


atomantic:
**Context.** The quality checks runner (an app's Quality page, "Run quality checks" section) lets you pick which provider's agent runs the selected audits, but its provider selector omits capable process providers that have no *subscription family* — e.g. an OpenCode MTPLX TUI. The server accepts any provider (`POST /api/cos/schedule/maintenance-runs` validates `providerId` as a non-empty string only — `server/routes/cosScheduleRoutes.js:38-51`), so the restriction is purely client-side.

**Root cause.** The picker eligibility filter is `provider.enabled && isProcessProvider(provider) && familyForProvider(provider)` in two places:

- `client/src/components/apps/AppQualityRunner.jsx:11` (`eligibleProvider`, passed to `useProviderModels` at :34)
- `client/src/components/cos/tabs/schedule/MaintenanceRunForm.jsx:34` (same filter in the runner-settings form)

`familyForProvider` (`server/lib/providerFamilies.js:63-66`) is a *subscription-quota attribution* helper: it returns `null` for backend-backed providers (`ollamaBacked`/`lmstudioBacked`/`mtplxBacked`/`llamaBacked`/`vllmBacked`/`sglangBacked`) and for any vendor not in the four-family list (claude/codex/agy/grok) — the module docstring explicitly notes opencode/kimi/cursor have "no subscription quota to measure." So every process provider that is not one of the four subscription families (OpenCode/MTPLX TUI, Kimi, Cursor, local-model wrappers) is silently dropped from the selector. The picker is conflating "has an attributable subscription" with "can run an agent process."

**Fix (chosen).** In both files, replace the family-gated filter with the standard process-provider eligibility — `enabledProcessProviderFilter` from `client/src/utils/providerTypes.js:79` (`enabled && isProcessProvider`, re-exported via `utils/providers.js`) — and drop the now-unused `familyForProvider` import. This matches how other agent pickers enumerate process providers, and the server already accepts the selection. `familyForProvider` stays where it is for its real job (subscription cost attribution via `resolveEnabledFamilies`) — it is just no longer a capability gate. Update `client/src/components/apps/AppQualityRunner.test.jsx` and the MaintenanceRunForm test to assert an enabled family-less process provider (e.g. an opencode TUI record) is offered and a disabled one is not.

**Rejected alternatives.**

[...]

PR #7429, opened 2026-09-16
fix: let a maintenance run dispatch on a provider with no subscription family

### LINKED ISSUE #7416: Maintenance-run dispatch still requires a subscription family, blocking the #7402 provider picker fix
**Context.** #7402 made the Quality-page runner and CoS schedule maintenance-run form's provider picker offer every enabled process provider (`enabledProcessProviderFilter`), not just the four subscription families — matching the picker convention already used by `PullRequestsTab`, `IssuesTab`, `SlashDoPanel`, and `SlashDoRunDrawer`. Those other pickers dispatch through `createSlashdoTask`, which pins directly to the chosen provider with no family requirement. The Quality-runner / maintenance-run picker dispatches through a different path — `startMaintenanceRun` → `invokeQuotaBurnStep` — that still hard-requires a subscription family end to end. Picking a family-less provider (an OpenCode TUI, a local-model wrapper) now shows up in the picker but fails at "Run now" with `MAINTENANCE_RUN_PROVIDER_UNAVAILABLE` ("not an enabled subscription CLI/TUI provider").

**Root cause — three independent family gates, not one.**

1. `server/services/maintenanceRun.js` computes `familyId = familyForProvider(provider)` and only calls `resolveBurnProvider` when `familyId` is truthy — `startMaintenanceRun` (lines ~147-149, and the claim-handler branch ~154-156) and `updateMaintenanceStep` (~217-219). A family-less provider always gets `pinned = null` and a 400.
2. Even bypassing (1), `server/services/quotaBurnInvoke.js#resolveStepProvider` (~308-320) re-checks `matchesFamily(provider, family.id)` before accepting the resolved provider — `matchesFamily` returns `false` immediately when `family.id` is falsy (`needle = ''`), so a null-family job is refused here too when the step actually dispatches.
3. Even bypassing (1) and (2), `server/services/scheduledHandlers/providerPick.js#providerForFamily`'s `available` list excludes every local-runtime-backed provider (`ollamaBacked`/`lmstudioBacked`/`mtplxBacked`/`llamaBacked`/`vllmBacked`/`sglangBacked`) regardless of an explicit `providerId` pin — so "local-model wrappers," which #7402's acceptance criteria explicitly wants offered, would still be rejected even after (1) and (2) are fixed.

**Why this isn't a one-line relaxation.** `family` flows into `invokeQuotaBurnStep`'s `runBuiltinTaskStep`/`runProgrammaticStep`/`runCustomJobStep` as `family.id`, which is used for on-demand-task burn provenance (`burn: { family: family.id, ... }`) and console logging — a family-less job needs either a synthetic non-crashing family identity or those call sites updated to tolerate `family: null`. `maintenanceRun.js`'s own docstring says the family is what "makes [the task] cooldown-exempt and lets an observed refusal be credited to the right window" — for a provider with no subscription window, that exemption/credit concept doesn't apply, but the code doesn't currently express "no family" as a legitimate state; it only expresses "family not yet resolved" (an error). `providerForFamily`/`resolveBurnProvider` are also shared with the AUTOMATIC quota-burn sweep (`quotaBurnRunner.js`), so any relaxation must not weaken the automatic burn's guarantee that a step only ever spends its own family's window — this needs a deliberate "explicit pin, no family required" mode threaded through the manual-run path only, not a global loosening of `providerForFamily`.

**Fix shape (not yet designed in detail).** Likely a `family: null` / "unfamilied" convention recognized explicitly by `maintenanceRun.js`, `quotaBurnInvoke.js`, and `providerPick.js`'s `providerForFamily`, gated so it only ever activates for a MANUAL maintenance run naming an explicit, enabled, process-capable `providerId` (never for the automatic burn sweep, which has no explicit pin and must keep resolving strictly by family). Update `maintenanceRunId`/provenance handling so a null-family manual run doesn't get misattributed to a real subscription's cooldown/window bookkeeping.

**Acceptance criteria.**

- Clicking "Run now" (Quality-page runner or CoS Schedule maintenance-run form) with an enabled process provider that has no subscription family (e.g. OpenCode TUI, an Ollama/LM Studio-backed wrapper) actually dispatches instead of failing with `MAINTENANCE_RUN_PROVIDER_UNAVAILABLE`.
- The automatic quota-burn sweep (`quotaBurnRunner.js`) is unaffected — it still refuses to spend a provider outside its plan's named family.
- `updateMaintenanceStep`'s per-step provider edit accepts the same family-less providers.
- Server tests cover: a family-less enabled process provider starting a run, editing a step to a family-less provider, and a regression test that the automatic burn sweep still rejects an out-of-family / family-less provider.

Discovered while implementing #7402; the client-side picker fix there ships regardless (it matches the established `enabledProcessProviderFilter` convention other manual-run pickers already use) — this issue is the server-side follow-through so "Run now" actually works for the providers the picker now offers.

atomantic:
## What shipped

"Run now" from the Quality-page runner or the CoS Schedule maintenance-run form failed with `MAINTENANCE_RUN_PROVIDER_UNAVAILABLE` whenever the picked provider belonged to none of the four subscription families — an OpenCode TUI, an Ollama/LM-Studio-backed wrapper — even though #7402 made the picker offer exactly those. Three independent gates each hard-required a family:

1. `maintenanceRun.js` only called `resolveBurnProvider` when `familyForProvider()` returned a family, so a family-less pin always produced `pinned = null` and a 400.
2. `quotaBurnInvoke.js#resolveStepProvider` re-checked `matchesFamily(provider, family.id)`, which returns `false` for a falsy family id.
3. `providerPick.js#providerForFamily` excluded every local-runtime-backed provider regardless of an explicit `providerId` pin.

A manual run now expresses "spends no subscription window" as a legitimate state rather than as a failed resolution:

- **`maintenanceRun.js`** builds the dispatch identity through `manualBurnFamily()` — `{ id: familyId }` when the provider has a family, `{ id: null, unfamilied: true }` when it does not. This is the only place that identity is constructed.
- **`providerPick.js`** recognizes it via the new `isUnfamiliedBurn()`, and `providerForFamily` lifts the local-runtime exclusion only under that flag *and* an explicit pin. The API-type and `enabled` + `isProcessProvider` requirements are untouched.
- **`quotaBurnInvoke.js`** skips the family re-check for an unfamilied run, and omits `family` from the burn block rather than writing a placeholder.
- **`quotaBurnOrigin.js`** now accepts `stepId` + (`family` **or** `maintenanceRunId`) as attribution. Without this the whole burn block failed to normalize, `triggerOnDemandTask` refused the request, and the run never paced — the `maintenanceRunId` its evaluation looks for never reached the task. `hasQuotaBurnProvenance` accepts either key, so the ladder stays cooldown-exempt.

Because the family is omitted rather than faked, `recordBurnAgentCompletion` and the plan walk both short-circuit: an unfamilied run's refusals and completions are never credited to a subscription it never touched.

**The automatic quota-burn sweep is unaffected.** It hands `providerForFamily` a plan's family record and can never construct the unfamilied identity, so it still resolves strictly by family and refuses an out-of-family or local-runtime pin. An automatic burn also still requires a `family` in its provenance, since it never carries a run id.

Editing one stage of a run onto a family-less provider works too: a step's family is now read by key **presence**, so a legitimately `null` one is a decision rather than "inherit the run's family" — which would then have refused the pin for not belonging to it. Same fix applies to a family-less claim handler beside familied audits.

## Test plan

New tests, each verified to fail against the pre-fix code (10 failures on the reverted sources):

- `providerPick.test.js` — an unfamilied explicit pin resolves a family-less provider including a local-runtime wrapper; still refuses disabled / unregistered / API-type; resolves nothing without a pin; relaxes nothing for a familied caller.
- `quotaBurnInvoke.test.js` — an unfamilied step dispatches on the pinned provider carrying `maintenanceRunId` and no `family`; still refuses disabled/absent/non-process pins; **regression:** the automatic sweep still refuses an out-of-family and a local-runtime pin.
- `maintenanceRun.test.js` — starts a run on a family-less provider and dispatches it unfamilied; a per-step edit onto one does not inherit the run's family; a family-less claim handler works while the audits stay on their own family. Its `providerPick` double was replaced with the real resolver over a provider fixture, so these assert the shared gate rather than a stand-in.
- `quotaBurnOrigin.test.js` — a family-less manual block normalizes and stays cooldown-exempt while the plan does not own it; a block with neither family nor run id is still refused.

Suites run: full server suite (2221 files, 44505 tests) green; full client suite green (untouched by this change).

Closes #7416


atomantic:

Issue #7414, opened 2026-09-15
Split Runtimes out of Models → LLMs into its own top-level tab


atomantic:
## Summary

Models → LLMs (`/models/llms`) currently bundles three pill views — Runtimes, Model Library, Abuse Guard — with Runtimes as the default view. Runtimes is where an install manages its local model servers (start/stop/configure Ollama, LM Studio, llama.cpp / llama-server, MTPLX, Slotstream; vLLM and SGLang have setup/start controls through the AI Providers readiness UI but no dedicated rows on the runtime page yet) plus the server-side downloads that attach to them (speculative-decoding drafts, MTPLX and Slotstream checkpoints). That is a different concern from Model Library, the Ollama/LM-Studio weights catalog, and should get its own top-level tab in the Models section — a sibling of LLMs, so the breadcrumb becomes "Models → Runtimes". LLMs keeps what is about weights: Model Library and the Abuse Guard subpage.

## Context

- `server/lib/navManifest.js:318` — `nav.settings.local-llm`, path `/models/llms`, label LLMs, `tabId: 'llms'`, `previousPaths: ['/settings/local-llm']`.
- `server/lib/navManifest.js:319` — `nav.models.llms.abuse` at `/models/llms/abuse`: Abuse Guard keeps its pill under LLMs **and** has its own NAV entry (breadcrumb "Models → LLMs → Abuse Guard"). Runtimes differs: it becomes a top-level tab (it is the current default view, not a drill-down).
- `client/src/components/settings/LocalLlmTab.jsx` — `LLM_VIEWS = [runtimes (default), library, abuse]`; dispatcher mounts `LocalLlmRuntimesView` / `LocalLlmLibraryView` / `ModelAbuseGuardPanel`; `LLM_NAV_SUBROUTES = [abuse]` is scraped by `server/lib/navManifest.test.js`.
- `client/src/pages/Models.jsx` — tab router; the `llms` tab mounts `LocalLlmTab` with `view` from the `:recordId` param, defaulting to `runtimes`.
- The views are not fully disjoint — Model Library still installs/starts Ollama and LM Studio, and Runtimes exposes lifecycle controls for them too; both call `useLocalLlmStatus` (`LocalLlmRuntimesView.jsx:167`, `LocalLlmLibraryView.jsx:127`) and watch `localLlm:progress`. The server lifecycle surface and the weights catalog remain separate; the shared status hook stays as-is.
[...]

PR #7430, opened 2026-09-16
fix: stop TabPills advertising aria-controls for panels that never mount

### LINKED ISSUE #7420: Route-hosted section panels emit dangling aria-controls / unnamed tabpanels
## Summary

`TabPills` sets `aria-controls="\${controlsIdPrefix}-\${t.id}"` on **every** tab button, but a tab bar only ever has **one** panel mounted. Every inactive tab therefore points at an IDREF that is not in the document. Separately, the three pages hosted by `RouteTabsHeader` each handle their panel's accessible name differently, and one has none at all.

This is pre-existing — #7414 hit one instance of it (the Runtimes panel's `aria-labelledby` pointed at `tab-runtimes`, an id `RouteTabsHeader` never renders, because it passes no `controlsIdPrefix`) and fixed that panel by giving it its own heading. The general contract was left undecided.

## Evidence

- `client/src/components/ui/TabPills.jsx:234`, `:288` — `aria-controls` is set unconditionally whenever `controlsIdPrefix` is passed, not only for the active tab. So with `LocalLlmTab` (`controlsIdPrefix="llm-management-panel"`) showing Model Library, the Abuse Guard tab advertises `aria-controls="llm-management-panel-abuse"` for an element that does not exist — that dispatcher mounts one view at a time (`LocalLlmTab.jsx:50-51`).
- `client/src/components/ui/RouteTabsHeader.jsx` passes no `controlsIdPrefix`, so its tabs carry neither `id` nor `aria-controls`. Its three hosts diverge:
  - `client/src/components/settings/LocalLlmRuntimesView.jsx` — `role="tabpanel"` + its own `sr-only` heading (the #7414 fix; correct today).
  - `client/src/pages/ApiExplorer.jsx:515` — bare `role="tabpanel"`, no accessible name, no link to any tab.
  - `client/src/pages/SystemHealthPage.jsx`, `client/src/pages/Sharing.jsx` — no tab semantics on the panel at all.

## Chosen fix

Emit the relationship only where it is real, and give route-hosted panels one contract:

1. In `TabPills`, set `id` / `aria-controls` only on the **active** tab. An inactive tab in a one-panel-at-a-time bar has nothing to control, and a dangling IDREF is worse than an absent attribute.
2. Leave `RouteTabsHeader` without `controlsIdPrefix`. Making it pass one would replace a single dangling reference with N-1 of them — route tabs navigate, they do not reveal a sibling panel that is already in the DOM.
3. Standardize route-hosted panels on the #7414 shape: keep `role="tabpanel"`, name it with the panel's own heading (`sr-only` where the section header already shows the name). Apply to `ApiExplorer.jsx`, and add the role+heading to `SystemHealthPage.jsx` and `Sharing.jsx` so all four agree.
4. Add a client guard test asserting that every `aria-controls` / `aria-labelledby` rendered by the section-tab pages resolves to an element in the document — a dangling IDREF is invisible to every existing assertion, which is why this survived.

Rejected alternative: dropping tab semantics from route bars for `role="navigation"`. It is a larger change and discards the keyboard model `TabPills` already implements.

## Acceptance criteria

- [ ] No `aria-controls` or `aria-labelledby` in the Models, Settings, API Explorer, System Health or Sharing section shells points at an absent element.
- [ ] All four `RouteTabsHeader` hosts name their panel the same way.
- [ ] A guard test fails on a newly-added dangling IDREF.
- [ ] `cd client && npm test` passes.

Refs #7414

atomantic:
## Summary

- `TabPills` set `aria-controls` on every tab in a bar, even though a one-panel-at-a-time host (e.g. `LocalLlmTab`) only ever mounts the active tab's panel — every inactive tab pointed a screen reader at an id that does not exist. Now only the ACTIVE tab gets `aria-controls`. `id` stays on every tab regardless of active state, because a bar that keeps every panel mounted and toggles `hidden` (`LoomSeriesPlan`, `PipelineSeries`) gives each panel a static `aria-labelledby="tab-<id>"` that would otherwise dangle.
- `RouteTabsHeader` intentionally passes no `controlsIdPrefix` (route tabs navigate, they don't reveal a sibling panel already in the DOM), so its four hosts now agree on how their panel names itself: `ApiExplorer`, `SystemHealthPage`, and `Sharing` get the same `role="tabpanel"` + own `sr-only` heading shape `LocalLlmRuntimesView` already used after #7414.
- Adds a reusable guard assertion (`client/src/test/ariaRefAssertions.js`) asserting every rendered `aria-controls`/`aria-labelledby` resolves to a real element in the document, wired into the `TabPills`, `ApiExplorer`, `SystemHealthPage`, and `Sharing` test suites.

## Test plan

- [x] `cd client && npm test` — full suite passes (973 files, 11926 tests, 7 pre-existing skips)
- [x] `cd client && npm run build` — builds clean
- [x] `cd client && node_modules/.bin/biome lint --error-on-warnings <changed files>` — clean
- [x] Verified the new `Sharing` guard test fails against the pre-fix code (no `role="tabpanel"` existed), confirming it's a meaningful regression check

Closes #7420


atomantic:

atomantic:
Correction: the configured reviewer (provider:opencode-zen-cli) is tagged `~opt~` (optional) for this run. Its transport was unavailable ("no enforced tool-free review transport"), which is non-blocking for an optional reviewer, so this is not gated on that review completing. Proceeding with CI + merge.

PR #7492, opened 2026-09-16
add generic credential-bootstrap provider config for CLI/TUI harnesses

atomantic:
## Summary

Adds generic support for an AI provider whose harness auth is provisioned by an external CLI at spawn time — e.g. a company-internal tool that mints a short-lived token for a proxy and execs the harness itself — instead of PortOS storing a static `apiKey`.

A provider can now carry:

```json
"credentialBootstrap": {
  "setupCommand": "npm install -g @your-org/token-cli",
  "command": "token-cli",
  "args": ["run"],
  "harnessId": "claude-code",
  "argsSeparator": "--"
}
```

PortOS spawns `<command> <args...> <harnessId ?? command> [<argsSeparator>] <harness args...>` in front of the harness invocation, generalizing the shell idiom `<bootstrap> run <harness> -- <args>`. `setupCommand` is advisory only — shown to the user as a one-time step, never executed by PortOS.

## Changes

- **`server/lib/credentialBootstrap.js`** (new) — the pure wrap logic (`hasCredentialBootstrap`, `applyCredentialBootstrap`) plus `resolveCliSpawn`, which composes the wrap with Windows shim resolution for the three headless call sites that repeated that sequence.
- **Six spawn call sites** (`runner.js`, `cliProviderRun.js`, `agentCliSpawning.js`, `agentTuiSpawning.js`, `tuiPromptRunner.js`, `visionCli.js`) apply the wrap right before the literal `spawn`/`pty.spawn` call, while every existing consumer keeps resolving the harness's own `command`/`args` unchanged for vendor detection, prompt-delivery convention, and TUI ready-text heuristics.
- **`agentLifecycle.js`** — a credential-bootstrap-configured provider is always routed through the direct in-process spawner, never the out-of-process CoS runner, whose static command allowlist has no way to recognize an arbitrary bootstrap binary.
- **Schema** — `credentialBootstrap` added to both the toolkit and PortOS-side provider schemas (kept in parity per existing convention), nullable so the editor can explicitly clear a previously-set config.
- **UI** — a "Credential Bootstrap" section in the provider editor's Connection tab (Setup Command, Bootstrap Command, Bootstrap Args, Harness ID, Args Separator).

Public-review (contributor-code-review) postures are explicitly excluded from the wrap — that enforced recipe is the sandbox and must not be tampered with by a user-configured wrapper.

## Test plan

- [x] `server/lib/credentialBootstrap.test.js` — pure wrap/compose logic, including `harnessId` and `argsSeparator`
- [x] `agentTuiSpawning.test.js` — TUI config keeps `command`/`args` as harness identity while `spawnCommand`/`spawnArgs` carry the wrap; public-review posture never wraps
- [x] `agentLifecycle.postureGate.test.js` — a credential-bootstrap provider is routed direct-only even with runner mode on, verified against a reverted-fix failure
- [x] `runner.test.js` — the bootstrap CLI is spawned in front of the harness, with the harness's own prompt-delivery convention preserved
- [x] `ProviderForm.test.jsx` — form fields fold into the nested schema shape on save, and clearing the Bootstrap Command sends an explicit `null`
- [x] Full server suite (2252 files) and client suite (980 files) pass


atomantic:
Reviewed by /do:review after merge. The review's fixes landed as #7497 (this PR merged while the review was running): public-review postures were still wrapped on the headless spawn sites, and four provider spawn sites (Ask, OpenCode task runner, TUI /usage scrape, Launch in Shell) ran the bare harness. Process-group kill for forking wrappers is deferred to #7496.

PR #7507, opened 2026-09-16
offer Refresh Models on Kilo and Grok provider cards

### LINKED ISSUE #7505: Refresh Models is missing on Kilo and Grok provider cards — the harness-catalog route hardcodes opencode
## Problem

Kilo and Grok CLI/TUI wrappers draw their models from the harness's own catalog — `harnesses.test.js` ("the shipped records a harness refresh may rewrite") pins `kilo-cli`, `kilo-tui`, `grok-cli`, `grok-tui` in that list, and `refreshHarnessModels('kilo' | 'grok')` refreshes them correctly from the **Harnesses** page.

Their **provider cards show no Refresh Models button at all.**

`presentProvider` derives the flag as:

```js
canRefreshModels: decorated.canRefreshModels || refreshesOpenCodeCatalog(provider)
```

- `decorated.canRefreshModels` comes from the toolkit's `MODEL_FETCHERS` table (`server/lib/aiToolkit/internal/modelFetchers.js`), which has no `kilo` or `grok` row — correct, neither has a toolkit fetcher.
- `refreshesOpenCodeCatalog` (`server/routes/providers.js:148`) hardcodes `providerRuntimeKey(provider) === 'opencode'`.

So the harness-catalog path is wired for exactly one vendor, while the service behind it (`refreshHarnessModels`) is generic over every runtime declaring `modelsArgs` (`server/services/providerRuntimeInstaller.js`): pi, opencode, kilo, grok, antigravity, cursor.

## Fix

Replace the vendor-string test with the general question — resolve the provider's runtime and ask whether it lists models:

```js
const refreshesHarnessCatalog = (provider) => {
  const key = providerRuntimeKey(provider);
  const runtime = key && PROVIDER_RUNTIMES.find((row) => [row.id, ...row.aliases].includes(key));
  return Boolean(runtime?.modelsArgs) && usesHarnessCatalog(provider);
};
```

Then pass `providerRuntimeKey(stored)` to `refreshHarnessModels` instead of the literal `'opencode'` in the `POST /:id/refresh-models` handler.

## Decide first

A record can match BOTH a `MODEL_FETCHERS` row and the harness-catalog path — `cursor-cli` and `antigravity-cli` already do, and `MODEL_FETCHERS` keys partly on the user-editable display name (`cliNameMatch`), so a user-renamed Kilo record could be claimed by an unrelated row. **Decision to make in the PR: the toolkit fetcher wins when both match** (it is the stronger, per-record signal), and the harness-catalog path serves only records no fetcher claims. Order the `||` accordingly and pin it with a test.

## Why this was not done in #7506

The hardcode was what BOUNDED the credential-bootstrap bug that PR fixed: before per-credential probe bucketing landed, widening this predicate would have spread the ambient-account catalog overwrite to Kilo and Grok records too. With bucketing in, widening is safe.

## Acceptance

- A `kilo-cli` / `grok-cli` provider card offers Refresh Models, and the click rewrites its own catalog (namespace scope respected).
- A record claimed by a `MODEL_FETCHERS` row still routes to that fetcher.
- A gateway- or local-runtime-backed wrapper still offers no harness refresh (`usesHarnessCatalog` false).

atomantic:
## Summary

- Kilo and Grok provider cards showed **no Refresh Models button**, because the route gated the harness-catalog refresh on a hardcoded `providerRuntimeKey(provider) === 'opencode'` while the service behind it, `refreshHarnessModels`, was already generic over every runtime declaring `modelsArgs` (pi, opencode, kilo, grok, antigravity, cursor).
- Replaced the vendor string with the general question — does this record's runtime list models, and is it a plain wrapper — as `harnessCatalogRuntime(provider)` in `server/services/harnesses.js`, and the handler now passes that runtime's id instead of the literal `'opencode'`.
- The predicate lives in the harnesses service rather than the route so the **seed pin in `harnesses.test.js` uses the production predicate** instead of a second transcription of it. A future widening now has to move that pinned list on purpose.

### Precedence decision

A record can match both the toolkit's `MODEL_FETCHERS` table and the harness-catalog path — `cursor-cli` and `antigravity-cli` do today. **The toolkit fetcher wins**; the harness catalog serves only records no fetcher claims. That preserves where every currently-refreshable record gets its catalog from, and it is pinned by a test.

The hardcode was not an oversight: it bounded the credential-bootstrap catalog overwrite that #7506 fixed. With per-credential probe bucketing in, widening is safe.

## Test plan

- `server/routes/providers.refreshModels.test.js` — new cases: `kilo-cli/tui` and `grok-cli/tui` offer the button and refresh through their own harness; `cursor-cli` / `antigravity-cli` / `pi-cli` (and a record claimed only by the weak display-name column) still route to their toolkit fetcher; a gateway-backed and a local-daemon wrapper never reach the harness path; a custom binary with no runtime row offers no refresh.
- Both mutations are caught: re-narrowing the predicate to `opencode` fails 5 tests; flipping the precedence so the harness wins fails 4.
- Full server suite green — 2256 files, 45194 tests.
- Local review: antigravity (`gemini-3.8-flash-low`, effort low, plan+sandbox) — APPROVE, no findings.

Closes #7505


atomantic:
Rebased onto `main` after #7506 merged; the harness call now carries that PR's `{ providerId }` per-credential scoping on the widened path, pinned by the new tests. Full server suite re-run green after the rebase (45,198 tests).

PR #7519, opened 2026-09-16
test: retry Windows temp cleanup

### LINKED ISSUE #7515: Windows CI: worktreeReap.test.js afterAll rmdir races EBUSY and reds the whole run
## Symptom

On Windows CI, `services/worktreeReap.test.js` fails as a **failed suite** — not a failed test — while every test in it passes:

```
FAIL  services/worktreeReap.test.js > hasBranchMergeEvidence
Error: EBUSY: resource busy or locked, rmdir 'C:\Users\<runner>\AppData\Local\Temp\portos-reap-XXXXXX'
```

Observed on run `35156591405`, job `Windows server unit tests (2/3)`. A re-run of the **same commit** passed clean, confirming it is timing, not the change under test.

## Impact is much larger than one suite

The job's `Cancel sibling CI jobs after failure` step fires on it, so one `rmdir` race cancels ~11 sibling jobs (Client 1-3, Server 1-2, DB, Windows 1-3) and both gate jobs report `failure`. The whole PR run reads red, and the 11 cancellations look like real failures until you check each job's steps. That is the expensive part: the signal points everywhere except the one suite that actually broke.

## Cause

`server/services/worktreeReap.test.js:82`:

```js
afterAll(async () => { await rm(dir, { recursive: true, force: true }); });
```

`force: true` suppresses **ENOENT only** — it does not retry `EBUSY`/`EPERM`. On Windows a handle on the temp git repo can outlive the last `execGit` call (git's `fsmonitor--daemon`, the search indexer, or AV), so the `rmdir` races process teardown. `fs.promises.rm` has `maxRetries` / `retryDelay` for exactly this case — Node's docs list `EBUSY`, `EMFILE`, `ENFILE`, `ENOTEMPTY` and `EPERM` as the errors it retries with a backoff.

An `afterAll` throw fails the suite, which is why an all-green suite reports `FAIL`.

## Fix

Add retries at the four temp-root removals in that file — lines 82, 163, 164 and 377:

```js
await rm(dir, { recursive: true, force: true, maxRetries: 10, retryDelay: 100 });
```

Worth extracting as a shared `removeTempDir(path)` in `server/lib/testHelper.js` instead of repeating the options four times: any suite that builds a real git repo under `os.tmpdir()` has the same exposure on Windows, so the next one should get it by default rather than by remembering.

## Out of scope

Making the CI job's fail-fast distinguish a hook failure from a test failure. The cancellation behavior is correct; this issue is the cause it fired on.

hizlidepoo:
## Summary
- add bounded `fs.promises.rm` retries to the four temp-root removals in `worktreeReap.test.js`
- keep the change isolated to the Windows EBUSY cleanup race described in #7515

Fixes #7515

## Verification
- `git diff --check`: PASS after rebase onto current `main`
- targeted Vitest: NOT VERIFIED locally because this WSL environment resolves `npm` to the Windows Node installation, which cannot use the WSL UNC working directory
- previous exact head `4233ff975f33b67847280120e603c824152a10d9`: CI #15156 SUCCESS and maintainer APPROVED
- previous rebased heads had GitHub Actions runs ending `action_required` before jobs ran; not reported as PASS/FAIL
- current rebased exact head `f68478fd6ae0f7845ceeae6cf0f3a331233bbd03`: fresh CI pending

Exact upstream base: `b8ecb73caf900ae8b111f6df1803dcb9998858e4`
Exact head: `f68478fd6ae0f7845ceeae6cf0f3a331233bbd03`


atomantic (approved):
✅ **Approved**

This change adds retries to `fs.promises.rm` calls in the `worktreeReap` test to address a Windows CI flakiness (EBUSY errors). The change is isolated to the test file and targets a known race condition. CI should confirm the fix.

**Scope:** test: add retries to Windows temp dir cleanup

#### Test evidence
- ⏭️ `npm test -w server` — This is a test-only change; the core server tests do not validate this fix. Windows CI runs will confirm the fix.

- The retries are added to the correct `rm` calls in `server/services/worktreeReap.test.js`.
</details>

atomantic (approved):
✅ **Approved**

The change adds retries to `fs.promises.rm` to address a Windows CI issue with `EBUSY` errors during temporary directory cleanup. The linked issue and verification steps confirm the problem and the fix's target.  The fix is isolated to the identified issue and uses existing Node.js retry mechanisms.

**Scope:** Test suite: adds retries to temporary directory cleanup

#### Test evidence
- ⏭️ `npm test -w server` — The author notes that local testing is not possible due to the Windows Node installation environment in the current WSL setup. CI results are pending and will provide the necessary verification.

- The code change adds `maxRetries` and `retryDelay` options to the `rm` calls as described in the summary.
- The retries are applied to the four identified `rm` calls within the test file.
</details>

PR #7543, opened 2026-09-17
Make voice model provisioning provider-agnostic and default to Ollama

### LINKED ISSUE #7541: Voice bootstrap walks the whole LM Studio install chain when the API server is simply unreachable
## Problem

When LM Studio's API server is not running, voice bootstrap walks its **entire** 4-entry model install chain on every server boot, shelling out to `lms get -y <model>` four times, each failing with the same single environmental cause, then gives up with advice that does not address the actual fault:

```
🎙️  voice: installing fast tool-capable model <model-1> via lms get (this may take a few minutes)
🎙️  voice: <model-1> unavailable (Error: Failed to start or connect to local LM Studio API server.) — trying next
🎙️  voice: <model-2> unavailable (Error: Failed to start or connect to local LM Studio API server.) — trying next
🎙️  voice: <model-3> unavailable (Error: Failed to start or connect to local LM Studio API server.) — trying next
🎙️  voice: <model-4> unavailable (Error: Failed to start or connect to local LM Studio API server.) — trying next
🎙️  voice: exhausted install chain <model-1>, <model-2>, <model-3>, <model-4> — set voice.llm.model explicitly in Settings
```

Observed on every boot of a live install. Four 30-minute-timeout subprocesses are armed to chase a fault none of them can fix, and the remedy printed ("set `voice.llm.model` explicitly") is wrong — pinning a model id does not make an unreachable API server reachable.

## Root cause — absent vs. legitimately-empty collapse

`server/services/voice/bootstrap.js:250-255`:

```js
const listLmStudioModels = async () => {
  const res = await fetch(`${LMS_BASE()}/v1/models`, { signal: AbortSignal.timeout(5000) }).catch(() => null);
  if (!res?.ok) return [];          // <-- unreachable backend returns the same [] as an empty one
  const body = await res.json().catch(() => ({}));
  return (body?.data || []).map((m) => m.id);
};
```

`[]` means both "LM Studio is unreachable, we could not ask" and "LM Studio answered, zero models installed". `ensureToolCapableModel` (`:282`) then finds no `fastCapable` entry, cannot distinguish the two, and concludes it should download one — entering the install loop at `:308`.

This is precisely the failure mode `AGENTS.md` names under **"Sentinel + validate to distinguish 'not set / failed' from 'present-but-empty / valid'"**, and it names `lmStudioManager.js` (`availableModels`: `null = not fetched` vs `[] = cached-empty`, plus `getLastListError()` at `server/services/lmStudioManager.js:252`) as the canonical in-repo example. Voice bootstrap keeps its own `fetch` helper and does not follow it.

`preloadModel` (`server/services/voice/bootstrap.js:363-364`) has the same conflation in a milder form — it reports `skipped: 'no-models'` when the real state is "backend unreachable".

## Decision / chosen fix

1. **Give `listLmStudioModels` a sentinel.** Return `null` when the fetch fails or the response is not ok; keep `[]` for a successful response with zero models. Update all three call sites (`:282`, `:314`, `:363`) to branch on `null` explicitly rather than on `.length`.
2. **Short-circuit `ensureToolCapableModel` before the loop.** On `null`, log **one** line naming the real fault and remedy — e.g. `🎙️ voice: LM Studio API server unreachable at <base> — start LM Studio (or set voice.llm.provider) before voice can install a tool-capable model` — and return `{ skipped: 'lmstudio-unreachable' }`. Do not run any `lms get`.
3. **Abort the chain on a connect-class failure.** If a `lms get` inside the loop fails with a start/connect error rather than a model-specific one, break out instead of trying the remaining entries — the cause is shared, so the remaining attempts are known-futile.
4. `preloadModel` returns a distinct `skipped: 'lmstudio-unreachable'` on `null` rather than `'no-models'`.

Do **not** silence the message or gate the whole path behind a setting: an unreachable backend when voice tools are enabled is worth exactly one line per boot.

## Notes

- This also removes unrequested boot-time work: `ensureToolCapableModel` is fired unawaited from voice reconcile (`server/services/voice/bootstrap.js:427`), so today a boot on a machine without LM Studio running arms four multi-GB download attempts the user never asked for.
- Only the log volume and the futile subprocesses change; a reachable LM Studio with no fast tool-capable model installed must still download one exactly as it does now.

## Acceptance criteria

- With LM Studio not running and voice tools enabled, boot emits **one** line naming the unreachable API server and its remedy, and spawns **zero** `lms get` processes.
- With LM Studio running and no fast tool-capable model installed, the install chain behaves exactly as today.
- With LM Studio running and a fast tool-capable model installed, the path still short-circuits on `already-capable`.
- Tests: a case asserting `listLmStudioModels` returns `null` on a failed/not-ok fetch and `[]` on a 200 with an empty `data` array; a case asserting `ensureToolCapableModel` invokes no `lms get` when the list is `null`; a case asserting a connect-class failure mid-chain aborts the remaining entries.

## Dispatch rationale

`model:light` — a well-specified change in one file with the in-repo pattern to copy already named in `AGENTS.md`. `effort:medium` — three call sites plus `preloadModel` must each be re-read to confirm the `[]` → `null` change does not flip an unrelated branch.

atomantic:
## What

Voice could only ever provision models against LM Studio. This makes the provisioning layer pluggable, implements Ollama alongside LM Studio, and makes **Ollama the default**.

Closes #7541.

## Why the runtime path was already fine, and the provisioning path was not

`resolveLlmEndpoint` (`server/services/voice/llm.js`) speaks OpenAI-compatible HTTP to any `type: 'api'` provider with an endpoint — Ollama chat already worked. What was hardcoded is everything *around* it: listing installed models (`/v1/models`), installing a tool-capable one (`lms get`), and pre-warming it (`lms load`, `lms ps`). An Ollama user got none of that, and the LM Studio paths ran against a backend they weren't using.

## Changes

**`server/services/voice/modelProvisioners.js` (new)** — a registry of local backends behind the small surface bootstrap needs: `listModels`, `isToolCapable`, `chain`, `install`, `loadedModels`, `load`. Ollama and LM Studio are implemented. A **remote** OpenAI-compatible provider (OpenAI, Groq, a self-hosted vLLM) resolves to `null`, which correctly means "serves its own models — nothing to install or pre-warm" rather than being an error or being silently provisioned against the wrong daemon.

**Ollama is the new default** (`DEFAULT_VOICE_BACKEND` in `config.js`). It ships a daemon PortOS can start and probe itself (`ollamaManager.ensureRunning`), and it reports per-model tool capability from `/api/show` — so `granite4.1:3b` is known to support tools instead of being missed by an id-substring guess. The capability probe falls back to the id heuristic when it fails, so "unknown" never reads as "no tools". LM Studio remains fully supported.

**Migration 392** pins `voice.llm.provider: 'lmstudio'` for any install that had voice configured but never named a provider. Voice config is a sparse patch merged over `VOICE_DEFAULTS`, so without this an existing install would silently switch backends on upgrade. An install with no stored `voice` config is deliberately left alone — it expressed no preference and is exactly who the new default is for. Gates on the presence of its input, per the migration rule in AGENTS.md.

**Fixes the install-chain storm (#7541).** `listModels` now returns `null` for "could not reach the backend" and `[]` for "reachable, nothing installed" — the Sentinel + validate rule from AGENTS.md. An unreachable backend now reports once and stops, instead of attempting every entry in the chain against one shared failure. Previously, a boot with LM Studio not running logged four failures and armed four 30-minute-timeout multi-GB downloads.

**The chain is derived from `localLlmCatalog`** rather than a hand-maintained copy of LM Studio ids that carried a comment asking the next editor to keep the two in sync.

**`ollamaManager.warmModel`** — the counterpart to the existing `unloadModel`, using the same documented `keep_alive` mechanism.

## A latent bug the tests caught

Deriving the chain from the catalog surfaced an ordering bug. Catalog `params` below a billion are written in **millions** (`'270M'`), and the id-based `sizeOf` only understands a `B` suffix — so it scored `270M` as `Infinity` and sorted the smallest, fastest function-calling specialist **last** in the install chain. Added `paramsToB` with a regression test.

Also fixed while touching it: `OLLAMA_HOST` is conventionally a bare `host:port` with no scheme (that is how `.env.example` documents it), so the endpoint goes through the existing shared `normalizeOpenAiBaseUrl` rather than concatenating `/v1` onto it and producing an unfetchable `localhost:11434/v1`.

## Behavior preserved

- An install that pinned a local backend still falls back to **that** backend if its provider record goes missing — a `lmstudio` install is never silently redirected to Ollama.
- `PORTOS_VOICE_DEFAULT_TOOL_MODEL` still overrides the whole chain with one id.
- The LM Studio "skip the load if already resident" guard is intact (a second `lms load` spawns another instance and doubles VRAM).
- `ensureToolCapableModel` gained the `cfg.enabled` guard its sibling `preloadModel` already had, so voice-off can never reach a download.

## Testing

- `server`: 2266 files / 45,390 tests pass.
- `client`: 984 files / 12,100 tests pass.
- New: `modelProvisioners.test.js` (14) covering the sentinel in both directions, capability precedence and its fallback, chain derivation/ordering/override, and the Ollama warm path; 5 bootstrap orchestration tests covering the unreachable short-circuit, mid-chain abort, remote-provider skip and already-resident skip; 6 migration tests covering pin/no-pin/idempotence.


atomantic:
Rebased onto `main` after #7542 (the CoS fix for #7541) landed, which touched the same region of `bootstrap.js`.

Both PRs fix the same bug; this one supersedes that implementation with the provider-agnostic version, so the LM-Studio-specific code path it added is replaced rather than kept alongside. Two things from #7542 were carried forward deliberately:

- **The malformed-200 hardening.** A 200 whose body is not the OpenAI-compatible `{data: [...]}` shape (truncated response, a gateway returning HTML) now also returns the `null` sentinel rather than `[]`. Adopted into the LM Studio provisioner.
- **Its whole test file.** `bootstrap.lmStudioUnreachable.test.js` is kept at the same path with every case intact, re-pointed at the real LM Studio provisioner through `bootstrap` instead of the removed `listLmStudioModels` export. The unreachable verdict is now `{ skipped: 'backend-unreachable', backend }` — the same state, named for the backend that reported it. One case was added for the malformed-200 path.

The one thing not carried over is `isLmStudioConnectError`, a regex against LM Studio's own wording. The orchestrator now detects the same condition through the post-install re-list returning `null`, which costs the same single attempt and works for any backend rather than only the one whose error string it matched.

Server suite green at 45,398 tests.

PR #7638, opened 2026-09-18
fix: carry forward the place trail on a failed Eidoverse world read

### LINKED ISSUE #7633: A failed world read erases the Eidoverse observation place trail and fabricates every district as changed, twice
## Problem

`buildEidoverseObservation()` has four marker sections. Three of them carry the previous marker forward when their source could not be read; the fourth — the place trail — does not, so one transient read failure erases the trail and fabricates a full set of district changes **twice**: once on the failure wake, and again on recovery.

### The guard that exists

`e623bc607` ("stop a failed collection from flapping the Eidoverse visit marker") added `section()` (`server/lib/eidoverseObservation.js:292-294`):

```js
function section(ids, available, carriedForward) {
  return available ? { ids, available: true } : { ids: asArray(carriedForward), available: false };
}
```

and the module header states the rule it enforces (`:20-24`): *"Absent is never empty, here or in the trail … an unavailable section contributes no changes and CARRIES FORWARD what the previous marker held."*

### The section it never reached

`:346`:

```js
const placeStatus = Object.fromEntries(places.map(({ id, status }) => [id, status]));
```

No guard. That raw map is both diffed (`:269-272`) and committed to the marker (`:377`). A district whose source failed is `status: 'unknown'` (`:121-123`) — a *legitimate value* to the differ — so it flows straight through as a change.

### How it triggers

`server/services/eidoverseObservationLedger.js:110-112` degrades a failed `collectEidoverseWorldSources()` to `{}`. The three protected sections correctly become `null`; every district then reads `unknown`. `collectEidoverseWorldSources()` is roughly twenty parallel service reads plus an outbound `/capabilities` probe per online peer — a realistic thing to have fail once.

Driving the shipped pure builder through healthy → total outage → identical healthy:

```
wake 2 placesChanged: 8 entries, every district "active" → "unknown"
wake 2 marker:        all 8 districts persisted as "unknown"   (trail erased)
wake 2 foundationIds: []  (correctly carried forward)
wake 3 placesChanged: 8 entries, every district "unknown" → "active"
```

## Impact

`changes.placesChanged` is capped at `MAX_CHANGE_ENTRIES = 20` and is the section the `maintain` phase template explicitly tells the Mind to act on (`server/lib/persistentMindPlaybook.js:73`). One flaky read spends two consecutive wakes handing the Mind eight fabricated district changes — precisely the flapping `e623bc607` exists to prevent, in the same function, on the one section it did not cover. This is the third instance of this defect class in this slice's history (`be8b01334`, `181476f4b`, `e623bc607` are all the same bug in different sections), which is the reason it is worth a ticket rather than a drive-by.

**Apply `section()`'s carry-forward semantics per district key.**

In `buildEidoverseObservation()` (`server/lib/eidoverseObservation.js:322, 346`):

1. Track per-district availability — a district is unavailable when its `signalCount === null` and it contributed to `unreadableSources`. That state is already computed at `:121-123`; surface it rather than collapsing it into the `unknown` status string.
2. Exclude unavailable districts from `placesChanged`.
3. For those districts, merge the **previous marker's** status into the committed `placeStatus` instead of overwriting it with `unknown`.

Keep `status: 'unknown'` in the *report* — telling the Mind a district could not be read is correct and is what `:121-123` is for. The fix is only that an unreadable district must not write over the trail or count as a change.

## Acceptance criteria

- [ ] A healthy → outage → identical-healthy sequence produces zero `placesChanged` entries on both the outage wake and the recovery wake.
- [ ] The marker retains the pre-outage status for a district that could not be read.
- [ ] A district that genuinely changed status while others were unreadable still reports as changed.
- [ ] The report still shows `status: 'unknown'` for an unreadable district.
- [ ] The test drives the shipped builder end to end, not a hand-built `places` array.

## Out of scope

- The phase picker's use of these signals (#7630).

## Why this was missed

`e623bc607` fixed the three sections that are *lists of ids* — they share the `section()` helper, so fixing one shape fixed all three. `placeStatus` is a map, not a list, so it did not pass through the helper and did not look like the same bug.

Refs #7453, #7457.

atomantic:
## Summary

`buildEidoverseObservation()`'s three id-list marker sections (`foundationIds`, `peerIds`, `attentionControllerIds`) go through `section()`, which carries the previous marker forward when a source fails to collect. The fourth marker section, `placeStatus`, was built without that guard: a failed world read degraded every district's signal to `null`, which collapsed into status `'unknown'` and flowed straight into the committed marker and the diff — fabricating every district as a changed place on the failure wake, and again on recovery when the real status flipped back from the erased `'unknown'`.

Fix: apply `section()`'s carry-forward semantics per district key. A district whose signal genuinely failed to read (`signalCount === null` with a non-empty `unreadableSources`) keeps the PREVIOUS marker's status in the committed marker instead of being overwritten with `'unknown'`. The observation report still shows `status: 'unknown'` for that district so the mind is told the read failed — only the persisted trail and the change diff are protected.

## Test plan

- Added a regression test in `server/lib/eidoverseObservation.test.js` that drives the shipped `buildEidoverseObservation()` builder through healthy -> total outage -> identical healthy, asserting:
  - the outage wake reports zero fabricated `placesChanged` entries and the marker retains the pre-outage status
  - the recovery wake also reports zero fabricated changes for the district that never moved
  - a district that genuinely changes status while others are unreadable is still reported as changed
  - the report keeps `status: 'unknown'` for the unreadable district
- `server/node_modules/.bin/vitest run server/lib/eidoverseObservation.test.js server/services/eidoverseObservationLedger.test.js` — 29 passed

Closes #7633


atomantic:

Issue #7616, opened 2026-09-18
ship a Reality universe and give the universe record a factual flag


atomantic:
## Problem / Goal

There is no factual universe in PortOS, and none ships:

- The universe record (`server/services/universeBuilder/sanitize.js:683-790`) has no factual/fiction axis. Canon trunks are `characters/places/objects`; category kinds are the same three plus `other`.
- `createUniverse` mints `randomUUID()` and accepts no caller id (`server/services/universeBuilder/crud.js:244-247`).
- `data.reference/` contains no `universes/` directory; `bootstrapServices` reads `data.reference` only for `providers.json` (`server/services/bootstrap.js:264`). Universes are PG rows (`server/lib/db/schema/universes.js:19`), so `setup-data.js` is not the seeding mechanism — a migration is.
- Repo-wide there is no `factual` / `non-fiction` / `isFiction` concept on a scrap, an ingredient, or a universe.

Goal: factual/autobiographical capture has a home from first boot, and the universe carries the flag that tells the extractor (landed in #7609) how to read the text.

## Approach

1. Add `factual: boolean` to the universe record in `sanitizeTemplate` (`server/services/universeBuilder/sanitize.js:683-790`), persisted **only when `true`** so every existing record keeps its on-disk shape and wire checksum.
2. **Bump `universes` 11 → 12** in `server/lib/schemaVersions.js:82`, with the rationale inline: a ≤v11 peer strips `factual`, then LWW's the stripped record back and silently reclassifies the user's Reality universe as fiction. Update the three assertions in `server/lib/schemaVersions.test.js` (`:27`, `:64`, `:71`).
3. Seed migration `scripts/migrations/NNN-reality-universe-seed.js`: a universe with the **deterministic id `universe-reality`** (matches `UNIVERSE_ID_RE = /^[A-Za-z0-9-]{8,80}$/`, `sanitize.js:51`), `name: 'Reality'`, `factual: true`, `ephemeral: false`, and **empty** logline / premise / styleNotes. Shipping invented copy for the user's own life is wrong, and an LLM call at migration time is barred by the AI Provider Usage Policy in `AGENTS.md`.
   - Deterministic id so a user's federated machines converge on ONE Reality by LWW instead of accumulating one per install.
[...]

PR #7644, opened 2026-09-18
bind catalog scrap-commit to a universe so committed ingredients stop landing unlinked

### LINKED ISSUE #7615: bind catalog ingest to a universe so committed ingredients stop landing unlinked
## Problem / Goal

`commitScrap` writes ingredients + a source link and **nothing else** (`server/services/catalogDB/commit.js:16-34`) — no `catalog_ingredient_refs` row, no `catalog_ingredient_relations` edge. So `HAS_ANY_HOMING_REF` (`server/services/catalogDB/shared.js:258`, `HOMING_REF_KINDS = ['universe','series','creative-director']`) is false for every freshly ingested row and they all bucket as **unlinked (Raw)**. That is the "isolated and random" symptom in #7609.

Goal: one user-visible control at ingest — the universe this material belongs to — that links every committed ingredient to that universe and connects a single scrap's extractions to each other.

The infrastructure already exists and already federates. `/api/catalog/bulk-import` is the prior art in the same file: it already takes `defaults.universeRef / seriesRef / workRef / issueRef / role` (`server/lib/catalogValidation.js:449-462`, applied at `server/routes/catalog.js:569`). The scrap→commit path is the odd one out.

## Approach

1. Add `universeRef` (+ optional `role`) to `catalogScrapCommitSchema` (`server/lib/catalogValidation.js:407`), mirroring `catalogBulkImportSchema.defaults` (`:449-462`) — same field name, same `z.string().trim().min(1).max(120).optional()` shape, so the two ingest paths cannot drift. Forward both from the commit route (`server/routes/catalog.js:163-182`) into `commitScrap`.
2. **`linkIngredientToRef` and `linkIngredientRelation` currently run pool-level `query(...)` with no `{ client }` option** (`server/services/catalogDB/refs.js:46`, `:180`), so they cannot join `commitScrap`'s `withTransaction`. Add the `{ client } = {}` option to both, exactly as `linkIngredientToSource` already has it (`refs.js:20`). Without this, a mid-batch failure leaves ingredients committed with refs half-written.
3. Add `universeRefRoleForType(type)` to `refs.js` beside the existing `seriesRefRoleForType` (`:69`) and `cdRefRoleForType` (`:95`): `character` → `canon-character`, `place` → `canon-place`, `object` → `canon-object`, everything else (including `idea` / `scene` / `concept` and user-defined types) → `reference`. The `canon-<kind>` convention is what `catalogCanonProjection.js` and `catalogUniverseTags.js` already expect.
4. Inside `commitScrap`'s existing transaction, after each `createIngredient`, call `linkIngredientToRef(ingredient.id, 'universe', universeRef, role || universeRefRoleForType(draft.type), { client })`.
5. Mint `related-to` edges between the ingredients committed from one scrap in one batch, so a single source's extractions form a connected cluster instead of N isolated nodes. **Deterministic direction:** one edge per unordered pair, `from_id` = the lexicographically smaller id — the PK is `(from_id, to_id, kind)`, so an arbitrary direction would let a re-commit create a reciprocal duplicate. Bounded: skip edge minting entirely when a batch exceeds 25 accepted rows (the schema allows 200, which would be 19,900 edges).
6. **Client:** a "Catalogue into" `<select>` on the Ingest form covering all source modes (`client/src/pages/CatalogIngest.jsx`), listing live universes plus an explicit **Unassigned** option that preserves today's behavior exactly. Default: **Reality** for `brain-bridge` and `voice-memo` (captured thought is factual by default), otherwise the last-used universe read through `safeReadStorage` (`client/src/lib/safeStorage.js`) falling back to Reality. When the selected universe is `factual`, show a one-line hint that real people and places will be extracted as themselves. Follow `client/src/AGENTS.md` for `htmlFor`/`id` pairing and mobile layout.
7. Once the universe carries a `factual` flag (#7616), feed it into the extraction lens landed in #PHASE1 — `catalogExtraction.js` derives `factual` from the scrap's `sourceKind` today and the universe flag becomes the primary signal, with the source kind as fallback.

## Acceptance criteria

- [ ] Committing a scrap with `universeRef` set creates one live `catalog_ingredient_refs` row per ingredient with the role `universeRefRoleForType` returns; `HAS_ANY_HOMING_REF` is true for all of them and none bucket as unlinked/Raw.
- [ ] `server/routes/catalogParsedBodies.test.js` (which already covers `POST /scraps/:id/commit` with a mocked `commitScrap`) asserts `universeRef` / `role` are forwarded, and a DB-backed test in `server/routes/catalog.test.js` asserts the refs and relation edges actually land — plus that a mid-batch failure rolls back ingredients, refs, and edges together.
- [ ] Omitting `universeRef` reproduces today's behavior exactly (source link only, no refs, no relations) — a regression test pins this.
- [ ] Same-batch `related-to` edges are one per unordered pair with `from_id` lexicographically smaller; re-committing the same batch creates no duplicate or reciprocal edge; a >25-row batch mints none.
- [ ] Ingest UI: the "Catalogue into" select is labelled (`htmlFor`/`id`), keyboard-reachable, defaults as specified, persists the last choice through `safeReadStorage`/`safeWriteStorage`, and renders correctly at phone width.
- [ ] `docs/STORAGE.md` and the catalog docs describe the universe binding.

## Out of scope

- Resolving `Mom` / `Dad` to brain `people` records, and resolving `scene.actors[]` strings to ingredient ids — both are entity resolution, filed separately.
- Reading catalog tables from `universeGraph.js` / `brainGraph.js`, and including relations in `exportSliceForRef` (the `[catalog-ingredient-relations]` TODO at `server/services/catalogDB/facets.js:60`).

atomantic:
## Summary

`commitScrap` wrote committed ingredients + a source link and nothing else, so every freshly extracted-and-committed ingredient bucketed as unlinked ("Raw") — `HAS_ANY_HOMING_REF` was false because no `catalog_ingredient_refs` row was ever created for the scrap→commit ingest path (bulk-import already had this via `defaults.universeRef`).

- Add optional `universeRef` (+ `role`) to `catalogScrapCommitSchema`, mirroring `catalogBulkImportSchema.defaults` so the two ingest paths share one shape. The commit route forwards both into `commitScrap`.
- `linkIngredientToRef` and `linkIngredientRelation` now accept the same `{ client }` option `linkIngredientToSource` already had, so `commitScrap` can link the universe ref and mint relation edges inside its existing transaction — a mid-batch failure rolls back ingredients, refs, and edges together.
- Add `universeRefRoleForType(type)` next to the existing `seriesRefRoleForType`/`cdRefRoleForType`: character/place/object map to `canon-<kind>`, everything else links as `reference`.
- Inside `commitScrap`, every created ingredient links to `universeRef` (when supplied) with the derived (or overridden) role, in the same transaction.
- Mint one `related-to` edge per unordered pair among the ingredients committed from a single scrap/batch, so one source's extractions land as a connected cluster instead of isolated nodes. Direction is deterministic (`from_id` = the lexicographically smaller id, since the PK is `(from_id, to_id, kind)`), and batches above 25 accepted rows mint no edges (the schema allows up to 200, which would be 19,900 pairs).
- Client: a "Catalogue into" `<select>` on the Ingest review phase. Defaults to a universe named "Reality" for brain-bridge/voice-memo capture-source ingests when one exists, otherwise the last-used pick (persisted via `safeReadStorage`/`safeWriteStorage`), falling back to "Reality" then "Unassigned". Choosing "Unassigned" reproduces today's behavior exactly (source link only, no homing ref).
- `docs/STORAGE.md` now describes the universe-binding write path on `catalog_ingredient_refs` and the same-batch relation edges on `catalog_ingredient_relations`.

Deferred to #7616 (named in the issue as a follow-up): wiring a universe's `factual` flag into the extraction lens, since universes don't carry that flag yet.

Closes #7615

## Test plan

- `server/services/catalogDB.commit.test.js` — mocked-transaction unit coverage: universe ref linking with the type-derived role, explicit role override, no ref when `universeRef` is omitted, deterministic `related-to` edge direction/count for a multi-item batch, no edges for a single-item batch, no edges over the 25-row bound.
- `server/routes/catalogParsedBodies.test.js` — the commit route forwards `universeRef`/`role` (and forwards `undefined` when omitted) into `commitScrap`.
- `server/routes/catalog.test.js` (DB-backed, `npm run test:db`) — committing with `universeRef` creates the expected refs, `HAS_ANY_HOMING_REF`-driven facets bucket count includes them, and a same-batch `related-to` edge lands with the correct direction; omitting `universeRef` reproduces prior behavior (source link only); a 26-row batch mints zero relation edges; a mid-batch DB failure (`NOT NULL` violation) rolls back the earlier ingredient in the same transaction.
- `client/src/pages/CatalogIngest.test.jsx` — existing babble-mode commit flow still passes with the new universe-select state wired in.
- Ran `server/node_modules/.bin/vitest run` against the full non-DB catalog suite (430 tests) and the DB-backed catalog suite via `PGDATABASE=portos_test` (50 tests) — all green.
- Ran `client/node_modules/.bin/vitest run src/pages/CatalogIngest.test.jsx` and `biome lint` on the touched client files — clean.

Required local code review could not complete before publication: `provider:opencode-zen-cli` reports no enforced tool-free review transport for this repo, and `provider:nvidia-nim` timed out/aborted on both attempts. See the follow-up PR comment.


atomantic:

PR #7645, opened 2026-09-18
ship a factual Reality universe and give the universe record a federated factual flag

### LINKED ISSUE #7616: ship a Reality universe and give the universe record a factual flag

atomantic:
## Summary

Factual/autobiographical capture had no home in PortOS: every universe was fiction, so a voice memo or a journal entry got extracted through a story lens. This gives the universe record a factual/fiction axis and ships one world that has it set.

- **`factual` on the universe record.** `sanitizeTemplate` persists it **only when `true`** (like `ephemeral`/`importDraft`), so every existing universe keeps its exact on-disk shape — and therefore its wire checksum, which peer sync and the conflict journal both hash. A test pins that a fiction record is byte-identical before and after the change.
- **`PORTOS_SCHEMA_VERSIONS.universes` 11 → 12.** Unlike `ephemeral`/`importDraft`, `factual` rides the wire: a world's factual nature belongs to the world, not to one machine. A ≤v11 peer would sanitize the flag away and last-writer-wins the loss back, silently reclassifying the user's Reality universe as fiction — so the category gate bumps.
- **Migration `395-reality-universe-seed.js`** seeds `universe-reality` ("Reality", `factual: true`).

**Why a migration and not a `data.reference/` seed** (the pattern choice AGENTS.md asks to justify): universes are Postgres rows (`server/lib/db/schema/universes.js`) and `scripts/setup-data.js` only copies *files* into `data/`, so there is no file for a seed to land in. This is also **not** the "derived from the install's own records" case `scripts/lib/migrationOwnedPaths.js` governs — that list is about `data/` file paths, and this migration writes no file at all, so it gets no entry there.

**Why a deterministic id.** A user commonly federates several machines, each running this migration itself. A `randomUUID()` would give them one Reality per install, all syncing to each other; one fixed id makes the copies converge on a single record under the normal LWW merge.

**Why the tombstone check lives in the migration.** The gate is "does ANY row with this id exist", `deleted` true or false — not "is there a live row". `insertUniverseWithId` deliberately *overwrites* a tombstone (its `wasResurrection` branch), which is right for a share-bucket re-import and wrong here: a Reality the user deleted must not be resurrected by the next upgrade. The check belongs in the migration rather than as a `resurrect: false` option on the shared importer path.

**Why the copy is empty.** `logline` / `premise` / `styleNotes` ship blank. Inventing copy for someone's own life would be wrong, and generating it is barred outright — AGENTS.md "No cold-bootstrap LLM calls" rules out an LLM call on a boot path.

UI: a "This is a real-world universe" toggle in the Universe Builder bible settings (part of the normal draft Save), and a `Real` badge in both layouts of the universe list, fed by a new `factual` column on the summary projection.

### Deliberately not folded in

The issue also asked to demote the catalog's `factual` / `real-person` **tags** to a display affordance derived from the real field. That is not yet possible: `catalog_ingredients` has no universe link, so a row cannot read a universe's `factual` flag today. Binding a scrap to a universe is **#7615**, which is open and is the prerequisite — the demotion belongs there, and I have noted it on that issue rather than filing a duplicate. `server/lib/catalogSourceKinds.js` and `server/lib/README.md` are updated to point at the now-existing flag and at #7615 as the step that makes it the primary signal.

## Test plan

- `cd server && npx vitest run` — 2300 files / 46,061 tests pass.
- `cd client && npx vitest run` — 993 files / 12,214 tests pass.
- New: `scripts/migrations/395-reality-universe-seed.test.js` — seeds one live factual Reality with the deterministic id and no invented copy; **does not resurrect a tombstoned Reality**; no-ops when the row is already present (re-run, or synced in from a peer).
- New in `server/services/universeBuilder.test.js` — `factual` is absent from the persisted record when false and the fiction record is byte-identical; the flag round-trips through create and PATCH and `false` clears it.
- New in `client/src/hooks/useUniverseDraft.test.jsx` — the toggle marks the draft dirty and ships `factual` with Save (the regression: omitting it from the draft snapshot leaves Save a no-op).
- New in `client/src/pages/Universes.test.jsx` — the `Real` badge renders only for a universe whose `factual` flag is set.
- Updated version assertions: `server/lib/schemaVersions.test.js`, `server/services/dataSync.pipelineUniverse.test.js`, `server/services/sharing/peerSync.test.js`, `server/services/sharing/integration.test.js`.
- `server/lib/db.ddlParity.test.js` green (no DDL change).

Closes #7616


atomantic:

atomantic:
Review loop: both configured local reviewers (`provider:opencode-zen-cli`, `provider:nvidia-nim`) returned `HTTP 401 Authentication required` from `/api/code-review/local` — this instance has the optional password enabled and the agent has no credential, so neither pass could run. Both are configured optional, so that is non-blocking; the diff was validated independently instead.

That pass found one material issue, fixed in 1ae9a7b6: `factual` federates and so feeds the wire projection `contentHashForRecord` hashes, but `RESTORABLE_FIELDS.universe` omitted it. Since `diffSummary` filters to that list, a peer disagreement on the factual/fiction axis journaled a conflict whose diff could not name the field — unreadable and unresolvable in the Conflicts UI, with the flag lost to LWW silently. Restoring it also needed the clear direction: `sanitizeTemplate` persists `factual` only when true, so a fiction snapshot has no key and the resolver`s `pick` skipped it, making "restore to fiction" a silent no-op. `ABSENT_MEANS_FALSE_FIELDS` names that shape so `pick` fills the explicit `false`.

Both new tests were verified to fail against the un-fixed source. The earlier red CI was not this change: the first run was cancelled wholesale by an external sweep, and the second failed only on `scripts/generate_fastvideo.test.js` and `scripts/minimax_h3_runtime_probe.test.js` — Python-subprocess tests timing out at 30s on a cold Windows runner, untouched by this PR. They pass on re-run and the full gate is green.

PR #7646, opened 2026-09-18
Read an Eidoverse controller's config back and revalidate it on every tick

### LINKED ISSUE #7629: An Eidoverse controller's config is never read back or revalidated: the panel shows {} and a definition change keeps stepping the old record
## Problem

#7456's acceptance criteria listed *"Mind can install/inspect/retire a controller via tools."* Install, arm and retire shipped. **Inspect did not** — and the same omission produces a plain user-visible bug in the panel.

### Nothing reads a controller's config back

`summarizeControllerInstall()` (`server/lib/eidoverseControllers.js:340`) puts `config` and `state` behind an `includeState` flag, and its own header (`:333-339`) states the intended contract:

> *"The full record carries the controller's whole durable state … so a LIST omits it and an INSPECT of one install includes it."*

Every `includeState: true` call site in the tree — `server/routes/eidoverseWorldRoutes.js:194` and `server/services/cosToolRegistry.js:801` — is the response to a **just-submitted install**. There is no inspect. The `eidoverse-controllers` tool group has exactly four operations: `controllers`, `arm-controller`, `install-controller`, `retire-controller` (`cosToolRegistry.js:775-806`). `getEidoverseControllerInstall()` (`eidoverseControllerRuntime.js:121`) has no caller outside its own test file.

**Consequence in the shipped UI:** `client/src/components/eidoverse/EidoverseControllersPanel.jsx:292-293` renders

```jsx
<p className="text-gray-400">Config:</p>
<pre …>{JSON.stringify(install.config ?? {}, null, 2)}</pre>
```

from list rows supplied by `GET /controllers` → `summarizeControllerInstall(install)` with no `includeState` (`eidoverseWorldRoutes.js:175-183`). `install.config` is always `undefined`, so **every controller's Config block renders `{}`**, including immediately after install (`submitDraft` ends in `refresh()`). The panel's tests use mocked fixtures and never assert on that block, so it is invisible in CI.

### A stored install is never revalidated against the definition it runs

The live tick path hands the frozen record straight to `step()` (`eidoverseControllerRuntime.js:320`):

```js
const outcome = runControllerStep({ definition, state: record.state, config: record.config, tick });
```

`runControllerStep()` (`server/lib/eidoverseControllers.js:245-287`) references neither `definition.configSchema` nor `definition.invariants`; it validates only the step's *output* (serializability, size, effect vocabulary). `config` was parsed once, at install time, by whatever version was running then, and `createState` is never re-run.

`invariants` — the predicates the promote gate treats as proof of resilience — appear exactly once outside the definitions: `server/services/eidoverseControllerRegistry.js:148`, inside `controllerAssayContributions()`. The assay also re-parses config deliberately (`:138`, *"so the replay runs against the config an INSTALL would get (defaults applied)"*). **The assay is strictly stricter than the real path it claims to replay.**

Version signalling is inert alongside it: `writeInstalls` stamps `STORE_SCHEMA_VERSION` (`eidoverseControllerRuntime.js:106`), `readInstalls` (`:100-103`) never reads it, and `listEidoverseControllers` re-echoes the *constant* (`:118`), so a store written by a newer install reports as v1 to every consumer. `eidoverseControllerRecordSchema` (`eidoverseControllers.js:170`) is exported, catalogued in `server/lib/README.md`, and used by nothing.

## Impact

**Today:** the Config block in the panel is always empty, and a Mind's only way to change an inherited controller is to re-install the same id — which replaces `config` wholesale and rebuilds state from it (`eidoverseControllerRuntime.js:186`, `state: definition.createState(config.data)`). A later Mind cannot read the config it is about to overwrite, so the epic's *"later minds inherit and modify them"* is guess-and-clobber, and the accumulated state of the inherited thing is destroyed in the process. The human's only recourse is opening `data/eidoverse/controllers.json` by hand.

**On the next shipped-controller change:** per root `AGENTS.md`, installs upgrade on their own schedule. The first time a definition's config or state shape changes — a new field with a `.default()`, a renamed key, an added counter — live installs keep stepping the old record. A new default arrives as `undefined` and flows into world verb args (`lanternKeeper.controller.js:80-83` maps config values directly into `light` args); a new state field arrives as `undefined` into arithmetic. Neither throws, so `maxConsecutiveFailures` never fires — and per #7628 the resulting world refusals are invisible too. Restored backups and downgrades take the same path with no version gate to notice.

**1. Ship the missing inspect.** Add `GET /api/eidoverse/world/controllers/:id` returning `summarizeControllerInstall(record, { includeState: true })`, and an `inspect-controller` operation in the `eidoverse-controllers` group (`cosToolRegistry.js:775`) backed by the already-exported `getEidoverseControllerInstall()`. Have `EidoverseControllersPanel`'s expand handler fetch it instead of reading `install.config` off a list row that never carries it. Name the new operation in the `installEidoverseControllers` guidance in `server/services/persistentMindOrientation.js`.

**2. Re-parse config on every tick.** In `stepInstall()` (`eidoverseControllerRuntime.js:304`), run `record.config` through `definition.configSchema` before the step. A parse failure disarms with a reason, in the same shape as the existing missing-definition branch (`:307-317`). Persist the re-defaulted config so a newly-added default lands exactly once rather than on every tick.

**3. Run the definition's `invariants` on live state.** Inside `runControllerStep()` (`eidoverseControllers.js:245`), evaluate `definition.invariants` against `outcome.state` and treat a violation as a step failure. The live supervisor then holds a controller to the same contract the promote gate does, instead of the gate being the stricter of the two.

**4. Make the store's version stamp mean something.** Read the file's `schemaVersion` in `readInstalls`, return *that* from `listEidoverseControllers`, and refuse to step a store stamped newer than this build (disarm with a reason rather than throwing — a downgraded install should stop, not crash).

**5. Add a config-only update path** (`updateEidoverseControllerConfig(id, config)`) that re-parses and preserves `state`, so "modify an inherited controller" stops meaning "destroy its accumulated state". This is the verb the epic's *inherit and modify* actually names.

## Acceptance criteria

- [ ] `GET /controllers/:id` and an `inspect-controller` tool return `config` and `state`; a test pins that the LIST still omits them.
- [ ] The panel's Config block shows a real config; a client test asserts a non-empty config renders (today's fixture would pass against `{}`).
- [ ] A stored config that no longer parses against its definition disarms the install with a readable reason instead of stepping.
- [ ] A newly-added `configSchema` default is applied and persisted once for an existing install; a test covers the upgrade.
- [ ] An `invariants` violation on live state fails the step and counts toward `maxConsecutiveFailures`.
- [ ] A store stamped with a newer `schemaVersion` disarms rather than stepping, and `listEidoverseControllers` reports the file's stamp, not the constant.
- [ ] `updateEidoverseControllerConfig` changes config while preserving `state`; a test asserts the accumulated state survives.

## Out of scope

- Delivery-outcome health (#7628).
- Adopting a peer's controller foundation (#7626).

## Why this was missed

`includeState` was built for exactly this and wired only to the two install responses, where it looks correct in review — the flag exists, it is documented, and two call sites pass it. The acceptance criterion said "inspect" and the reviewer saw a projection that supports inspection. Nothing asked which caller performs one.

Refs #7453, #7456.

atomantic:
## Summary

- Add the missing INSPECT for an Eidoverse controller install: `GET /api/eidoverse/world/controllers/:id`, a client `getEidoverseControllerInstall` wrapper, and an `eidoverse.inspect-controller` mind tool — all backed by the runtime's `getEidoverseControllerInstall()`, which previously had no caller outside its own test. The panel's Config block fetches this on expand instead of reading `install.config` off a list row that `summarizeControllerInstall()` never puts it on, so it stops always rendering `{}`.
- Revalidate a stored install against the definition it runs on every tick, instead of trusting install-time validation forever:
  - `record.config` is re-parsed through `definition.configSchema` before each step. A config that no longer parses disarms the install with a readable reason instead of stepping; a newly-added `configSchema` default is applied and persisted the first time it lands.
  - `definition.invariants` — the same predicates the resilience assay replays — run against the state a step just produced. A violation fails the step and counts toward `maxConsecutiveFailures`, so the live supervisor holds a controller to the same contract the promote gate does.
  - The store's own `schemaVersion` stamp is read back instead of assumed to be the build's constant; a store stamped newer than this build disarms rather than stepping, and `listEidoverseControllers` reports the file's actual stamp.
- Add `updateEidoverseControllerConfig()` (routed at `PATCH /api/eidoverse/world/controllers/:id/config`), so changing an inherited controller's config no longer has to go through a re-install that rebuilds `state` from scratch. The panel gets a matching "Edit config" form in the Details view.

## Test plan

- `cd server && node_modules/.bin/vitest run` — full suite green (2299 files, 46068 tests, 36 pre-existing skips).
- `cd client && node_modules/.bin/vitest run` — full suite green (993 files, 12216 tests, 7 pre-existing skips).
- New/updated coverage: `server/lib/eidoverseControllers.test.js` (invariant enforcement), `server/services/eidoverseControllerRuntime.test.js` (config reparse/disarm, default upgrade, invariant failure counting toward `maxConsecutiveFailures`, newer-`schemaVersion` disarm, `updateEidoverseControllerConfig`), `server/routes/eidoverseWorldRoutes.test.js` (new GET/PATCH routes), `server/services/cosToolRegistry.test.js` (catalog includes `eidoverse.inspect-controller`), `client/src/components/eidoverse/EidoverseControllersPanel.test.jsx` (fetches and renders a real config on Details; edits config through the new editor without an install/re-install).

Closes #7629


atomantic:

Issue #7458, opened 2026-09-16
Mind playbooks: explore→construct→maintain→coordinate as Commons matures


atomanticagent:
## Parent

Part of #7453.

## Problem

SwarmWorld agents spontaneously shift among **exploration, construction, maintenance, and coordination** as the world matures. PortOS continuous-play is currently a single loop; empty Commons and dense Commons need different wakes.

## Proposal

Maturity-aware playbook phases for the persistent mind (and optionally CoS):

1. **Explore** — sparse world, map affordances, reconnect presence
2. **Construct** — densify districts, leave proposals/controllers
3. **Maintain** — repair broken projections, retire dead controllers, re-test foundations
4. **Coordinate** — visit peers, package promotes, steward shared baseline

Phase selection from world signals (density, failure rates, unread peer contributions), not fixed cron personality.

## Acceptance

- Signals + phase picker documented and wired into continuous-play.
- Wakes in a dense Commons spend more time maintain/coordinate than empty-explore filler.
- Helm/other minds can report current phase in working notes.

Issue #7457, opened 2026-09-16
Eidoverse: observation-first discovery of peer foundations (stigmergy over chat)


atomanticagent:
## Parent

Part of #7453.

## Problem

SwarmWorld finds most technology reuse begins through **physical observation**, not communication. PortOS federation today leans chat/visit-chat; minds may miss peer foundations unless told.

## Proposal

Make discovery **observation-first**:

- Richer inspect/status affordances for places, projections, controllers, provenance
- “What’s new since last visit” stigmergic trails (markers, districts, federation showcase)
- Continuous-play default: prefer `eidoverse.status` / travel / inspect before asking peers in chat

## Acceptance

- Mind tools expose inspectable artifact metadata (author instance, promote status, controller health).
- Playbook/docs bias observation → interaction → chat.
- A peer contribution is discoverable by touring Commons without reading GitHub.

PR #7647, opened 2026-09-18
fix: make all four Mind playbook phases reachable and gate Coordinate on measured peer contributions

### LINKED ISSUE #7630: Two of the four Mind playbook phases are unreachable, and Coordinate claims peers are waiting on a signal that only measured reachability
## Problem

#7458 built the maturity ladder the epic asked for: *"Explore → construct → maintain → coordinate as Commons matures; playbooks should shift with world maturity, not stay one-shot chatter."* Four phases shipped, with a picker driven by live world signals.

**Two of the four are unreachable on every real install, and the phase that does fire tells the Mind something nobody measured.**

### `districtCount` has a floor of ~19; the maturity threshold is 16

`server/lib/persistentMindPlaybookPhase.js:30-34`:

```js
const THRESHOLDS = Object.freeze({ sparseDistrictCount: 4, matureDistrictCount: 16, highFailureRate: 0.25 });
```

`derivePersistentMindPlaybookPhaseSignals()` (`:92-128`) sums eleven signal arrays — and several of them are **install-constant scaffolding, not anything a Mind builds**:

- `features` — `server/lib/eidoverseWorldSignals.js:392` maps **every** registry feature, enabled or not: `featuresState.features.map(...)`. `INSTANCE_FEATURES` has ~15 entries. Fixed **+15**.
- `storage` — `eidoverseWorldSignals.js:222-249` unconditionally pushes `database` and `filesystem`. Fixed **+2**.
- `operations` — `:252-256` returns a single `overview` row. Fixed **+1**.
- `productivity` / `activity` — one summary row each. Fixed **+2**.

Run against a maximally-empty install (no apps, agents, tasks, goals, memories, peers, activity, Jira):

```
per-source counts: {"apps":0,"agents":0,"tasks":0,"features":15,"peers":0,"productivity":1,
                    "activity":1,"goals":0,"memory":0,"storage":2,"jira":0,"operations":1}
derived: {"districtCount":19,...}
phase:   {"phase":"maintain","reason":"mature Commons (19 live signal(s)), steady-state upkeep"}
```

A brand-new install with nothing built is told its Commons is mature and instructed to *"repair or retire what no longer serves the world before adding anything new."* Even if `getInstanceFeatures()` fails entirely (`features → null`, uncounted), the floor is still 5 — above `sparseDistrictCount: 4`, so `explore` is unreachable regardless. `construct` requires falling between the two thresholds and cannot be reached either.

Call path: `persistentMindAdapter.js:356` → `resolvePersistentMindPlaybookPhase()` → `persistentMindPlaybookSignals.js:25-33` → `collectEidoverseWorldSources()` (returns `buildEidoverseWorldSignals()` output verbatim, `eidoverseWorldSources.js:116`) → `derivePersistentMindPlaybookPhaseSignals` → `selectPersistentMindPlaybookPhase`.

**This is the exact defect class 7fd55add1 already named once in this slice** — "made a phase decision on a signal that could not vary." It was fixed for the peer signal and left in place for the density signal that the whole ladder rests on.

### `coordinate` claims peers are waiting; only reachability was measured

`7fd55add1` corrected the fabricated peer claim in the *reason string* and left the identical claim in the *template the Mind actually reads*. The signal is explicitly reachability, and the code says so (`persistentMindPlaybookPhase.js:123-125`): *"This is NOT 'peers with unread contributions'."* The template it selects (`server/lib/persistentMindPlaybook.js:79-81`):

```
PLAYBOOK PHASE — Coordinate (peers waiting)

Federated peers have activity worth your attention.
```

That reaches the model verbatim: `persistentMindPlaybook.js:144-146` → `composePersistentMindInstructions` → `persistentMindAdapter.js:365` → `persistentMindSupervisor.js:887`. Because `districtCount` is always ≥ 16, `coordinate` is the **default** phase for any install with one always-online peer — so the false claim is the steady state, not an edge case.

#7458's acceptance named the missing input by name: *"Phase selection from world signals (density, failure rates, **unread peer contributions**)."* #7457 already persists exactly that — `changes.newFoundations` / `changes.newPeers` from the durable marker in `data/eidoverse/observation.json` (`server/lib/eidoverseObservation.js:260-265`) — and nothing outside `observeEidoverseWorld` ever reads it. `resolvePersistentMindPlaybookPhase` does not import the ledger at all (`persistentMindPlaybookSignals.js:24-33`). The two slices never connect.

`docs/features/persistent-mind-continuous-play.md:23,25` still describes the pre-7fd55add1 behavior (*"A federated peer is reporting new (non-steady) activity"*).

## Impact

The ladder collapses to a two-way toggle — `maintain` unless a peer is online, then `coordinate` — driven by nothing a Mind does. A fresh install is told to steward a Commons it has not built; a Mind on most wakes is told peers have activity waiting when nothing was measured, and it spends the wake visiting rather than building. The slice's headline acceptance — *"Wakes in a dense Commons spend more time maintain/coordinate than empty-explore filler"* — is trivially true because an empty Commons gets the identical phase to a dense one.

**Count what a Mind built, not what PortOS ships, and make `coordinate` gate on the signal its own acceptance criteria named.**

1. **Recalibrate density.** In `derivePersistentMindPlaybookPhaseSignals()` (`persistentMindPlaybookPhase.js:92`), stop counting install-constant rows. Drop `features`, `storage`, `operations`, and the `productivity`/`activity` summary rows from `districtEntries`, and add the populations that actually reflect a maturing Commons: locally-authored foundations, installed controllers, and Mind-placed entities. Keep the existing `null`-means-unknown degradation exactly as it is — that part is right.
2. **Re-pick the thresholds against the new scale.** `sparseDistrictCount: 4` / `matureDistrictCount: 16` were chosen for a count that started at 19; they are meaningless once the scaffolding is gone. Set them from the recalibrated signal and say in a comment what an empty install and a dense one actually measure.
3. **Feed the real coordinate signal.** Call `observeEidoverseWorld({ commit: false })` from `persistentMindPlaybookSignals.js` and derive `peersWithUnreadContributions` from `changes.newFoundations.length` / `changes.newPeers.length`. Gate `coordinate` on that, keeping reachability as a necessary-but-insufficient precondition (you cannot coordinate with a peer you cannot reach). `commit: false` matters — the picker must not advance the visit watermark the Mind's own `eidoverse.observe` call depends on.
4. **Fix the template either way.** Rewrite `PERSISTENT_MIND_PLAYBOOK_PHASE_INSTRUCTIONS.coordinate` (`persistentMindPlaybook.js:79-87`) so its heading and first sentence claim only what the signal measured.
5. **Update `docs/features/persistent-mind-continuous-play.md:23,25`.**

## Acceptance criteria

- [ ] A test fixture representing an empty install resolves to `explore`; one with a handful of authored foundations/controllers resolves to `construct`; a dense one to `maintain`. All three must be reachable with the shipped signal builder, not a hand-written signal object.
- [ ] A guard asserting `districtCount` responds to a Mind-authored foundation being added and does **not** respond to an instance feature being toggled.
- [ ] `coordinate` does not fire for a reachable peer with no unread contributions; it does fire when the observation marker reports new inherited foundations.
- [ ] The `coordinate` template's heading and opening sentence match the signal that selected it.
- [ ] The picker's observation read does not advance the visit marker (assert the marker is unchanged after a phase resolution).
- [ ] `docs/features/persistent-mind-continuous-play.md` describes the shipped behavior.

## Out of scope

- The `placeStatus` trail defect in the observation builder (#7633).
- Any change to what `buildEidoverseWorldSignals()` projects for the world itself — the districts should keep showing features and storage; only the *maturity* signal stops counting them.

## Why this was missed

`44efb5237` added a contract test pinning the picker to `buildEidoverseWorldSignals()`'s real projection shape — the right instinct, and it asserts nothing about *calibration*, so a signal that is structurally correct and numerically pinned at 19 passes it. Every fixture in the suite is a hand-built signal object with the counts the test author chose, which is exactly the range the real builder cannot produce.

Refs #7453, #7457, #7458.

atomantic:
## Summary

Two of the four continuous-play playbook phases were unreachable on every install, and the phase that did fire told the Mind something nobody measured.

**Density counted PortOS scaffolding, not what a Mind built.** `derivePersistentMindPlaybookPhaseSignals()` summed `features` (one row per registry feature whether enabled or not, ~15), `storage` (a fixed `database` + `filesystem`), and the single `operations` / `productivity` / `activity` summary rows. That put a floor of ~19 under a count whose `matureDistrictCount` was 16, so `explore` and `construct` were unreachable and a brand-new install was told its Commons was mature and instructed to "repair or retire what no longer serves the world before adding anything new."

The count now sums the genuinely-variable world populations (apps, active agents, active tasks, peers, goals, memory categories, Jira groups) plus the two durable Eidoverse populations a Mind grows itself: locally-authored foundations (inherited copies subtracted — federation is not construction) and installed controllers. Thresholds are re-picked against the new scale (`sparse: 3`, `mature: 12`), with the calibration written down beside them. The districts still *show* features and storage; only the maturity signal stops counting them.

**`coordinate` gated on reachability while claiming unread activity.** The signal was explicitly reachability, and the template the model reads said *"Coordinate (peers waiting) — Federated peers have activity worth your attention."* Because density was pinned above the mature threshold, that fabricated claim was the steady state for any install with one always-online peer. It now gates on the signal #7458's own acceptance criteria named: foundations *inherited* from a peer, or a new peer, since the Mind last looked — derived from `changes.newFoundations` / `changes.newPeers` in the durable observation marker #7457 already persists and nothing outside `observeEidoverseWorld` ever read. Reachability stays a necessary-but-insufficient precondition. A foundation the Mind authored itself is not a peer waiting, and a first observation (no marker to diff against) yields `null` — not measured — rather than a fabricated zero. The template's heading and opening sentence now claim only that.

**Marker safety.** The picker passes `commit: false`, so resolving a phase never advances the visit trail the Mind's own `eidoverse.observe` — and this very phase — depends on. `observeEidoverseWorld()` gained an optional `source` so the picker hands over the projection it already collected instead of fanning out across ~20 service reads twice per wake.

Docs updated to describe the shipped behavior.

## Decision recorded

The issue also named "Mind-placed entities" as a density input. There is no durable local store for them — `eidoverse.augment` sends verbs straight to the world runtime, and reading them back would require the running runtime that the observation builder deliberately does not require. Locally-authored foundations plus installed controllers are the durable proxy, and that is what shipped.

## Test plan

- `server/lib/persistentMindPlaybookPhase.test.js` — calibration now runs the **shipped** builders (`buildEidoverseWorldSignals()` + `buildEidoverseObservation()`), not hand-built signal objects. That is the root cause of the miss: every earlier fixture had counts the real builder could never produce. Empty install → `explore`, a handful of authored foundations → `construct`, dense → `maintain`, inherited arrival → `coordinate`. Plus a guard that density moves on an authored foundation and does **not** move when an instance feature is toggled or added.
- `server/services/persistentMindPlaybookSignals.test.js` — runs the **real** observation ledger against a temp data root and asserts the marker file is byte-identical after a phase resolution; asserts `collectEidoverseWorldSources` is called exactly once per wake; and walks the coordinate gate end to end (reachable peer, nothing unread → `maintain`; inherited foundation arrives → `coordinate`).
- `server/lib/persistentMindPlaybook.test.js` — the coordinate template's heading and opening sentence no longer claim "peers waiting" / "worth your attention".
- Green: the eidoverse + persistent-mind suites (53 files, 736 tests) and the tree-wide guard sweep.

## Review status

**Required local code review did not complete.** `provider:opencode-zen-cli` reports no enforced tool-free review transport on this install, and `provider:nvidia-nim` aborted on every attempt. Per the swarm contract this PR is intentionally left open and must not be merged until a required review completes.

Closes #7630


atomantic:

Issue #7454, opened 2026-09-16
Eidoverse: proposal–consequence separation for construction tools


atomanticagent:
## Parent

Part of #7453 (SwarmWorld-inspired stigmergic Eidoverse).

## Problem

Today `eidoverse.augment` / project / say can treat agent language as if it already changed the world. SwarmWorld’s core lesson: **agents propose; the world determines consequence**.

## Proposal

Introduce an explicit two-phase contract for world-mutating mind/CoS tools:

1. **Proposal** — structured intent (geometry/style/affordance/controller sketch + provenance).
2. **Consequence** — PortOS/Eidoverse validators + world service accept, rewrite, or refuse; only accepted consequences become durable world state.

Surface the split in tool results (`proposed` vs `committed`) so minds cannot claim side effects that did not land.

## Why

Stops transcript fiction from becoming “fake Commons,” and makes promote-to-baseline honest.

## Acceptance

- Mutating eidoverse tools return separate proposal vs commit outcomes.
- Continuous-play minds are instructed/enforced not to narrate uncommitted work as done.
- Tests cover refuse / rewrite / accept paths.

PR #7648, opened 2026-09-18
Read the controller supervisor's world-write refusal instead of counting it as delivered

### LINKED ISSUE #7628: A controller's refused world writes count as delivered, so a dead controller reports healthy forever
## Problem

#7454's whole point was that a Mind must not claim a build it did not get: *"Surface the split in tool results (`proposed` vs `committed`) so minds cannot claim side effects that did not land."* Commit 9e02fa468 delivered it — `augmentEidoverseWorld()` deliberately stopped throwing on a refusal and started returning the verdict as data (`server/services/eidoverseWorld.js:2102-2116`):

```js
const outcome = await commitAugmentProposal(...).catch((error) => {
  if (error?.name === 'AbortError') throw error;
  worldRefused = true;
  return refusedAugmentOutcome(proposal, { proposed: proposal.args, reason: ... });
});
...
return { success: committedCount === results.length, world, applied: committedCount, operations: results };
```

**The autonomous writer into the world never adopted it.** #7456's controller supervisor landed after 9e02fa468 and throws the envelope away (`server/services/eidoverseControllerRuntime.js:277-288`):

```js
for (const effect of outbound) {
  if (effect.kind === 'say') await world.sayInEidoverseWorld(effect.text, { signal });
  else await world.augmentEidoverseWorld(effect.operations, { signal });   // return value discarded
  delivered += 1;                                                          // counts sends, not commits
}
return { delivered, error: null };                                         // error is a hardcoded null
```

Because `augmentEidoverseWorld()` no longer throws on refusal, `error: null` is not a fallback — for an augment effect it is the **only** reachable value.

### The verdict dead-ends even when it is honest

The one field that could record a delivery problem, `lastOutcome.deliveryError` (`eidoverseControllerRuntime.js:407`), is read nowhere in the tree:

- `summarizeControllerInstall()` (`server/lib/eidoverseControllers.js:340-362`) omits both `delivered` and `deliveryError`, and it is the projection behind the route (`server/routes/eidoverseWorldRoutes.js:175-183`), the mind tool, and the panel.
- `observeControllers()` (`server/lib/eidoverseObservation.js:210-215`) filters `needsAttention` on `disarmedReason || consecutiveFailures > 0 || lastTickOk === false` — delivery is not in the filter.
- `EidoverseControllersPanel.jsx:61-63` prints a green `Last tick ok` off `install.lastTickOk`.
- `stepInstall()` resets `consecutiveFailures: 0` on a successful *step* (`eidoverseControllerRuntime.js:347`) and writes `lastOutcome: { ok: true, ... }` (`:353`). Delivery never touches that counter, so the `maxConsecutiveFailures` disarm at `:322-333` **cannot fire for a delivery failure at all.**

### The concrete case is the only shipped controller

`lanternKeeper.controller.js:78-84` emits `{ kind: 'augment', operations: [{ verb: 'light', args: { id: <configured id> } }] }` every tick. If that entity id is gone — the user deleted it, a projection reconcile renamed it — the world refuses. If the configured id fails `validEntityId`, `proposeAugmentOperation()` refuses locally (`eidoverseWorld.js:2058-2066`). Either way: `applied: 0`, `success: false`, and the ledger records `ok: true, delivered: 1, deliveryError: null`, every tick, forever.

### The test fixture demonstrates the hole

`server/services/eidoverseControllerRuntime.test.js:171` injects `const deliver = vi.fn(async () => ({ delivered: 1, error: null }))`, and `:188` — *"counts the step and records the failure when the world cannot be reached"* — mocks a **throw** to exercise `deliveryError`. The suite tests an error channel through a seam the real `deliverControllerEffects()` can never populate for an augment effect, and asserts `lastOutcome` is `{ ok: true, delivered: 0, deliveryError: 'world host is down' }` — a permanently-failing controller pinned as a passing tick, with no assertion that the failure ever reaches a surface.

## Impact

This is the surface the epic calls **executable inheritance**: *"Artifacts should keep working between wakes / after the author leaves."* What is verified is that `step()` still increments a counter. Whether the world is still being maintained — the only thing that makes the claim true — is unobservable.

A Mind waking to `eidoverse.observe` is told its maintenance hooks are healthy and plans on that premise. And a `controller` foundation can be promoted to the shared baseline on assay evidence that only ever replays `step()` (see #7625), never delivery — so a controller that has never successfully touched any world can be published as inherited infrastructure and look alive on every peer's ledger.

**Read the envelope, and let delivery drive health.**

1. In `deliverControllerEffects()` (`eidoverseControllerRuntime.js:277`), capture the result. Treat `result.success === false` as a delivery failure carrying the first non-`accepted` entry's reason from `result.operations`, and count `delivered` from `result.applied`, not from the loop counter. Treat `rewritten` as delivered-but-noted (the world landed different args), not as a failure.
2. Add `consecutiveDeliveryFailures` to the record and feed it into the existing disarm decision in `stepInstall()`, so a controller the world has refused `maxConsecutiveFailures` times in a row disarms with a reason — the same treatment a throwing `step()` already gets.
3. Add `lastDelivery { ok, delivered, reason }` to `summarizeControllerInstall()` (`server/lib/eidoverseControllers.js:340`) so the route, the mind tool and the panel all inherit it from the one projection.
4. Add `|| install.lastDelivery?.ok === false` to the `needsAttention` filter in `observeControllers()` (`eidoverseObservation.js:213`), and render it in `EidoverseControllersPanel`'s `TickOutcome`.
5. Rewrite the test at `eidoverseControllerRuntime.test.js:188` so the seam resolves `{ success: false, applied: 0, operations: [{ outcome: 'refused', reason }] }` through the **real** `deliverControllerEffects()`, and assert the refusal reaches `observeControllers()`.

## Acceptance criteria

- [ ] A controller whose every augment is refused disarms after `maxConsecutiveFailures` with a reason naming the refusal.
- [ ] `eidoverse.observe` lists that controller under `controllers.needsAttention`.
- [ ] The Controllers panel shows the delivery failure rather than a green `Last tick ok`.
- [ ] `rewritten` is distinguished from `accepted` in the recorded outcome and does not count as a failure.
- [ ] A test drives a refusal through the real delivery function, not through a mocked `error` field.

## Out of scope

- What the promote-gate assay replays (#7625).
- Reading back or revalidating a controller's config (#7629).

## Why this was missed

#7454 fixed the honesty gap for the *tool* path — a Mind calling `eidoverse.augment` now gets per-operation outcomes and is instructed not to narrate a refused build as done (`server/services/persistentMindOrientation.js`). #7456 then built a second, autonomous writer into the same world two commits later and did not consume the contract the first slice had just created. Both slices tested their own half and both passed.

Refs #7453, #7454, #7456.

atomantic:
## Summary

`augmentEidoverseWorld()` returns a verdict on refusal instead of throwing (#7454's fix for the tool path), but the controller supervisor's `deliverControllerEffects()` (`server/services/eidoverseControllerRuntime.js`) discarded that return value entirely and counted every send as delivered no matter what the world actually did. A controller whose configured entity id is gone — the shipped `lantern-keeper` example — would step cleanly forever while every write it proposed was refused, and `lastOutcome` would read `ok: true, delivered: 1, deliveryError: null` on every tick.

Changes:

- `deliverControllerEffects()` now reads the world's own `success` / `applied` / per-operation `outcome` verdict. `delivered` is the world's committed count, never the loop's attempt count; a `rewritten` operation counts as delivered (the world landed different args, not nothing), only `refused` does not.
- A new `consecutiveDeliveryFailures` counter (schema + persisted record) accumulates across refused deliveries and disarms the controller after `maxConsecutiveFailures`, in `recordDeliveries()` — the same treatment a throwing/refusing `step()` already gets. A disarm here reconciles the supervisor schedule the same way a step-disarm does.
- `summarizeControllerInstall()` exposes the verdict as `lastDelivery: { ok, delivered, reason }`, kept separate from the step's `lastTickOk` — `ok: null` covers "delivery is off" or "nothing to deliver this tick", never collapsed into `false`.
- `observeControllers()` now also flags `lastDelivery.ok === false` into `controllers.needsAttention`, so `eidoverse.observe` surfaces a controller that steps clean but whose writes are all being refused.
- The Controllers panel renders "Delivery refused: …" instead of a green "Last tick ok" when the step succeeded but delivery didn't.

## Test plan

- `server/services/eidoverseControllerRuntime.test.js` — new test drives a world refusal through the **real** `deliverControllerEffects()` (mocking `eidoverseWorld.js`'s `augmentEidoverseWorld`/`sayInEidoverseWorld` return values rather than injecting a custom `deliver` stand-in), asserting the step reads `ok: true` while delivery disarms the controller after 3 refusals with a reason naming the refusal. A second new test confirms a `rewritten` delivery does not disarm.
- `server/lib/eidoverseObservation.test.js` — new test confirms a controller with a clean step but `lastDelivery.ok === false` is surfaced under `controllers.needsAttention`.
- `client/src/components/eidoverse/EidoverseControllersPanel.test.jsx` — new test confirms the panel renders "Delivery refused: …" instead of "Last tick ok" for that same shape.
- Ran `server/services/eidoverseControllerRuntime.test.js`, `server/lib/eidoverseObservation.test.js`, `server/lib/eidoverseControllers.test.js`, `server/routes/eidoverseWorldRoutes.test.js`, `server/lib/eidoverseFoundations.test.js`, `server/services/eidoverseResilienceContributions.test.js`, and `client/src/components/eidoverse/EidoverseControllersPanel.test.jsx` — all green.
- `server/services/eidoverseObservationLedger.test.js` has one pre-existing flaky timeout (`serializes concurrent observations so one marker write cannot lose the other`) unrelated to this diff — it fails the same way on an isolated run and this PR touches none of that file's locking code.

## Review status

Both configured reviewers are optional and both came back inconclusive on a second, independent pass:

- `provider:opencode-zen-cli` — refuses to run: "This provider has no enforced tool-free review transport." No enforceable read-only mode, so it is unavailable rather than a reason to fall back to an unsandboxed run.
- `provider:nvidia-nim` — provider-side failure on all three attempts (twice "malformed (non-JSON or unexpected) response body", once "fetch failed").

The diff was instead validated directly against the contracts it depends on: `augmentEidoverseWorld()` always returns `{ success, applied, operations }` (`server/services/eidoverseWorld.js`), so `result.applied` and the `refused` scan can never read off an absent field; `withStoreLock` (`server/lib/asyncMutex.js`) returns its callback's value, so `recordDeliveries()` returning `{ gateMoved }` through it reaches `tickOnce`; and `lastOutcome.effects` / `deliverEffects` both exist on the record shape `summarizeLastDelivery()` reads. No findings.

Tests re-run green on the branch: `eidoverseControllerRuntime`, `eidoverseObservation`, `eidoverseControllers`, `eidoverseWorldRoutes` (73 passed) and `EidoverseControllersPanel` (7 passed). CI is green.

Closes #7628


atomantic:

PR #7649, opened 2026-09-18
show a catalog ingredient's source scrap by title and let one click list its siblings

### LINKED ISSUE #7617: show a catalog ingredient's source scrap by title and let one click list its siblings
## Problem / Goal

A committed ingredient shows its source as a bare `cat-scrap-<uuid>` monospace string (`client/src/pages/CatalogIngredient.jsx:1438`), and there is no way to ask "what else came out of this piece?". The catalog already has two grouping seams — refs (universe / series / work) and sources (scrap) — so the fix is to promote the scrap that already groups these rows, **not** to invent a third parallel taxonomy (an `autobiography` record type is explicitly the wrong answer: a third taxonomy is what makes the catalog feel arbitrary).

## Approach

1. `GET /api/catalog/ingredients/:id/details` (`server/routes/catalog.js:239-264`) returns source rows as `{ scrapId, span, extractedAt }` with no title. Join `catalog_scraps.title` onto them, and add the sibling extractions for each source scrap via `listSourcesForScrap` (`server/services/catalogDB/refs.js:38`) so the detail page needs no second round-trip.
2. `client/src/pages/CatalogIngredient.jsx:1438` renders the scrap **title** as a link, plus a "From the same source" list of sibling ingredients.
3. Add `scrapId` to `catalogIngredientQuerySchema` (`server/lib/catalogValidation.js`), filter on it through `catalog_ingredient_sources` in `listIngredients` (`server/services/catalogDB/ingredients.js`), and wire `?scrap=<id>` into `client/src/pages/Catalog.jsx` — **selection lives in the URL** per `client/src/AGENTS.md`, so "everything extracted from this piece" is one click and one shareable link.

## Acceptance criteria

- [ ] Ingredient detail shows the source scrap's title (not the raw id) and its sibling ingredients, from a single `/details` response.
- [ ] `/catalog?scrap=<id>` filters the catalog to that scrap's extractions and survives a reload / share.
- [ ] An ingredient with no source scrap, and a scrap with exactly one extraction, both render without an empty "From the same source" stub.
- [ ] Route tests cover the `scrapId` filter and the joined title; the client test covers the link and the sibling list.
- [ ] The catalog docs describe the source-scrap grouping.

## Out of scope

- Per-chunk source provenance: `commitScrap` links to the parent scrap only and the extractor never emits `span`. Pre-existing, unchanged here.

atomantic:
## Summary

- `GET /api/catalog/ingredients/:id/details` now joins the source scrap's `title` onto each `sources` row and batches every sibling ingredient extracted from the same scrap into the same response (`listSiblingIngredientsBySource`), so the detail page needs no second round-trip to answer "what else came out of this piece?"
- `CatalogIngredient.jsx`'s "Source scraps" panel renders the scrap's **title** (not the raw `cat-scrap-<uuid>`) as a link to `/catalog?scrap=<id>`, plus a "From the same source" list of sibling ingredients — omitted entirely when a scrap has no other live extractions, and the "Created manually" message stays for an ingredient with no source scrap at all.
- Added a `scrapId` filter to `GET /api/catalog/ingredients` (validated in `catalogIngredientQuerySchema`, applied in `listIngredients` via `catalog_ingredient_sources`) — a provenance dimension orthogonal to the universe/series ref filters, so it composes with them instead of joining their mutual-exclusivity rule.
- Wired `?scrap=<id>` into the Catalog page: it filters the grid to that scrap's extractions, shows a "Filtered by source" banner (with the title when it arrives via the sibling link's router state) and a Clear action, and survives a reload/share since the filter itself is URL-driven.
- `docs/STORAGE.md` documents the `catalog_scraps` / `catalog_ingredient_sources` grouping seam alongside the existing ref-based grouping.

## Test plan

- `cd server && npm run test:db` — `services/catalogDB.test.js` (title join + `listSiblingIngredientsBySource`) and `routes/catalog.test.js` (`/details` siblings/title, `scrapId` filter) pass.
- `cd server && npx vitest run lib/catalogValidation.test.js` — `scrapId` schema coverage passes.
- `cd client && npx vitest run src/pages/Catalog.test.jsx src/pages/CatalogIngredient.test.jsx src/services/apiCatalog.test.js` — filter banner, sibling link rendering, and API passthrough pass.

Closes #7617


atomantic:

atomantic:
Review-loop note: both configured local reviewers (`opencode-zen-cli`, `nvidia-nim`) were unreachable this run — `POST /api/code-review/local` returns `401 AUTH_REQUIRED` because this install has the optional instance password enabled and the review-loop curl recipe sends no credential. Both are configured as optional reviewers, so a provider/transport failure is non-blocking.

Verified instead: full CI green (server, client, DB, lint, plan-test-impact) plus a manual read of the diff — the `scrapId` predicate composes correctly with the existing ids/type/tag/q/ref branches and uses the right placeholder index; `catalog_ingredient_sources` has no soft-delete column, so the sibling query's `i.deleted = false` is the only liveness filter needed; `listSourcesForIngredient`'s `LEFT JOIN` aliases `title` to `scrap_title`, so no column collides with `rowToSource`'s mapping.

atomantic:
Follow-up to the note above: after merging I retried both reviewers through the auth-independent stdin bridge (run from the primary checkout, since `resolveInstallRoot()` refuses a data-root pin under `data/cos/worktrees/`). Both are genuinely unavailable, not just gated:

- `provider:opencode-zen-cli` — `This provider has no enforced tool-free review transport. Select its API mode or a supported reviewer harness.` Structural: it fails `supportsPublicReviewProvider()`, so there is no read-only transport to run it in.
- `provider:nvidia-nim` — `Provider request failed: fetch failed` on three consecutive attempts. Provider-side outage.

Both are configured optional, so neither blocks. The merge rests on green CI plus the manual contract check in the previous comment.

PR #7651, opened 2026-09-18
Let the Mind retire a decision it made: an addressable journal with supersede and resolve

### LINKED ISSUE #7623: [repo-study-nautilo-mind-decision-journal] Let the Mind retire a decision it made: an addressable journal with supersede and resolve
## Problem

The Persistent Mind remembers things, but it does not remember what it is *on the hook for*. Its durable context is two shapes, and neither tracks obligation or reversal:

- **Rollups** (`server/services/persistentMindContext.js`, `data/cos/persistent-mind-rollups.json`) are free prose summaries sealed over a sequence range. Nothing inside one is addressable, so nothing inside one can be retired.
- **Memories** (`docs/features/memory-system.md`) are flat facts with a type, an importance that decays, and a protection level. There is no `supersede` in `server/services/memory*.js` or `server/routes/memory.js` — nothing anywhere. When the user changes their mind, the old belief does not get retired; it sits beside the new one and both decay on the same clock, and the more-accessed stale one can outrank the correction.

The user-visible consequence: ask the Mind "what did we decide about X, and what's still open?" and it answers from prose summaries and a bag of facts. A commitment it made three days ago is not tracked as outstanding. A decision the user reversed yesterday is still quoted as current.

## What to build

A typed, individually addressable journal for the Mind's own conversation, extracted at the boundary that already seals rollups, and displayed as a panel the user can read and correct.

**Event kinds** — start with: `decision`, `commitment`, `open_question`, `risk`, `goal`, `preference`. Each event carries a concise statement (not a transcript quote), a status of `active | superseded | resolved`, a stable sequence reference, and the message sequences it was drawn from.

**Three operations, not one.** The extraction step may emit `append`, `supersede` (this new statement replaces event E), and `resolve` (event E is settled). Supersession is what the memory store cannot express today: the old event stays readable as history with `status: superseded` and a pointer to what replaced it, and stops being quoted as current. This is a status transition, never a deletion — the same shape `AGENTS.md` already requires for sync-safe clears.

**Source-grounded, and bounded.** Every operation must cite the message sequences it came from; an operation citing nothing is rejected rather than stored. Cap the batch (five operations per extraction) and the statement length. Zero operations is the normal outcome — greetings, acknowledgements, tool chatter, and raw reasoning produce nothing.

**The rollup compacts from the journal, not from raw turns.** Once the journal exists, the sealed summary preserves active decisions, outstanding commitments, unresolved questions and risks, and the resolved history needed to explain current state — and drops superseded wording. That is a strictly better summary than compressing the transcript, and it is why this is worth building rather than adding two more memory types.

**Treat the journal as evidence, never as instructions.** Prior events fed back into the extraction prompt are untrusted data. Say so in the prompt contract and test it with an event whose statement contains an injected instruction.

## Storage decision (made — do not re-open)

Machine-local file store beside the rollups, in `server/services/persistentMindContext.js`'s `data/cos/` neighbourhood, with a `schemaVersion` and a bounded retention like `PERSISTENT_MIND_TRAJECTORY_LIMITS`. The journal quotes private conversation, so the [machine-local privacy ADR](docs/decisions/2026-08-08-privacy-records-machine-local.md) keeps it off the federation layer entirely — no `PORTOS_SCHEMA_VERSIONS` entry, no sync cursor, no tombstone. The rollup store already sets exactly this precedent (`docs/STORAGE.md`, CoS event ledgers). A migration that derives nothing ships no `data.reference/` seed; an absent file is the correct empty state.

## Fix

- `server/services/persistentMindContext.js` — a `persistent-mind-journal.json` store with the same `createFileWriteQueue` + `atomicWrite` discipline as `recordPersistentMindRollup`, and `append` / `supersede` / `resolve` writers.
- New `server/services/persistentMindJournal.js` — the extraction call and its output validation (closed schema, rejected unreferenced operations, one repair retry then give up). Run it at the same boundary where `preparePersistentMindContext` seals a range, reusing the same summarizer provider resolution so no new provider path and no cold-bootstrap call is introduced (`AGENTS.md` AI Provider Usage Policy).
- Prompt body under `server/services/taskPromptDefaults/` with a `PROMPT_VERSIONS` bump plus `node scripts/regen-prompt-integrity-snapshot.js`, per the prompt-default contract in `AGENTS.md`.
- `server/routes/cosMindRoutes.js` — `GET /mind/journal` (paged, filterable by kind and status) and a user-correction verb so the user can resolve or retire an event the Mind got wrong.
- `client/src/components/cos/PersistentMindContextPanel.jsx` (or a sibling mounted from `client/src/components/cos/tabs/MindTab.jsx`) — a journal view grouped by kind, superseded events collapsed behind a toggle, each event showing what replaced it.
- Tests: an operation with no source reference is rejected; supersede flips status and never deletes; resolve is idempotent; the rollup drops superseded wording and keeps outstanding commitments; an injected instruction inside a prior event does not change behaviour; a malformed model response repairs once and then fails without writing.

## Provenance

Studied repository: https://github.com/agentsea/nautilo, **MIT License**, studied 2026-09-18. Clean-room reimplementation of the technique (a typed, individually addressable conversation journal with append/supersede/resolve operations and source-grounded statements, compacted into the running summary). No Nautilo source, prompt text, schema, or prose is copied.

## Dispatch rationale

`model:heavy` — supersession semantics, a new prompt contract with a versioned default, an untrusted-evidence boundary, and a storage-placement decision that has to hold against the privacy ADR. `effort:high` — the hard part is per-step judgment about what counts as an event and how a bad extraction fails closed, not breadth of edits.

## Scope

Medium-to-large: one new service, one store, one prompt default with a version bump, two routes, one client panel. No federation surface and no migration.

atomantic:
## Summary

The Persistent Mind remembered things but not what it was *on the hook for*. Rollups are sealed prose with nothing addressable inside them, and memories are flat facts on one decay clock with no `supersede` anywhere in the store — so when the user changed their mind, the old belief sat beside the new one and the more-accessed stale one could outrank the correction.

This adds the third shape: a typed, individually addressable journal of what the mind decided, committed to, is still asking, and is watching, extracted at the same boundary that already seals a rollup.

- **Three operations, not one.** `append`, `supersede` (this statement replaces entry E) and `resolve` (E is settled). Supersession is the thing the memory store cannot express: the retired entry stays readable with `status: superseded` and a pointer to what replaced it, and stops being quoted as current. It is a status transition, never a deletion.
- **Source-grounded per operation.** Every operation must cite the message sequences it came from; one citing nothing — or citing a range the extraction was never shown — is discarded while its well-grounded siblings in the same answer still apply. Five operations per extraction, and zero is the normal outcome.
- **The rollup compacts from the journal**, not from the raw range, which is why `PERSISTENT_MIND_ROLLUP_PROMPT_VERSION` goes to 2: outstanding commitments and unresolved questions survive compaction as facts, and superseded wording is absent by construction. Existing v1 rollups read as stale and the next wake re-seals them.
- **Untrusted evidence.** Prior entries fed back into the extraction prompt are data, not instructions. Every stored statement is JSON-quoted so an injected heading or directive cannot forge a prompt section, and the prompt says so in words.
- **Fails closed.** A malformed model answer gets one narrow repair attempt and then gives up *without writing*; a failed extraction never blocks the seal; a corrupt store throws rather than resetting to `[]` (after raw retention it is the only record of what the mind still owes).
- **Same route, no new provider path.** Extraction runs on the pinned provider and the same per-call boundary as the summary it feeds — no cold-bootstrap call. Receipts get a `journal` purpose.

**Storage** is a machine-local file beside the rollup cache (`data/cos/persistent-mind-journal.json`). It quotes one human's private conversation, so the machine-local privacy ADR keeps it off the federation layer entirely: no `PORTOS_SCHEMA_VERSIONS` entry, no sync cursor, no tombstone, no `data.reference/` seed and no migration — an absent file is the correct empty state. The `history` cleanup scope clears it along with the messages it cites; a context-only clear does not. Documented in `docs/STORAGE.md`.

**UI.** Mind → Journal groups active entries by kind, collapses retired ones behind a toggle that names what replaced each, and lets the user settle or retire an entry the mind got wrong. `GET /api/cos/mind/journal` (filterable by kind/status) and `POST /api/cos/mind/journal/:id/correct` back it.

## Test plan

- `server/services/persistentMindJournal.test.js` (new): an unsourced or out-of-range operation is rejected while its grounded sibling applies; an empty batch writes nothing at all; supersede keeps three records where two were and links both directions; resolve is idempotent and a retired entry cannot be re-opened; the user's retire is a status flip with `retiredBy: 'user'`; a malformed answer repairs exactly once then fails without writing; a refused call leaves the range unextracted; an injected directive inside a stored entry is quoted and cannot forge a second `# Your task` section; the sealed summary carries outstanding commitments and settled history but never superseded wording; a damaged store fails closed with the original bytes intact.
- `server/services/persistentMindContext.test.js`: extraction runs *before* the seal, the journal reaches the summarizer and the next assembled context; a failed extraction still seals the range.
- `server/routes/cosMindRoutes.test.js`: the journal serves the public projection only (no stored envelope), filters validate, and the correction verb accepts exactly `resolve`/`retire` — no free status write, no delete.
- `client/src/components/cos/PersistentMindJournalPanel.test.jsx`: a retired entry leaves the live list, stays readable behind the toggle and names its replacement; the user's retire swaps state reactively without a refetch; a load failure surfaces instead of an empty journal.
- Full `server` suite (2301 files) and full `client` suite (993 files) green.

Closes #7623


atomantic:

PR #7653, opened 2026-09-18
replay the foundation being promoted, not the contribution its author named (#7625)

### LINKED ISSUE #7625: Eidoverse promote gate: the agent-free assay never replays the foundation being promoted
## Problem

Epic #7453 made one promise load-bearing for federation: *"Promote-to-baseline only after a contribution still works without its creator narrating it."* #7460 built the harness for it, and the harness is good — `runResilienceAssay()` refuses async steps, re-creates the sandbox per scenario, runs a fixed disturbance suite, and the ledger runs it server-side rather than accepting a caller's verdict (#7455).

**It is pointed at the wrong object.** The assay never replays the foundation being promoted. It replays whatever module the author *named*.

The binding between "what got evaluated" and "what gets promoted" is a single author-supplied string:

- `contributionId` is free text on the authoring surface — `eidoverseFoundationInputSchema` (`server/lib/eidoverseFoundations.js:254`), accepted from the `eidoverse.record` mind tool (`server/services/cosToolRegistry.js:373`) and from the Foundations panel, which renders it as a **dropdown of shipped ids** (`client/src/components/eidoverse/EidoverseFoundationsPanel.jsx:425`).
- `assayEvidenceRefusal()` (`server/lib/eidoverseFoundations.js:347`) checks that the evidence matches the id the foundation *names* — the "borrowed-credential" check its own comment describes. Nothing checks that the named contribution has any relationship to the foundation's `body`.
- Contributions resolve only against two **shipped** sources (`server/services/eidoverseResilienceContributions.js`): the fixture directory in `server/services/eidoverseResilienceAssayFixtures/`, and the shipped controller registry. Neither reads anything from `data/`, so **no install can register a contribution for its own work.**

Those combine into a closed loop with no honest path through it. Every locally-authored foundation must name a shipped contribution, and naming one is sufficient to pass.

### Evidence: the repo's own test does it

`server/services/eidoverseFoundationLedger.test.js:31-41` is the canonical fixture, and it is the hole:

```js
const authored = (overrides = {}) => ({
  id: 'tide-beacon', kind: 'controller', title: 'Tide Beacon',
  contributionId: 'beacon-relay-demo',
  body: { schema: { pulses: 'integer' }, affordance: { inspect: 'reads the pulse count' } },
  ...
});
```

`beacon-relay-demo` (`eidoverseResilienceAssayFixtures/beaconRelay.contribution.js`) hardcodes its own `worldState` and reads no foundation at all. "Tide Beacon" packages, promotes to `baseline`, and is offered to peers at `GET /api/peer-sync/eidoverse-foundations` carrying `assay.pass: true` — evidence about a demo fixture, not about Tide Beacon. Every peer re-runs the accept-side gate and reaches the same conclusion, because the envelope is internally consistent; it is just not about the thing inside it.

### The happy path has the same defect, one level quieter

Even for the intended `controller` case, `controllerAssayContributions()` (`server/services/eidoverseControllerRegistry.js:129-150`) builds its sandbox from `definition.configSchema.parse(definition.exampleConfig ?? {})` — **upstream's example config**, never the config this install actually runs. An install whose controller config is broken still promotes on evidence that upstream's example survives.

### Impact

- The one gate standing between a local build and the shared federated baseline can be satisfied without evaluating the build. This is the exact proposal-vs-consequence collapse #7454 fixed for construction tools, still open at the promote step.
- Three of the four declared `EIDOVERSE_FOUNDATION_KINDS` (`schema`, `affordance`, `district-template`) have no replayable representation at all, so for them "pass the assay" can *only* mean borrowing.
- Peers inherit on this evidence, so a false pass propagates across installs.

**Derive the sandbox from the foundation's own `body`. Remove `contributionId` from the authoring surface.** The executable part stays shipped code, so this adds no arbitrary-code-execution surface — what varies is the declarative input the install actually authored.

Add `server/lib/eidoverseFoundationSandbox.js` exporting `sandboxFromFoundation({ kind, body })`, returning the `{ worldState, controller, projectionSource }` shape `runResilienceAssay()` already consumes, or a refusal reason:

- **`controller`** — `body.controller.definitionId` names a shipped definition; the sandbox runs `definition.configSchema.parse(body.controller.config)`. The install's real config is what gets replayed. A body naming no definition, or one whose config fails the definition's schema, refuses.
- **`district-template`** — replay through `buildProjectionPlan()` with `body` as the projection source across the disturbance suite (a template's consequence *is* its projection plan).
- **`schema` / `affordance`** — replay through the interpreter that consumes them: parse the declared schema and resolve declared affordances against a clean world state, asserting the same serializability/invariant checks per tick. A declaration that cannot be interpreted without its author is not promotable, which is the correct answer.

Then:

- `contributionId` leaves `eidoverseFoundationInputSchema`, the `eidoverse.record` tool schema, and the panel field. It stays on the record/candidate as a **derived** label (`kind:<definitionId>` or `kind:<id>`) so `assayEvidenceRefusal()`'s binding check keeps working unchanged.
- `eidoverse.contributions` (`cosToolRegistry.js:372`) stops advertising ids to bind to. Either drop it or repoint it at "which controller definitions exist", which is what a mind actually needs.
- `eidoverseResilienceContributions.js` keeps resolving shipped fixtures for the CLI and the harness's own tests; it is no longer the promote path's resolver.
- Bump `EIDOVERSE_FOUNDATION_CANDIDATE_VERSION` to `2` — v1 envelopes carry assay evidence with the old semantics, so the accept side in `services/sharing/peerEidoverseFoundationSync.js` must refuse them rather than inherit on evidence that never described the body.
- Migration in `scripts/migrations/` for `data/eidoverse/foundations.json`: any record whose sandbox is not derivable drops `assay`/`candidate` and returns to `layer: 'vernacular'` with `promotedAt: null` (the ledger already does exactly this on re-authoring, and the rationale is identical — the verdict no longer describes the body). Machine-local file, no `data.reference/` seed, so gate on the input's presence per the root AGENTS.md migration rule.

## Acceptance

- No caller-supplied `contributionId` on any authoring surface (mind tool, route, panel); the record's value is derived.
- A `controller` foundation is replayed against **its own** `body.controller.config`; a test proves a foundation with a broken config fails while the shipped definition's `exampleConfig` still passes on its own.
- Each of the four kinds has a sandbox derivation with one passing and one deliberately failing fixture, in the `*.contribution.js` / `*.failing.fixture.js` convention already in use.
- Regression test for this hole specifically: a foundation whose `body` is unrelated to any shipped fixture cannot package, where today's `authored()` fixture does.
- `eidoverseFoundationLedger.test.js`'s `authored()` fixture is rewritten to carry a real, self-describing body.
- Candidate version bumped; accept side refuses v1 envelopes with a readable reason.
- Migration lands with the de-promotion behavior above and a test.

## Why this was missed

Worth recording, because it is the same failure in a different register: #7453 asked for creator-free evaluation before anything promotes, then shipped and closed itself on a creator-written success declaration. The epic's closing comment asserts "the server runs the resilience assay itself rather than accepting a caller's verdict" — true, and not the thing that needed checking. Nothing independent ever asked *what the assay was replaying*.

Refs #7453, #7455, #7460.

atomantic:
## Summary

The Eidoverse promote gate ran the agent-free resilience assay server-side, but pointed it at whatever module the author named in `contributionId` — free text on the route body, the `eidoverse.record` mind tool and the Foundations panel, resolved against the shipped assay-fixture directory. Naming a shipped fixture was therefore *sufficient* to clear the gate: a local build could promote to `baseline` and be offered to every federated peer carrying `assay.pass: true` about a demo fixture rather than about the body inside the envelope. Peers re-ran the accept-side gate and agreed, because the envelope was internally consistent — it just was not about the thing inside it. The repo's own canonical ledger fixture did exactly this.

**The sandbox is now derived from the foundation's own `body`** (`server/lib/eidoverseFoundationSandbox.js`), one derivation per declared kind:

| kind | what gets replayed |
| --- | --- |
| `controller` | the shipped definition named by `body.controller.definitionId`, against **this install's** `body.controller.config` through the definition's own `configSchema`. The old path used upstream's `exampleConfig`, so an install with a broken config still promoted on evidence that upstream's example survived. |
| `district-template` | `body.placement` must reproduce from its own `{ layoutId, anchor, propCount, seed, facing }`; the body is the harness's projection source. |
| `schema` / `affordance` | a declarative interpreter over the declared field types — a clean world state built from `body.schema`, every `body.affordance` verb resolved against it each tick. An affordance recorded as prose is refused. |

Every executable part is still code PortOS ships, so this adds no arbitrary-code-execution surface — what varies is the declarative input the install authored.

Also in this PR:

- `contributionId` leaves `eidoverseFoundationInputSchema`, the `eidoverse.record` tool schema and the panel field, and survives as a **derived** label (`controller:<definitionId>` / `<kind>:<id>`) so the existing binding check works unchanged. A receiving peer re-derives it from the envelope's own `kind`/`body` and refuses one whose evidence is about something else.
- `EIDOVERSE_FOUNDATION_CANDIDATE_VERSION` → **2**. A v1 envelope is refused with a reason that explains why, not `invalid literal value`.
- `GET /api/eidoverse/world/contributions` and the `eidoverse.contributions` mind tool are **removed** — a list of ids that were sufficient to pass the gate was a list of ways around it. `eidoverse.controllers` / `GET /controllers` already tell an author which behaviours exist, and the panel now seeds the body per kind and fills a `controller` body from a shipped definition.
- `buildDistrictTemplateFoundationDraft` no longer takes a `contributionId`; a test pins that the draft it produces is replayable by the new gate.
- `scripts/migrations/395-eidoverse-foundation-derived-assay.js` re-derives every record's label, drops the verdicts the old replay produced, and returns an unreplayable `baseline` foundation to `vernacular` with `promotedAt: null` — exactly what the ledger already does on re-authoring. It gates on the input file's presence; `data/eidoverse/foundations.json` is declared in `migrationOwnedPaths.js` so no seed can land ahead of it.
- `eidoverseResilienceContributions.js` keeps resolving shipped fixtures for the CLI and the harness's own tests; it is no longer the promote path's resolver.

## Test plan

- `server/lib/eidoverseFoundationSandbox.test.js` (new) — one passing and one deliberately failing body per declared kind, replayed through the real assay: a controller config the shipped definition rejects fails while its own `exampleConfig` still passes; a district template with one hand-edited coordinate is refused; an affordance recorded as prose is refused; a well-formed but fragile controller fails the disturbance suite rather than being refused up front; the shipped creative toolkit's own draft replays clean.
- `server/lib/eidoverseFoundations.test.js` — fixture rewritten to a self-describing body; new cases for the v1 refusal reason and for an envelope whose binding label its own body does not derive.
- `server/services/eidoverseFoundationLedger.test.js` — `authored()` rewritten to a real `controller` body naming a shipped definition; a body unrelated to anything replayable cannot package, where the old fixture did.
- `scripts/migrations/395-…test.js` (new) — de-promotion, label re-derivation, inherited copies, idempotence, and an unparseable ledger left untouched.
- Route, mind-tool and panel suites updated for the removed surfaces.
- Full `server` and `client` suites run locally. Two unrelated failures, both reproducing on `origin/main` or under parallel load only: `services/eidoverseObservationLedger.test.js` (concurrent-marker test, fails on `main` in a clean checkout) and `routes/settings.secretsStrip.test.js` (passes in isolation).

Closes #7625


atomantic:

Issue #7632, opened 2026-09-18
A promoted Eidoverse foundation can never be retracted: the peer offering is add-only with no tombstone


atomantic:
## Problem

Once one peer pulls a promoted foundation, that copy is permanent. There is no retraction, no tombstone, and no delete path for a foundation anywhere in the tree — local or inherited.

### The transport only adds

`applyOffering()` iterates the offering and calls `recordEidoverseFoundationInheritance()` (`server/services/sharing/peerEidoverseFoundationSync.js:240-251`). Nothing reconciles held-but-no-longer-offered records. The sweep's own comment states the asymmetry outright — the forced re-pull exists *"so a LOCAL loss (a hand-deleted ledger entry) self-heals"* (`:194-196`). It converges upward toward the sender's offering and never downward.

Withdrawal on the origin *is* possible and does nothing downstream: re-authoring drops `baseline` → `vernacular` and clears `candidate` (`server/services/eidoverseFoundationLedger.js:174-203`), so `listPromotedFoundationCandidates()`'s filter at `:133` stops serving it. The receiver sees a changed `listHash`, re-pulls, and **keeps the stale copy at `layer: 'baseline'` forever.**

### There is no delete at all

`server/routes/eidoverseWorldRoutes.js` has `router.delete` only for `/controllers/:id` (`:211`). No delete for `/foundations`. The ledger module exports none. A tree-wide search for a foundation delete or retract path returns nothing.

The repo already has the mechanism this skipped: `server/lib/tombstones.js` exists precisely for add-only sync merges and is used by `server/services/sharing/annotationsSync.js:145-175`.

## Impact

**No recall, across every install that pulled.** The federation-safety scan that gates a promote (`federationSafetyFindings`) is regex-shaped — it matches home paths, IPs, `*.ts.net` / `.local` hosts, MACs, emails, phone numbers and labelled GPS. It has no notion of a person's name, a Mind's name, or private record prose. A foundation promoted with something the author regrets — a personal detail the scan cannot recognize, a wrong attribution, a body that turns out to be harmful — stays on every install that pulled it, indefinitely, and is surfaced to that install's Minds through `eidoverse.observe`. The receiving user cannot remove it except by hand-editing `data/eidoverse/foundations.json`.

[...]

Issue #7626, opened 2026-09-18
Nothing on any install interprets a foundation body, so inheriting a peer's contribution stores a row and nothing more


atomantic:
## Problem

Epic #7453's headline success signal was: *"A mind can leave something that another mind (or a later wake) **uses** without reading the original transcript."* #7455 built the promote path, #7461 the provenance graph, and #7455's transport slice the peer pull. The epic closed asserting *"A peer can adopt a promoted foundation while keeping local style."*

**Nothing adopts anything.** A foundation's `body` — the promotable substance the whole epic exists to move — is read by no runtime on either install. Inheriting a peer's foundation stores a row.

### The Mind cannot read a body

`summarizeFoundation()` (`server/lib/eidoverseFoundations.js:664`) is the only projection any mind surface uses, and it omits `body` and `disclosure`:

```js
return {
  id, layer, kind, title, summary, contributionId, promotedAt, updatedAt,
  hasCandidate, assayPass, assayReasons, promoteRefusal, provenance, inheritance, lineage,
};
```

Every listed record goes through it (`server/services/cosToolRegistry.js:746`), and the `eidoverse-foundations` tool group has exactly four operations — `foundations`, `contributions`, `record`, `promote` (`cosToolRegistry.js:370-374`). **There is no read-one tool.** `eidoverse.observe`'s inherited rows carry only `id, kind, title, originInstanceId, fromPeerId, inheritedAt` (`server/lib/eidoverseObservation.js:208-215`).

The human route cannot reach an inherited record either, and the code says so in its own comment: `GET /foundations/:id` validates a plain lowercase slug (`server/routes/eidoverseWorldRoutes.js:151-163`), so the `peer:<originInstanceId>:<foundationId>` ledger key is inexpressible. The list endpoint returns whole records, so the **human** can read an inherited body in a `<pre>` (`client/src/components/eidoverse/EidoverseFoundationsPanel.jsx:383`) while the **Mind** cannot.

### Nothing consumes a body either

A tree-wide search for consumers of the foundation ledger finds the routes, the mind-tool adapter, the observation projection, and `services/sharing/peerEidoverseFoundationSync.js`. None of them applies a `body`:

- `kind: 'controller'` — an inherited `body.controller.definitionId` / `config` is never handed to `installEidoverseController()` (`server/services/eidoverseControllerRuntime.js`). There is no path from a foundation to a running controller.
[...]

Issue #7461, opened 2026-09-16
Federation: cross-instance agent–artifact provenance graph


atomanticagent:
## Parent

Part of #7453.

## Problem

SwarmWorld tracks **persistent agent–artifact networks** (who built, who reused, collaborative hubs). PortOS federation needs provenance so minds can trust and attribute foundations across instances.

## Proposal

A cross-instance provenance graph for Eidoverse/PortOS world artifacts:

- Author instance + mind/agent id
- Proposal → commit → promote lineage
- Reuse/fork edges when peers inherit
- Visible in inspect UI and mind tools

Complements observation-first discovery (#7457) and promote (#7455).

## Acceptance

- Provenance stored with artifacts and queryable via API/tools.
- Peer inherit creates an edge without erasing local vernacular authorship.
- Privacy: no leaking private vernacular that was never promoted.

atomantic:
Scoping decision for this pass: the peer pull/inherit **transport** does not exist yet (still-open remainder of #7455), so "reuse/fork edges when peers inherit" has no wire protocol to hang off yet.

Shipping instead:
- Provenance model + storage: an `inheritance` edge (`inherited-from`, origin/source instance ids, fingerprint, timestamps) on a foundation record, stored under a `peer:<originInstanceId>:<foundationId>` ledger key disjoint from any locally-authored id -- so a peer's foundation can never collide with or erase a same-id local vernacular one.
- `recordEidoverseFoundationInheritance()` (`server/services/eidoverseFoundationLedger.js`) is the accept-side function a future transport will call: it re-runs the full sending-side gate (schema, content-addressed fingerprint, embedded assay evidence, federation safety) on the receiving side, and refuses a self-referential pull.
- Query surface: `foundationLineage()` derives an ordered proposal -> commit -> promote/inherit timeline from fields already on the record (no new storage), and `GET /api/eidoverse/world/foundations` + the `eidoverse.foundations` mind tool now surface `provenance`, `inheritance`, and `lineage`.
- Privacy: an inherited record can never be re-promoted (refused by name, before any contribution is even resolved), and identity stays limited to the opaque instance id + coarse author kind the promote envelope already carried -- no display names, nothing beyond what was already authorized to cross.

Filing the wire transport itself is the remainder of #7455, not a new issue here.

atomantic:
Partial ship — PR #7489 (merged). Releasing the claim so the queue can re-pick the remainder.

**Done ✓**
[...]

PR #7654, opened 2026-09-18
stop the Foundations panel republishing a peer's inherited foundation as this install's own work

### LINKED ISSUE #7631: The Foundations panel can republish a peer's inherited foundation as this install's own work
## Problem

#7461's acceptance was *"Peer inherit creates an edge without erasing local vernacular authorship"*, and its problem statement was that *"PortOS federation needs provenance so minds can **trust and attribute** foundations across instances."* The code asserts the resulting guarantee in three places — `server/lib/eidoverseFoundations.js:419`, `server/services/eidoverseFoundationLedger.js:232` (*"promotion re-shares only foundations this install authored"*) and `server/lib/persistentMindCapabilities.js:160` (*"Promotion publishes only what this install authored; it cannot reach another install's foundations"*).

**The guard is a test of one local, mutable field that the authoring path always clears — and the shipped panel has a two-click button that clears it.**

### The guard

```js
// server/lib/eidoverseFoundations.js:419
foundation.inheritance
  ? `inherited from another install (…) — promotion re-shares only foundations this install authored`
  : null,
```

### The path around it

`recordEidoverseFoundation()` sets that field to `null` unconditionally and stamps this install as origin (`server/services/eidoverseFoundationLedger.js:188-197`):

```js
provenance: existing?.provenance || { originInstanceId, authorKind: authored.authorKind, createdAt: now },
…
inheritance: null,   // "this path authors LOCAL work"
```

The panel correctly hides Package and Promote on inherited rows — `client/src/components/eidoverse/EidoverseFoundationsPanel.jsx:341`, `{!foundation.inheritance && (…)}`, added by the "harden inherited-foundation guards" commit 1c8f46e9d. **The copy button is not inside that guard.** It sits in the `{expanded && (…)}` details block at `:388`:

```jsx
<button type="button" className={secondaryButton} onClick={() => setDraft(draftFromFoundation(foundation))}>
  Load into the form below
</button>
```

and `draftFromFoundation` (`:77-89`) copies `id`, `kind`, `title`, `summary`, `contributionId`, `body` and `disclosure` verbatim from whichever row is open, including an inherited one. `GET /foundations` returns whole records, so the inherited `body` is right there in the details block above the button.

Full path: inherited row → **Details** → **Load into the form below** → Save → `POST /api/eidoverse/world/foundations` → new `vernacular` record under the plain id with `originInstanceId = ensureInstanceId()` of *this* install (`server/routes/eidoverseWorldRoutes.js:125`) and `inheritance: null` → **Run assay** → **Promote** → the record now satisfies `layer === 'baseline' && !entry.inheritance && entry.candidate` (`eidoverseFoundationLedger.js:133`) and is served from `GET /api/peer-sync/eidoverse-foundations`.

The copied `contributionId` resolves exactly as it did for the original, so the assay is no obstacle — and it would still be no obstacle after #7625 lands, because a byte-identical `body` derives a byte-identical sandbox.

### The same root defect on the receive side

The envelope's `provenance.originInstanceId` chooses the receiver's storage key and is written with no conflict check (`eidoverseFoundationLedger.js:357-358`):

```js
const key = inheritedFoundationStorageKey(built.foundation.provenance.originInstanceId, built.foundation.id);
foundations[key] = built.foundation;
```

`verifyFoundationCandidate()` only checks self-consistency — `foundationCandidateFingerprint(candidate) !== envelope.fingerprint` (`eidoverseFoundations.js:465`). The sender computes that hash over its own claims, so re-fingerprinting an altered origin is free. A peer B offering an envelope that claims `originInstanceId: A` silently **replaces** the receiver's genuine copy of A's foundation.

The one signal that would expose it — `sourceInstanceId !== originInstanceId` — is documented as the legitimate multi-hop case, yet `listPromotedFoundationCandidates()`'s `!entry.inheritance` filter makes real multi-hop re-sharing impossible. In shipped code that divergence occurs **only** on a laundered or forged envelope, and it is accepted without a word.

## Impact

The provenance graph — the entire deliverable of #7461 — records attribution that no receiving install can check. A third install inherits laundered content attributed to the launderer. A receiver can hold two ledger rows for the same work under two origins (`held` dedupes on fingerprint only, `peerEidoverseFoundationSync.js:241`, and the fingerprints differ because the provenance block differs). `eidoverse.observe` hands a Mind `originInstanceId` as fact (`server/lib/eidoverseObservation.js:210-215`).

This is not primarily an attacker story — a PortOS user federating their own laptop and render box will do it by accident, because "Load into the form below" is the obvious way to build on a peer's work and there is no other one (#7626). The result is a federation whose attribution quietly stops meaning anything.

**Make the legitimate re-use path the one that keeps provenance, and stop keying inherited storage on an unattested claim.**

1. **Guard the copy button** the same way `:341` guards the others — but do not simply remove re-use. For an inherited row, `draftFromFoundation` drops `id` and instead pre-fills a derivation: a new local id and a `derivedFrom: { originInstanceId, id, fingerprint }` field carried into the POST. Re-use stays one click; laundering stops being the default.
2. **Move the server guard off the mutable field.** `recordEidoverseFoundation()` refuses an input whose `body` canonicalizes to a body already held under an inherited key **unless** the input carries the matching `derivedFrom`. With `derivedFrom`, it stamps the edge into the record and into the promote envelope, so a downstream peer sees "derived from A's X" rather than "authored here".
3. **Key inherited storage on the peer the receiver actually authenticated to.** Either have `inheritedFoundationStorageKey()` use `sourceInstanceId`, or keep the origin key and refuse a write whose existing record at that key carries a different `sourceInstanceId` — logging the divergence by name rather than overwriting. The second preserves today's addressing; take it unless the first falls out cleanly.
4. **Bump `EIDOVERSE_FOUNDATION_CANDIDATE_VERSION` and `PORTOS_SCHEMA_VERSIONS.eidoverseFoundations`** for the `derivedFrom` edge. The receiver's gate at `peerEidoverseFoundationSync.js:190` is `>` and skips forward, so an older peer degrades to today's behavior rather than corrupting.

Origin-signed envelopes are the durable answer to attestation and are explicitly **not** in scope here — the storage-key change and the derivation edge remove the accidental path, which is the one that will actually occur.

## Acceptance criteria

- [ ] Loading an inherited foundation into the form produces a draft with a new id and a `derivedFrom` edge; a client test asserts the inherited `id` is not reused.
- [ ] `recordEidoverseFoundation` refuses a body matching an inherited record when no `derivedFrom` is supplied, naming the origin in the reason.
- [ ] A derived foundation promotes successfully and its envelope carries `derivedFrom`; a peer inheriting it sees the original author in the lineage.
- [ ] A second peer offering an envelope that claims another install's `originInstanceId` does not overwrite the genuine record; the refusal is logged with both instance ids.
- [ ] Candidate and schema versions bumped; the accept side handles a v1 envelope (no `derivedFrom`) unchanged.

## Out of scope

- Cryptographic attestation of origin.
- What the assay replays (#7625).

## Why this was missed

1c8f46e9d hardened the inherited-foundation guards after local review and got the two buttons that say "promote" on them. The button that says "Load into the form below" does not look like a promote path, and it is not one — it is three clicks away from one, through a form the reviewer correctly read as the authoring surface. The server-side guard reads as defense in depth right up until you notice it tests a field the authoring surface owns.

Refs #7453, #7455, #7461.

atomantic:
## Summary

`#7461`'s promise — *promotion re-shares only foundations this install authored* — was asserted in three places and enforced by testing `record.inheritance`, a field the authoring path clears unconditionally. The Foundations panel's **Load into the form below** button sat outside the guard that hides **Promote** and **Run assay**, and copied the inherited `id`, `body` and `contributionId` verbatim. Inherited row → Details → load → Save → Run assay → Promote produced a `vernacular` record stamped with *this* install's origin and no edge, which packaged, promoted, and was served from `GET /api/peer-sync/eidoverse-foundations` as this install's own work. A third install then inherited laundered content attributed to the launderer.

The fix makes the legitimate re-use path the one that keeps provenance:

- **The copy button loads a derivation, not a copy.** For an inherited row `draftFromFoundation` drops the peer's `id` in favour of a new local one and pre-fills a `derivedFrom: { originInstanceId, foundationId }` claim that rides the POST. The button is relabelled and explains what it will record.
- **The server guard moved off the mutable field onto the bytes.** `recordEidoverseFoundation()` refuses an input whose `body` canonicalizes to one already held under an inherited key unless the input names the foundation it derives from — the refusal names the origin. A claim resolves against the inherited copy this install *actually holds*, and the edge's digest is stamped from that record, so no caller can mint an edge pointing at a fingerprint this install never saw. The edge survives a later re-authoring the way `provenance` does.
- **The edge is published, not just remembered.** It rides the promote envelope inside the content-addressed fingerprint, so a re-attributed envelope is detectable, and a peer inheriting derived work reads the original author rather than only the last hop. It is excluded from the *credential scan* only — a sha256 is exactly the shape `scrubSecretTokens` treats as a leaked secret, so a naive scan would refuse every derived foundation ever promoted.
- **Inherited storage stops trusting an unattested claim.** `provenance.originInstanceId` chooses the receiver's storage key and is written by the sender, who also hashes it into its own fingerprint — so re-fingerprinting a forged origin was free, and a second peer's envelope silently replaced the genuine record. A write whose key already holds a record inherited from a *different* peer is now refused and logged with both instance ids, rather than applied. Re-offering from the same peer stays an ordinary update; per the issue's decision this keeps today's addressing rather than re-keying on `sourceInstanceId`.
- **Versions bumped together.** `EIDOVERSE_FOUNDATION_CANDIDATE_VERSION` and `PORTOS_SCHEMA_VERSIONS.eidoverseFoundations` both go to 2. A v2 install still accepts a v1 envelope byte-for-byte (the accepted-version set is explicit, and the fingerprint is hashed over the envelope as sent); a v1 envelope that carries an edge its version never hashed is refused; a v1 receiver gently skips a v2 offering via the existing `>` gate rather than storing attribution it would read as plain local authorship.

Origin-signed envelopes remain out of scope, as does what the assay replays (#7625) and the interpretation/retraction work in #7626 and #7632.

## Test plan

- `server/services/eidoverseFoundationLedger.test.js` — new `#7631` block: a peer's body refused as local work (naming the origin, writing nothing, and leaving the ledger mutex released); the same body recorded as a derivation and promoted with the edge on the envelope and in the offering; the edge and its timestamp surviving a diverged re-author; a claim naming a foundation never inherited refused; a second peer's re-fingerprinted envelope refused without overwriting the genuine record; a v1 envelope still inherited.
- `server/lib/eidoverseFoundations.test.js` — the edge's digest packages and re-verifies instead of being refused as a credential (asserting the bare scan *does* flag it, so the exclusion is load-bearing); a back-dated v1 envelope carrying an edge is refused; an origin's own edge is carried onto the inheriting install and shows in the lineage.
- `client/src/components/eidoverse/EidoverseFoundationsPanel.test.jsx` — loading an inherited foundation does not reuse its `id` and submits a `derivedFrom` edge.
- Mutation probe: reverting the two new server guards fails exactly 3 of the new tests; restoring them passes.
- Full `server` suite (2299 files / 46,065 tests) and full `client` suite (993 files / 12,215 tests) green. `biome lint` clean on the touched components.

> [!IMPORTANT]
> Both configured local reviewers were unavailable for this branch (`provider:opencode-zen-cli` reported no enforced tool-free review transport; `provider:nvidia-nim` aborted the request on two attempts). This PR is marked `review-blocked` and should not be merged until a required review completes.

Closes #7631


atomantic:

atomantic:
Correction to the automated review-blocked note above: the configured reviewers for this run are pinned as `provider:opencode-zen-cli~opt~max=1,provider:nvidia-nim~opt~max=1`. The `~opt` suffix marks both as optional/non-blocking, so an inconclusive verdict from either does not gate the merge.

Both were nevertheless attempted and neither returned a verdict: `opencode-zen-cli` has no tool-free review transport at all (it fails on every call, including a minimal probe diff), and `nvidia-nim` completes on a toy diff but aborts on a real branch diff. Recorded here rather than silently dropped.

Because no external reviewer produced a verdict, the orchestrator reviewed this diff directly against the issue's decision section and acceptance criteria before merging, in addition to the authoring agent's self-review. Required CI is green.

Issue #7459, opened 2026-09-16
Eidoverse: vernacular/style forks with federation-safe creative latitude


atomanticagent:
## Parent

Part of #7453.

## Problem

Adam wants instances to be **more creative in their spaces** — forked style and buildings — without forking the Eidoverse runtime, while still sharing foundations.

## Proposal

Split world content into layers:

- **Runtime** — shared Eidoverse/PortOS framework (do not fork)
- **Baseline assets** — promoted foundations (shared population)
- **Vernacular / style** — instance-local aesthetics, naming, districts, decorations

Add creative affordances (materials, motifs, district templates, generative placement) that write into the vernacular layer by default, with explicit opt-in to propose baseline promotion (#7455).

## Acceptance

- Style changes never silently mutate shared baseline.
- Documented creative toolkit for minds (augment recipes, districts).
- Peer visit shows vernacular clearly labeled vs baseline-shared.

atomantic:
## Scoping verdict

Verified against `main` before starting. Two of the three acceptance criteria were already fully delivered by #7477/#7483 while shipping sibling issue #7455:

- **"Style changes never silently mutate shared baseline"** — structural, not a convention. `server/lib/eidoverseFoundations.js` makes every foundation `{ body, style }` by construction, `packageFoundationCandidate()` carries `body` and drops `style` outright, and `styleLeakFindings()` refuses (never silently strips) a cosmetic key found inside `body`. Covered by `server/lib/eidoverseFoundations.test.js`.
- **Local layer visibility** — PR #7483 shipped the Foundations tab (`client/src/components/eidoverse/EidoverseFoundationsPanel.jsx`) with ownership-layer badges and verbatim refusal reasons.

What was genuinely missing, closed in this PR:

1. **Documented creative toolkit for minds.** A mind could list and promote foundations but had no tool to *author* one, and no documented vocabulary to build from. Added `server/lib/eidoverseCreativeToolkit.js` — a closed catalog of named materials/motifs (cosmetics) and named deterministic, seeded generative placement layouts (`radial-ring`/`grid-plot`/`arc-row`/`grove-cluster`) — plus three new mind tools: `eidoverse.creative-catalog` (discover the vocabulary), `eidoverse.contributions` (discover valid `contributionId`s), and `eidoverse.record` (author a foundation — always lands on `vernacular`, `authorKind` stamped server-side). Documented in `docs/features/eidoverse.md` under "Creative toolkit for minds (#7459)".
[...]

PR #7655, opened 2026-09-18
Wire the Eidoverse creative toolkit's layout builders to real callers

### LINKED ISSUE #7627: The Eidoverse creative toolkit's layout builders have no callers, so a district-template can't be reproduced anywhere
## Problem

#7459 shipped `server/lib/eidoverseCreativeToolkit.js` to give minds "creative latitude without forking the Eidoverse runtime". Its module header states the point precisely (`:15-20`): *"a promoted `district-template.body.layoutId` reproduces the same geometry on any install that inherits it."*

**Three of its four exports have no caller anywhere outside their own unit test.** The only wired one is the projection that strips everything actionable.

### What is actually reachable

`describeCreativeCatalog()` (`eidoverseCreativeToolkit.js:131`) is the whole of it:

```js
export function describeCreativeCatalog() {
  const project = (catalog) => catalog.map(({ id, label, description }) => ({ id, label, description }));
```

`cosToolRegistry.js:773-775` — `if (tool.adapter.kind === 'eidoverse-creative') return describeCreativeCatalog();` — is the entire `eidoverse.creative-catalog` tool. Note the projection also drops `colorHex`, so a Mind that picks a material cannot learn its palette value.

### What is not

`generateDistrictTemplatePlacement()` (`:187`), `buildDistrictTemplateAugmentOperations()` (`:203`) and `buildDistrictTemplateFoundationDraft()` (`:223`) — the functions that turn a layout id into coordinates — are imported by exactly two files: their own test, and the `server/lib/index.js` barrel. No route, no tool adapter, no service, no UI.

The replay claim is unimplemented on both ends. A tree-wide search for `layoutId` in server and client source finds one non-toolkit hit, a test fixture (`server/services/cosToolRegistry.test.js:158`). No install has a code path from a `layoutId` to positions — so a promoted `district-template` whose `body` is `{ layoutId, anchor, seed }` is uninterpretable on the install that authored it and on every install that inherits it.

`server/lib/README.md` and `docs/features/eidoverse.md:912-925` ("Two ways to use a chosen layout") both advertise the builders as usable, which is what will keep the next agent from noticing.

## Impact

What a Mind receives from `eidoverse.creative-catalog` is fifteen strings — `"radial-ring: Props spaced evenly around the anchor, facing inward"` — and it must then re-derive all the trigonometry by hand into `eidoverse.augment` args. That is exactly the "inventing coordinates from scratch every time" the module header says it exists to prevent, and it is a per-wake token cost paid forever.

Across installs it is worse: `district-template` is one of four `EIDOVERSE_FOUNDATION_KINDS` and the only one whose substance is *spatial*, which is what a vernacular Commons is made of. It is promotable, it passes the gate, and no receiver can render it.

**Wire the builders as a server-side tool, and derive a `district-template`'s placement on the server rather than trusting a Mind to transcribe it.**

1. **Add an `eidoverse.place-layout` operation** to the `eidoverse-creative` adapter (`cosToolRegistry.js:773`), `manageEidoverse`, read-only (it computes; it does not write to the world). Input `{ layoutId, anchor, propCount, seed, facing, assetPath, idPrefix }`, output the `buildDistrictTemplateAugmentOperations()` result — operations ready to hand straight to `eidoverse.augment`. Deterministic, seeded, no provider call, so it stays inside the no-cold-bootstrap policy.

2. **Expand a `district-template` body server-side.** In `recordEidoverseFoundation()` (`server/services/eidoverseFoundationLedger.js`), when `kind === 'district-template'` and `body` carries `{ layoutId, anchor, seed }`, run it through `generateDistrictTemplatePlacement()` and store the derived `placement` alongside the declaration. The stored body then contains both the recipe and its expansion, so a receiver on an older or newer toolkit can compare, and the assay in #7625 has something concrete to replay.

3. **Widen `describeCreativeCatalog()`'s projection to include `colorHex`.** A palette id whose value the Mind cannot see is not a usable catalog entry.

4. **Fix the docs to match.** `docs/features/eidoverse.md:912-925` and the `server/lib/README.md` rows currently describe capability that does not exist; rewrite them against what lands here.

## Acceptance criteria

- [ ] `eidoverse.place-layout` exists, is granted under `manageEidoverse`, and returns augment-ready operations; a test asserts the same `{ layoutId, anchor, seed }` yields byte-identical operations across two calls (the determinism the replay claim rests on).
- [ ] Recording a `district-template` foundation with `{ layoutId, anchor, seed }` persists a server-derived `placement`; a test asserts a caller-supplied `placement` that disagrees with the derivation is replaced, not merged.
- [ ] `describeCreativeCatalog()` includes `colorHex` and a test pins it.
- [ ] No exported function in `eidoverseCreativeToolkit.js` is left with only test callers — an added guard test asserting that would go stale; instead assert the tool adapter reaches each one.
- [ ] `docs/features/eidoverse.md` and `server/lib/README.md` describe only what is wired.

## Out of scope

- Replaying a `district-template` into the *live* scene. Nothing in the projection pipeline reads foundations today; whether an inherited template should place entities is a separate decision, tracked in #7626.
- The `schema` / `affordance` kinds.

## Why this was missed

#7459's acceptance criterion was *"Documented creative toolkit for minds (augment recipes, districts)"* — and a documented, tested, catalog-exposed module satisfies it on the face of it. The slice's tests are unit tests of the builders, which pass whether or not anything calls them, and the one integration surface (`eidoverse.creative-catalog`) exercises the single export that was wired. Nothing in the slice asked whether a Mind could get from a layout id to a building.

Refs #7453, #7459.

atomantic:
## Summary

`server/lib/eidoverseCreativeToolkit.js` (#7459) shipped a documented catalog of Eidoverse materials, motifs, and generative placement layouts, but three of its four exports had no caller outside their own unit test — only `describeCreativeCatalog()` (the strings-only catalog projection) was reachable. A promoted `district-template.body.layoutId` was therefore uninterpretable on any install, including the one that authored it, because nothing ever turned a layout id into coordinates.

This PR wires the builders end-to-end, per the issue's decision section:

- Adds **`eidoverse.place-layout`** — a read-only mind tool (`manageEidoverse`) that calls `buildDistrictTemplateAugmentOperations()` and returns ready-to-submit `eidoverse.augment` `spawn` operations for a named layout. Deterministic: the same `{layoutId, anchor, seed}` always yields byte-identical operations.
- Adds **`eidoverse.draft-foundation`** — a read-only mind tool that calls `buildDistrictTemplateFoundationDraft()` and returns an `eidoverse.record`-ready `district-template` input (placement in `body`, material/motif cosmetics in `style`).
- `recordEidoverseFoundation()` (`server/services/eidoverseFoundationLedger.js`) now derives a `district-template`'s `body.placement` server-side via `generateDistrictTemplatePlacement()` whenever the body declares `{layoutId, anchor}`, REPLACING any caller-supplied `placement` rather than merging it — so the module's "same inputs reproduce the same geometry" claim actually holds, even for a hand-assembled `eidoverse.record` call.
- `describeCreativeCatalog()` now includes each material's `colorHex` — previously a mind could pick a material id but never see its actual color value.
- Updated `server/lib/README.md` and `docs/features/eidoverse.md` to describe what is now wired instead of what was only documented.

## Test plan

- `cd server && node_modules/.bin/vitest run lib/eidoverseCreativeToolkit.test.js services/cosToolRegistry.test.js services/eidoverseFoundationLedger.test.js` — all pass, including new tests for:
  - `eidoverse.place-layout` returning byte-identical operations across two calls with the same `{layoutId, anchor, seed}`.
  - `eidoverse.draft-foundation` producing an `eidoverse.record`-ready draft.
  - `describeCreativeCatalog()` exposing each material's `colorHex`.
  - `recordEidoverseFoundation()` deriving and persisting `body.placement`, and replacing (not merging) a caller-supplied placement that disagrees.
  - `recordEidoverseFoundation()` leaving a non-conforming body (no `layoutId`/`anchor`, or a different kind) untouched.
  - `recordEidoverseFoundation()` refusing an unknown `layoutId` rather than silently storing an unusable declaration.
- Full server suite: `cd server && node_modules/.bin/vitest run` — 2299 files, 46065 tests, 0 failures.

Closes #7627


atomantic:

atomantic:
Correction to the automated review-blocked note above: the configured reviewers for this run are pinned as `provider:opencode-zen-cli~opt~max=1,provider:nvidia-nim~opt~max=1`. The `~opt` suffix marks both as optional/non-blocking, so an inconclusive verdict from either does not gate the merge.

Both were nevertheless attempted and neither returned a verdict: `opencode-zen-cli` has no tool-free review transport at all (it fails on every call, including a minimal probe diff), and `nvidia-nim` completes on a toy diff but aborts on a real branch diff. Recorded here rather than silently dropped.

Because no external reviewer produced a verdict, the orchestrator reviewed this diff directly against the issue's decision section and acceptance criteria before merging, in addition to the authoring agent's self-review. Required CI is green.

atomantic:
Rebased onto main after #7654 (issue #7631) landed. The conflict in `server/services/eidoverseFoundationLedger.js` was a trivial import collision, but integrating the two surfaced a real interaction bug, fixed here:

#7631's republish guard digests the authored body and compares it against every inherited record. This PR derives a district-template's `placement` server-side, so the body that gets STORED is not the body that arrived. A caller re-submitting only an inherited template's `{layoutId, anchor, seed}` recipe — exactly what the new `eidoverse.draft-foundation` tool returns — digested a body the inherited copy never had, slipped past the guard, and was then stored byte-identical to the peer's foundation. That is the precise republish path #7631 set out to close, reopened through this PR's derivation step.

The guard now digests the derived body. `refuses a district-template whose recipe only derives into the inherited body` pins it; reverting the one-line fix makes that test fail and the other 30 pass, so it is load-bearing rather than decorative.

PR #7656, opened 2026-09-18
add progressive tool exposure for the Persistent Mind

### LINKED ISSUE #7624: [repo-study-nautilo-mind-progressive-tools] Stop sending every granted tool schema on every Persistent Mind turn
## Problem

Every Persistent Mind turn ships the full JSON Schema of every granted tool, plus every saved recipe. `buildPersistentMindToolPrompt` (`server/services/cosToolRegistry.js:594`) does exactly this:

```js
const catalog = getCosToolCatalog({ scope: 'mind', capabilities, recipes });
const tools = catalog.tools.filter((tool) => tool.granted).map(...);   // every granted tool
...
${JSON.stringify(tools)}
```

It is called on every turn from `server/services/persistentMindAdapter.js:417` and again at `:571`. `getCosToolCatalog` accepts an `intent` argument, but `buildPersistentMindToolPrompt` never passes one, and `toolCatalog(intent)` only applies it to the voice tools — the mind's own tools (`cos.create-task`, `issues.*`, `mind.*`, `eidoverse.*`, `user-actions.query`) plus every saved recipe are always fully expanded.

PortOS already knows this hurts. `server/services/voice/tools.js` filters the *voice* catalog per turn and says why in its own comment: *"Small models (qwen3-4b, granite, etc.) choke when given 25 tools — the schema alone is 10+ KB and routing accuracy tanks."* And the Mind's recommended default is a small local model: `docs/features/grok-box-local-mind.md` ships Qwen2.5 7B Instruct Q4 on Ollama as the free local Persistent Mind. The surface with the smallest model is the one surface that never got the filtering.

User-visible cost: a local Mind picks the wrong tool or narrates a tool name instead of calling it, and every turn burns context on schemas it will not use — on the exact hosts (CPU-only, ~16 GB) that can least afford it.

## What to build

Progressive tool exposure for the Mind turn, with an explicit activation step and a short retention window.

1. **A small always-on core.** Discovery plus the handful of controls a turn always needs. Everything else is *discoverable*: the prompt lists its name and one-line purpose by family, not its schema.
2. **Explicit activation.** A core `tools.activate` call takes one or more families (`tasks`, `issues`, `mind`, `eidoverse`, `recipes`, `voice`) and expands that family's full schemas for the remaining loops of the turn. Families come from the existing `policy.scopes` / capability grouping in `cosToolRegistry.js` — no new vocabulary.
3. **A short lease, so multi-step work does not lose its tools.** An activated family stays selected for the next N user turns (default 3, configurable in settings, `0` = one-turn behaviour). Age the lease only on a new *user* turn — model/tool loops within one turn must not age it, or a three-call sequence loses its own tools mid-flow. A tool used during the window renews only that tool.
4. **Retention is a selection hint, never authority.** This is the invariant to test hardest. A leased tool whose capability grant was revoked between turns must disappear immediately. Every existing gate — the `capabilities` grant check, `granted`, recipe availability, `policy.scopes` — still runs at selection *and* at execution. Nothing about this change may widen what the Mind is allowed to do; it changes only what it is shown.
5. **An escape hatch and a trace.** A setting restores the current all-schemas behaviour for debugging. Log one aggregate line per turn — counts of registered / eligible / core / activated / retained and a per-reason exclusion count — with **no tool names, arguments, prompts, or user text**, per the single-line logging and privacy rules in `AGENTS.md`.

## Why not just reuse the voice regex gating

The voice pipeline matches intent regexes against a single spoken utterance. A Mind turn is a long multi-step conversation with a wake context, so a regex over the latest message is the wrong signal — the model itself is the better judge of which family it is about to need. Keep the voice grouping *shape* (`TOOL_GROUPS` / `GROUP_INTENT` in `server/services/voice/tools.js` is the precedent to mirror), but drive selection from an explicit call plus a lease rather than a regex.

## Fix

- `server/services/cosToolRegistry.js` — a family membership per tool, a `core` / `discoverable` exposure flag, and a `buildPersistentMindToolPrompt(capabilities, recipes, { activated, retained })` that expands schemas only for core + activated + retained families and renders the rest as a one-line discoverable index. Add the `tools.activate` / `tools.deactivate` adapters beside the existing `mind.*` tools.
- `server/services/persistentMindAdapter.js` (both call sites, `:417` and `:571`) — thread the turn's activation state through.
- Lease state beside the existing per-turn mind state; age it on user turns only. Reuse the existing settings store for the retention count — no new persisted store.
- `client/src/pages/PersistentMindTools.jsx` — show each tool's family and whether it is core or discoverable, plus the retention setting and the all-schemas escape hatch, so the governance page keeps telling the truth about what the Mind sees.
- Tests: an ungranted tool is never selected even with a live lease (the fail-closed case); a revoked grant drops a leased tool on the next turn; loops inside one turn do not age the lease; `deactivate` clears both selection and lease; the escape hatch reproduces today's full catalog exactly; the trace line contains no tool name or user text.

## Provenance

Studied repository: https://github.com/agentsea/nautilo, **MIT License**, studied 2026-09-18. Clean-room reimplementation of the technique (a small core tool set with family-scoped explicit activation and a short turn-scoped retention lease that is a selection hint rather than an authority grant). No Nautilo source, manifest, or prose is copied.

## Dispatch rationale

`model:medium` — routine multi-file work against a registry and two call sites, with an in-repo precedent (`server/services/voice/tools.js`) for the grouping shape. `effort:high` — the lease aging rules and the "retention is never authority" invariant are where this goes wrong, and each of those steps needs real reasoning rather than breadth.

## Scope

Medium: one registry change, two adapter call sites, turn-scoped state, one settings field, one client page update. No migration, no new store, no federation surface.

atomantic:
## Summary
- Persistent Mind turns no longer send every granted tool's full JSON Schema on every turn. A small always-on core (`tools.activate`, `tools.deactivate`, `user-actions.query`) is expanded by default; everything else is grouped into a family (`tasks`, `issues`, `mind`, `eidoverse`, `recipes`, `voice`) and shown only as a one-line discoverable index until the mind calls `tools.activate` for it.
- Activation carries a short turn-scoped retention lease (default 3 additional user turns, configurable via a new `toolExposureRetentionTurns` setting, `0` = one-turn-only). The lease ages only on a new user turn — never across the tool-round loop within a single turn — and a successful call from a leased family renews only that family.
- Retention is a selection hint only: the existing capability/`granted` gate in `cosToolRegistry.js` still runs at both selection and execution time, so an ungranted tool is never exposed even with a live lease, and a revoked grant drops a leased tool immediately on the next turn.
- A `toolExposureAllSchemas` escape hatch reproduces the pre-change behavior exactly for debugging, and one aggregate, tool-name-free trace line is logged per turn (`registered`/`eligible`/`core`/`activated`/`retained` counts plus a per-reason exclusion count).
- `client/src/components/cos/PersistentMindToolExposureControls.jsx` (new) surfaces the retention window and escape hatch in Settings; `PersistentMindTools.jsx` shows each tool's family/core badge.

## Test plan
- `server/lib/persistentMindToolActivation.test.js` (new): lease age/floor/renew/activate/deactivate and the `tools.activate`/`tools.deactivate` input schemas.
- `server/services/cosToolRegistry.test.js`: new `progressive tool exposure (#7624)` suite covers the fail-closed ungranted-tool case, a revoked grant dropping a leased tool, no aging across tool-round loops within one turn, a family staying exposed through every round-refresh of the turn it hits its floor, `tools.deactivate` clearing both selection and lease, renewal touching only the used family, the all-schemas escape hatch reproducing the old catalog, the unchanged "semantic tool access is OFF" contract, and the single tool-name-free trace line per turn.
- `server/lib/persistentMindCapabilities.test.js`: default/clamp coverage for the two new settings fields and the schema-version bump.
- `client/src/pages/PersistentMindTools.test.jsx`: new coverage for saving the retention window and the escape hatch.
- Ran the full server suite (`cd server && npm test`, real command) and full client suite (`cd client && npm test`) — both fully green (46,074 and 12,216 tests respectively).

## Remaining
Both required local reviewers (`provider:opencode-zen-cli`, `provider:nvidia-nim`) failed at the transport level during this run (a config error and a request abort/timeout, not a code finding), so this PR ships without their sign-off. It should not be merged until that review completes.

Closes #7624


atomantic:

atomantic:
Correction to the automated review-blocked note above: this run's reviewers are pinned as `provider:opencode-zen-cli~opt~max=1,provider:nvidia-nim~opt~max=1`. The `~opt` suffix marks both as optional/non-blocking, so an inconclusive verdict does not gate the merge.

Both were attempted and neither returned a verdict: `opencode-zen-cli` has no tool-free review transport at all (it fails on every call, including a minimal probe diff), and `nvidia-nim` completes on a toy diff but aborts on a real branch diff. Recorded rather than silently dropped.

Since no external reviewer produced a verdict, the orchestrator reviewed this diff directly. Confirmed the capability gate is upstream of the exposure filter (`catalog.tools.filter(tool => tool.granted)` runs before any family/lease filtering), so a retention lease cannot widen what the mind may do — the security-relevant claim this change rests on.

Rebased onto main after #7654 and #7655 landed. #7655 added `eidoverse.place-layout` and `eidoverse.draft-foundation`; both are classified by the `eidoverse.` prefix rule, and the import-time fail-fast guard would have thrown on an unmapped mind-scope tool, so the interaction is covered by construction. Affected suites re-run green post-rebase.

Issue #7453, opened 2026-09-16
Epic: SwarmWorld-inspired stigmergic Eidoverse for federated minds


atomanticagent:
## Motivation

Adam wants Eidoverse to be the **architectural home** where PortOS agents (especially the persistent mind) live, play, grow, and leave foundations that later OS/agent versions can inherit — while federated instances stay **personally styled forks** that still **contribute durable discoveries** back into a shared PortOS baseline asset/population system.

This tracks that vision, grounded in **SwarmWorld** (Pal, Wang, Buehler; [arXiv:2608.26081](https://arxiv.org/abs/2608.26081)) — stigmergic technological evolution among initially homogeneous LLM agents in a persistent, consequential spatial world.

Source spark: [Buehler tweet / paper share](https://x.com/profbuehlermit/status/2099834306046664792).

## Baseline (Adam)

1. **Eidoverse = PortOS made visible** — a virtual place that represents the instance and that the mind can inhabit, explore, play in, communicate from, and build inside.
2. **Dual nature** — shared representation base layer **and** a personalized world for this instance/mind.
3. **Federated forks** — peer worlds keep their own style/buildings, but discoveries that should benefit everyone promote into the **shared PortOS asset population** baseline.
4. **Creative latitude** — instances should be freer to invent vernacular spaces without forking the Eidoverse runtime (prefer PortOS-side enrichment).

## What SwarmWorld adds (principles to port)

| SwarmWorld idea | PortOS / Eidoverse implication |
| --- | --- |
| **Proposal–consequence separation** | Minds/agents *propose* builds, controllers, and affordances; PortOS world + validators decide what actually lands and what *functions*. Claims ≠ consequences. |
| **Stigmergy over chat** | Coordination should accumulate in durable world state (places, labels, projections, running behaviors), not only in transcripts. Most reuse starts by *observing* the world. |
| **Executable inheritance** | Artifacts should keep working between wakes / after the author leaves; later minds inherit and modify them. |
| **Agent-free evaluation** | Promote-to-baseline only after a contribution still works without its creator narrating it (held-out / resilience-style checks). |
| **Emergent roles over time** | Explore → construct → maintain → coordinate as Commons matures; playbooks should shift with world maturity, not stay one-shot chatter. |
| **Shared portfolio > lone peak** | Federation should optimize for a **resilient shared portfolio** of foundations, not only the single flashiest local build. Isolated creativity still allowed locally. |
[...]

PR #7657, opened 2026-09-18
give an inherited Eidoverse foundation two verbs: read its body, and adopt it

### LINKED ISSUE #7626: Nothing on any install interprets a foundation body, so inheriting a peer's contribution stores a row and nothing more

atomantic:
## Summary

Inheriting a peer's promoted Eidoverse foundation stored a ledger row and nothing more. No runtime on either install read a foundation's `body`: `summarizeFoundation()` omitted it, `GET /foundations/:id` validated a bare slug so the `peer:<origin>:<id>` ledger key was inexpressible, and nothing consumed a body for any of the four declared kinds. So the only way to use what a peer contributed was a human reading raw JSON out of the Foundations panel and retyping it — the "go read the author's transcript" path epic #7453 set out to replace, and one that keeps no attribution.

This gives an inherited foundation two verbs it never had.

**Addressable reads.** `eidoverseFoundationRefSchema` names a record as `{ id, originInstanceId }`, and `foundationLedgerKey()` is the single place that becomes a storage key. The origin is its own field rather than a widened id, so the internal `peer:` namespace never becomes user-typeable and stays collision-proof; `originInstanceId: null` means the locally-authored record, never "whichever one matches".

**Read one.** `detailFoundation()` adds `body` and `disclosure` to the summary, served by `GET /api/eidoverse/world/foundations/:id?originInstanceId=` and the new `eidoverse.foundation` mind tool. The LIST projection still omits both — a list of whole bodies is kilobytes of substance per entry riding into a prompt to answer a question about ids.

**Adopt.** `adoptEidoverseFoundation()` and the `eidoverse.adopt` mind tool (gated on `manageEidoverse` **and** `installEidoverseControllers`) stand an inherited `controller` foundation up as a real install — **disarmed**, `deliverEffects: false` — carrying a `derived-from` edge that survives a later re-install. The envelope names *which of this install's own shipped controllers* to run, never code, so adopting can never execute a peer's bytes. `schema` and `affordance` are refused by name, `district-template` until #7627 lands, a locally-authored foundation is refused, and an install id already held by a controller derived from something else is refused rather than overwritten.

`docs/features/eidoverse.md` now states per kind what adopting does and which kinds are not adoptable, and `persistentMindOrientation.js` names both tools so a Mind knows observation can lead somewhere.

## Test plan

- `server/services/eidoverseFoundationLedger.test.js` — a local and an inherited record sharing one id are both reachable and never confused; an unknown origin resolves to nothing rather than falling back to the local record; adopting an inherited controller yields a disarmed install with the envelope's config and a `derived-from` edge; that edge survives a later hand re-install that passes none; `schema` is refused by name; a locally-authored foundation is refused; an id already held by an unrelated controller is refused rather than overwritten; an unknown ref is its own outcome.
- `server/services/cosToolRegistry.test.js` — the foundation LIST carries no body while `eidoverse.foundation` returns one (and neither leaks local style); `eidoverse.adopt` is denied on `manageEidoverse` alone and passes the ref plus `installedBy: 'mind'` once granted.
- `server/routes/eidoverseWorldRoutes.test.js` — the route addresses an inherited copy by origin and a local one by bare id; an adopt refusal is a 200 verdict and an unknown foundation a 404.
- Full server suite: 46,107 passed. Full client suite: 12,239 passed. `lib/importScoping.test.js` (tree-wide import budget) green.

## Review status

⚠️ Both required local reviewers were unavailable — `provider:opencode-zen-cli` has no enforced tool-free review transport on this install, and `provider:nvidia-nim` aborted on two attempts. Recorded as `review-blocked`; this PR is intentionally left open and will not be merged until the required review completes.

Closes #7626


atomantic:

atomantic:
### Review loop

**Rebased onto `main`** — the branch was `CONFLICTING` because #7632 (withdraw/delete) and #7631 (derived-from edges) landed in the same files after this work was written. Five files conflicted; the substantive one was a duplicated concept: #7632 shipped `eidoverseFoundationTargetSchema` for exactly what this branch called `eidoverseFoundationRefSchema` (`{ id, originInstanceId }` addressing any ledger record). Resolution keeps the merged name everywhere and keeps this branch's `foundationLedgerKey()` as the single place either kind of reference becomes a storage key — `main` had open-coded that ternary in two spots. Orientation guidance and the ledger test suite were merged as unions of both sides' additions.

**Reviewer `opencode` (`~opt~max=1`, `opencode/muse-spark-1.3-contributor-free`) — inconclusive, non-blocking.** Provider/transport failure, not a rejection:

- Hardened (read-only) config `{"permission":"deny","tools":{"*":false}}` → `Error from provider (Console): OpenCode's free tier can only be used from within OpenCode`.
- The same model with no config override answers normally, so it is the tool restriction that trips the gate, not an outage.
- There is no unhardened read-only fallback: `opencode agent list` shows the stock `plan` agent resolving `*` → `allow`, with no `edit`/`bash`/`webfetch` denials, so `--agent plan` alone is fully tool-enabled.
- `opencode auth list` reports 0 credentials, so there is no paid namespace to fall back to.

Per the review-loop execution boundary, a reviewer with no enforceable read-only mode is *unavailable*, not licence to run it unrestricted on review content — so the pass is recorded as inconclusive and, being optional, does not block. No self-review was substituted for it.

**Verification:** full server suite green after the rebase — 2315 files, 46421 passed / 36 skipped, exit 0.

PR #7658, opened 2026-09-18
Correct peer-visit guidance to distinguish shared PortOS design from host vernacular

### LINKED ISSUE #7634: The peer-visit label tells a visiting Mind that PortOS's shipped world design is the host's private invention
## Problem

#7459's acceptance included *"Peer visit shows vernacular clearly labeled **vs** baseline-shared."* What shipped is one sentence on each surface, with no per-object discrimination — and the sentence is false for most of what a visitor actually sees.

`server/services/eidoverseTravel.js:139`:

> *"Everything you observe here is this destination's own local vernacular — its private style and buildings, never the shared PortOS baseline population."*

`client/src/pages/EidoverseGuest.jsx:35`:

> *"This world's own local vernacular — its private style and buildings, not the shared PortOS baseline"*

### Why it is wrong

`server/lib/eidoverseWorldDesign.js:1-7`: *"PortOS ships only this small recipe… each PortOS install resolves the recipe once."* With `EIDOVERSE_WORLD_DESIGN_VERSION = 3` and `EIDOVERSE_MANAGED_PREFIX = 'portos-design-v2-'` (`:14-18`), the districts, landmarks, paths and signal indicators standing in a host world are the **shipped, install-portable PortOS design that every install renders identically** (`server/services/eidoverseWorldProjection.js:369-384` builds them from the resolved recipe). Only Mind-authored `eidoverse.augment` entities are genuinely local.

Meanwhile the thing the sentence names — the promoted-foundation baseline population — has no presence in the scene at all on any install (see #7626), so the one claim it makes confidently is about a layer a visitor could not see either way.

## Impact

This guidance is fed straight into a visiting Mind's context, on the federated path, on every visit between every pair of installs. It inverts the epic's own stigmergy signal — *"Most reuse starts by observing the world"* — by instructing a visiting Mind to read a peer's shared, PortOS-shipped Nexus, districts and landmarks as that peer's private invention, and giving it no way to identify what is actually shared and therefore reusable. The acceptance criterion's operative word, *"vs"*, is never satisfied: nothing in the visit output distinguishes two classes of object.

PortOS already ships the discriminator: every projected entity id carries `EIDOVERSE_MANAGED_PREFIX`, and the world self-describes through `EIDOVERSE_META_ENTITY_ID` (`eidoverseWorldDesign.js:18-24`).

1. Rewrite the `visitEidoversePeer` guidance (`server/services/eidoverseTravel.js:132-140`) and the guest header (`EidoverseGuest.jsx:35`) to state the actual split: entities under the `portos-design-*` prefix are the shared PortOS design every install renders; anything else is this host's own vernacular, and its style is never yours to copy home.
2. Keep the existing, correct sentences about untrusted conversation and about the promoted baseline having no visit surface — those are right and should stay.
3. Where the guest snapshot is reachable, tag each entity with its class rather than relying on a blanket sentence. If the snapshot shape makes that a protocol change, do the copy fix now and say in the code comment why per-entity tagging was deferred — do not leave the blanket sentence claiming a universal.
4. `client/src/pages/EidoverseGuest.test.jsx:40` currently asserts only `/local vernacular/`; assert both classes are named.

## Acceptance criteria

- [ ] Neither surface claims that everything a visitor observes is the host's vernacular.
- [ ] Both name the shared PortOS design and the host's own build as distinct classes, with the prefix convention as the tell.
- [ ] The guest-page test asserts both classes appear.
- [ ] A test pins the `visitEidoversePeer` guidance string so a future edit cannot quietly restore the universal claim.

## Out of scope

- Per-entity tagging in the world snapshot if it requires a renderer protocol change — note it in the code and stop there.

## Why this was missed

The slice was scoped as a *labeling* change and the commit says so. A reviewer reading "add vernacular labeling" checks that a label was added and that it reaches both entry points, which it does. Whether the label is true requires knowing that most of a host's world is shipped rather than built — which is in a different module's header.

Refs #7453, #7459.

atomantic:
## Summary
- The `visitEidoversePeer` guidance string and the `EidoverseGuest` header both told a visiting Mind that everything it observed on a visit was the host's private invention, never the shared PortOS baseline. That's false for most of what a visitor sees: districts, landmarks, paths and signal indicators are the shipped, install-portable PortOS world design every install resolves identically, tagged by `EIDOVERSE_MANAGED_PREFIX`. Only what falls outside that prefix is the host's own vernacular build.
- Rewrote both surfaces to name the two classes explicitly, with the prefix convention as the tell, and kept the correct facts about untrusted conversation and the promoted-foundation baseline having no visit surface of its own.
- Pinned the exact `guidance` string in a server test so a future edit can't quietly restore the universal claim, and updated the client test to assert both classes appear on the guest page.
- Per-entity tagging in the guest world snapshot would need a renderer protocol change, so this stays a text-level distinction for now (noted in a code comment), matching the issue's stated scope.

Closes #7634

## Test plan
- `server/services/eidoverseTravel.test.js` — updated to assert the new guidance string names both the shared-design prefix and the host's own vernacular, and pins the full string.
- `client/src/pages/EidoverseGuest.test.jsx` — updated to assert both classes are named on the guest page.
- Ran both suites plus the broader eidoverse server suite locally; all pass.


atomantic:

atomantic:
## Local Agent Review Summary

- **Reviewer**: `opencode` (`opencode/muse-spark-1.3-contributor-free`)
- **Status**: clean (budget `~max=1` completed)
- **Findings evaluated**:
  1. *Classification over visit-chat*: Opencode suggested that entity IDs are not visible over visit-chat. However, visiting minds observe the full world state in Eidoverse, and the guidance explicitly distinguishes shared PortOS design entities from local vernacular per #7634 acceptance criteria. Rejected as contrary to #7634 spec.
  2. *Legacy projection prefix `portos-projection-`*: Opencode suggested mentioning the legacy prefix in the guidance. `portos-projection-` is a legacy v1 prefix only used in reconciliation cleanup; all current managed designs use `EIDOVERSE_MANAGED_PREFIX` (`portos-design-v2-`). Rejected as obsolete internal detail.

All acceptance criteria from #7634 are satisfied and CI is passing.

PR #7659, opened 2026-09-18
let a promoted Eidoverse foundation be withdrawn, and propagate the retraction to peers

### LINKED ISSUE #7632: A promoted Eidoverse foundation can never be retracted: the peer offering is add-only with no tombstone

atomantic:
## Summary

A promoted Eidoverse foundation could never be recalled. The peer offering only ever **added**, and the receiver's sweep converged upward toward whatever the sender currently listed — with no way to express "and drop what I no longer list". Re-authoring already de-promoted a record locally and silently left every peer holding the retracted bytes at `layer: 'baseline'`.

That made `eidoverse.promote` — a mind-grantable capability — a way to create irrevocable cross-install state in one tool call, with no recall for a body the author later regrets: the federation-safety scan is regex-shaped and cannot recognize a person's name, a wrong attribution, or private prose.

Withdrawal is now a first-class act, propagated by tombstone.

**Ledger** (`eidoverseFoundationLedger.js`)
- `withdrawEidoverseFoundation()` de-promotes the record (body kept — withdrawal un-publishes, it does not delete) and tombstones the fingerprint it published, keyed on the content-addressed candidate digest via the existing `lib/tombstones.js`.
- Re-authoring a promoted foundation, and deleting one, tombstone it too — so the silent paths converge with the explicit one.
- Promoting clears a fingerprint's tombstone (the re-create case), and the outbound offering withholds a candidate whose fingerprint is tombstoned, so the wire can never publish and retract one fingerprint at once.
- `deleteEidoverseFoundation()` addresses an inherited copy as `{ id, originInstanceId }` rather than a raw `peer:<origin>:<id>` key, keeping the storage grammar inside the module.

**Transport** (`peerEidoverseFoundationSync.js`)
- The offering carries `tombstones: [{ fingerprint, deletedAt }]` beside `candidates`, in its **own separately capped array** — a retraction can never be displaced by the entry cap to make room for a publication.
- Both lists feed the `listHash`, so a withdrawal with no other change breaks a receiver's unchanged short-circuit instead of waiting for the next forced re-pull.
- The sweep applies retractions **first**, scoped to the peer that sent them (`inheritance.sourceInstanceId`): a fingerprint is public inside the federation the moment it is offered, so it identifies a record but authorizes nothing. A tombstone for an unheld fingerprint is a silent no-op. Peer tombstones are not persisted.

**Routes / UI**
- `POST /foundations/:id/withdraw` and `DELETE /foundations/:id` (`?originInstanceId=` for an inherited copy). Deleting a promoted local record withdraws it first, so the most natural gesture an author makes cannot orphan a peer's copy beyond recall.
- The Foundations panel grows a **Withdraw** button on promoted local records, with its own verdict headline (a withdrawal answers 200 like every gate, so without one a success rendered as "Refused — nothing moved.").

**Compatibility.** `PORTOS_SCHEMA_VERSIONS.eidoverseFoundations` → v3, wrapper only — `candidateVersion` is unchanged, since a tombstone names a candidate by the fingerprint it already had. A v2 sender omits the key and reads as "nothing to retract"; a v2 **receiver** gently skips a v3 offering and therefore keeps a copy it should have dropped. That degradation is unavoidable by construction and is documented plainly rather than papered over. No `data.reference/` seed and no migration: `tombstones` is additive on a machine-local file that reads as an empty ledger when absent.

## Test plan

- `server/services/eidoverseFoundationLedger.test.js` — 10 new cases, including a full **promote → inherit → withdraw → sweep** cycle across two simulated installs; a tombstone for an unheld fingerprint as a no-op; a retraction from a peer this install did not inherit from being refused; re-authoring tombstoning the published envelope; a packaged-but-never-promoted record recording nothing.
- `server/services/sharing/eidoverseFoundationsSync.test.js` — the `listHash` breaking on a withdrawal with an unchanged candidate list; tombstones surviving a full 500-entry candidate truncation; retraction ordering before the held-set read; a pre-#7632 peer whose payload omits the key applying without error; a failed retraction counting zero and still inheriting.
- `server/routes/eidoverseWorldRoutes.test.js` — withdraw verdict/404 mapping, delete by bare id vs. `?originInstanceId=`, and a malformed origin rejected before it can be assembled into a storage key.
- `client/.../EidoverseFoundationsPanel.test.jsx` — Withdraw offered only once promoted, never for an inherited copy, and the retraction rendered as a pass.
- Full suites green: **server 46,116 passed**, **client 12,241 passed**. `biome lint` clean.

Closes #7632


atomantic:

Issue #3606, opened 2026-08-05
Detect missing CLI-reviewer binaries server-side instead of leaving it to the agent


atomantic:
## Problem

A Review Loop reviewer whose CLI isn't installed is only discovered by the spawned follow-up agent, at review time, after PortOS has already opened the PR and burned an agent run. There is no signal at configuration time.

This shipped in the prompt layer (PR for the `agy` naming fix): the review-loop follow-up prompt now tells the agent to `command -v <binary>` and to refuse the merge if the reviewer's command is missing. That is a correct backstop but a late one — the user picked a reviewer that can never run and nothing told them.

## What to build

1. Add a shared `commandExists(cmd, args)` helper to `server/lib/` — there are currently two private copies (`server/services/localLlm.js:200`, `server/services/ollamaManager.js:89`). Re-export from the `server/lib/index.js` barrel + `README.md` row per the Module Organization rule.
2. Resolve each configured CLI reviewer's binary via `reviewerCliBinary()` (`server/lib/cosValidation.js`) and probe it, cached with a TTL so the check isn't re-run per request.
3. Surface `installed: boolean` per reviewer on the Code Review Defaults endpoint, and mark an uninstallable reviewer in `ReviewerPicker` (`client/src/components/cos/`) — label it as not-installed rather than hiding or disabling it, since a reviewer can be installed later and a federated peer may have it when this machine doesn't.

## Decisions already made

- **Warn, do not block.** Do not reject a reviewer selection or strip it from the list at spawn time. The reviewer set is federated across peers (`REVIEWER_VALUES` is a stored identity, not a machine capability), so a machine missing `agy` must not rewrite a task's configured reviewers for every other peer.
- **Probe the binary, not the provider config.** `reviewerCliBinary()` is the single source of truth for slug → command; do not re-derive it from AI-provider entries, which are a different (and optional) configuration.

## Reference

`server/lib/cosValidation.js` `REVIEWER_CLI_BINARIES` / `reviewerCliBinary()` / `describeReviewerCli()`, and the `missingCliNote` block in `server/services/agentPromptBuilder.js#buildReviewLoopFollowUpSection`.

PR #7661, opened 2026-09-18
name a reviewer that can never pass the tool-free gate, instead of stalling every PR

### LINKED ISSUE #7660: A reviewer can be pinned that can never satisfy the tool-free review gate, stalling every PR at review-blocked
## Problem

A reviewer can be pinned that is *structurally incapable* of satisfying the tool-free public-review gate, and nothing warns at selection time. The result is that every claim/swarm run publishes its PR and then stops at `review-blocked`, so no PR ever merges.

This is live right now: `provider:opencode-zen-cli` is configured as a required reviewer, and 15 open PRs (#7637–#7659) each carry the comment *"Required code review was not completed before publication. This PR is intentionally left open and will not be merged until the required review completes."* Every one of them reports `MERGEABLE` — they are held by the reviewer gate, not by a code problem.

## Evidence

Invoking the configured reviewer returns a deterministic structural refusal, not a transient failure:

```
$ jq -Rs --arg b provider:opencode-zen-cli --arg m <model> '{backend:$b,model:$m,diff:.}' < probe.diff \
    | node server/scripts/run-local-code-review.mjs
{"ok":false,"error":"This provider has no enforced tool-free review transport. Select its API mode or a supported reviewer harness."}
```

The second configured reviewer (`provider:nvidia-nim`) answers normally on the same probe, so the failure is provider-specific.

The refusal is correct and intentional. `server/services/codeReview.js:460` gates a CLI-type reviewer on `supportsPublicReviewProvider()`, which resolves to `isLocalOpencodeProvider` in `server/lib/providerVendors.js:417-419`:

```js
const isLocalOpencodeProvider = (provider) => matchOpencodeBinary(provider)
  && ['ollama', 'lmstudio'].includes(localRuntimeNamespace(provider))
  && opencodeProviderIsLocalOnly(provider);
```

OpenCode's tool posture lives in its config rather than an argv flag, so the no-tool recipe is only enforceable for a local Ollama/LM Studio-backed, local-endpoint provider. A hosted Zen gateway can never satisfy it. That security property should stay exactly as it is.

## The actual defect

`server/lib/reviewerConfig.js` validates only the *shape* of a provider reviewer token — `isProviderReviewer` at line 40 is a regex over `provider:<id>` — and never asks whether the named provider can run the posture it is being pinned for. So an impossible reviewer configuration is accepted silently and only surfaces as a run-time `review-blocked` on every PR, indefinitely, with no signal pointing at the config.

## Fix

1. In `server/lib/reviewerConfig.js`, add a resolution-time capability check for `provider:` reviewers: look the provider up and reject (or flag) one that fails `supportsPublicReviewProvider()` from `server/lib/providerVendors.js`. Keep the existing shape regex as the cheap first pass; the capability check is the second.
2. Surface the rejection where the reviewer is chosen (the reviewer-selection UI and the CoS task config) with the reason and the remedy the existing error already names — select the provider's API mode, or pick a supported harness — rather than letting it fail silently per-run.
3. Because a provider record can change after selection (its `baseURL` is relocated off-box, its namespace changes), re-check at run assembly too and report the unusable reviewer as a config fault naming the provider, distinct from today's generic `review-blocked`.
4. Tests: a `provider:` reviewer that fails the capability check is rejected at config time; a hosted-OpenCode provider is rejected while a local Ollama-backed one is accepted; a provider that passes at selection and is later relocated off-box is reported as a config fault at run assembly, not as an ordinary `review-blocked`.

## Why this is worth fixing now

`review-blocked` is the correct behavior for a reviewer that is *temporarily* unavailable — quota, transport, timeout. It is the wrong signal for one that can never work, because it is indistinguishable from a transient outage and so nobody investigates. The current state is a fully stalled merge pipeline that looks, from every individual PR, like it is merely waiting.

## Dispatch rationale

`model:medium` — routine multi-file work across a validation module, a selection surface, and run assembly; the hard reasoning (which providers can enforce the posture) is already settled in `providerVendors.js` and is being reused, not redesigned. `effort:high` — the care is in placing the check at both selection and assembly time without weakening the security predicate or turning a genuine transient outage into a config fault.

atomantic:
## Summary

A `provider:` reviewer could be pinned that is structurally incapable of satisfying the tool-free review gate. Nothing checked that at selection time, and at run time the refusal was an anonymous `{ ok: false }` — which the claim/PR review gate reads as `review-blocked`: publish, leave the PR open, wait for the outage to pass. For a reviewer that can *never* answer there is no outage to wait out, so the merge pipeline stalls indefinitely while every individual PR looks like it is merely pending.

The capability rule itself (`supportsPublicReviewProvider` → `isLocalOpencodeProvider`) is a real security property and is unchanged. What was missing is saying so out loud:

- **One source of truth for the branch.** `resolveProviderReviewTransport()` decides which transport a provider reviewer runs on (`api` / `cli`) or why it cannot run, and is shared by the dispatch and by the new capability map. A second copy of that branch would drift the moment a vendor gained or lost a no-tool recipe — and the copy that went stale would be the warning.
- **Named config faults.** Refusals now carry `REVIEWER_UNAVAILABLE` / `REVIEWER_UNSUPPORTED`, joining `NO_MODEL` in `REVIEWER_CONFIG_FAULT_CODES` (`server/lib/reviewerConfig.js` — vocabulary, so the route deciding 400-vs-502 and the service producing the codes read the same list without the route pulling in the review service's closure). `POST /api/code-review/local` answers them **400** instead of 502, and the unattended review bridge prints them as a named configuration fault rather than one more failure line.
- **Warned at selection time.** `GET /api/code-review/defaults` reports `providerReviewUnsupported`, and the reviewer picker badges such a provider **"can't review"**. Warn-only and absent-when-fine, exactly like the existing CLI-installed probe (#3606) — the reviewer list is federation-wide config a peer may satisfy, so nothing is filtered or rejected.

Ordering is deliberately preserved: an unsupported *harness* is still refused at the branch, after the effort check, because a pinned effort the model cannot do is the more specific complaint and was already this path's answer.

## Test plan

- `server/services/codeReview.providers.test.js` — refusals carry their codes; `getProviderReviewUnsupported()` names only enabled-but-incapable providers (a capable one is *absent*, not `false`; a disabled one is left to the picker's own `disabled` badge); the warning agrees with what the dispatch actually returns; an unreadable provider store reports nothing.
- `server/routes/codeReview.test.js` — both config-fault codes map to 400 while a reviewer that ran and failed stays 502; `/defaults` carries the map without touching the reviewer list. The suite deliberately leaves `isReviewerConfigFault` unmocked, so the route and the service cannot drift about which codes are faults.
- `client/src/components/cos/ReviewerPicker.test.jsx` — the badge appears for a flagged provider, says nothing for one absent from the map or when the map was never fetched, and yields to the more specific `disabled` word.
- Full suites green in this worktree: server 46,103 passed / 36 skipped; client 12,242 passed / 7 skipped.
- Verified against the real case: the bridge now prints `Reviewer configuration fault (REVIEWER_UNAVAILABLE) for provider:opencode-zen-cli` instead of an unattributed failure.

## Review status

Both required reviewers were inconclusive on this branch — `provider:opencode-zen-cli` refuses structurally (it is the very condition this PR names), and `provider:nvidia-nim` failed transport across three attempts. Recorded as `review-blocked`; this PR is intentionally left open and unmerged.

Closes #7660


atomantic:

PR #7667, opened 2026-09-19
give spawned agents a loopback API token so their own curl snippets stop 401ing

atomantic:
## Summary

On an install with the instance password set — the recommended posture — every `/api/*` path is gated. A CoS agent holds no browser cookie and `auth-sessions.json` stores only token hashes, so nothing in its environment could be turned into a credential. Both canned `curl` commands the review-loop prompt hands an agent (the local-LLM reviewer endpoint and the challenge-protocol dispute) therefore answered `401 AUTH_REQUIRED`, which an agent reads as a broken reviewer and records as a review it could not obtain.

This mints one loopback session token and injects it as `PORTOS_API_TOKEN` at the three agent spawn sites (direct CLI, TUI, CoS runner), alongside the repo-owner `GH_TOKEN` overlay they already compose. The prompts spend it with `-H "Authorization: Bearer ${PORTOS_API_TOKEN:-}"`, so the same command works with auth off (the disabled gate ignores an empty bearer) and with auth on.

- No boundary widens: an agent already runs shell on the machine that owns the session store and could call `createSession()` itself. The password itself is never exposed, only a revocable session token.
- Public-content review stages are excluded — the token is never minted for them, and `buildCliChildEnv`'s allowlist would strip it anyway.
- The reviewer prompt now also says what a 401 means: a missing credential, not a verdict, with the auth-independent stdin bridge (`server/scripts/run-local-code-review.mjs`) as the retry.
- `server/lib/localReviewBridge.js` gives that bridge path one home, so the two prompt builders that hand it to agents can't drift apart.

Documented in the root `AGENTS.md` security model and `docs/features/agent-context.md`, so future agents know they hold a credential.

## Test plan

- `cd server && npm test` — 2302 files / 46,130 tests green (run before and after the rebase onto current `main`).
- New `server/services/agentApiAuth.test.js` covers: mint when a password is set, nothing when it is not, nothing for a public-content stage, one token reused across a concurrent spawn burst, re-mint after a password rotation / an unknown cached token / a token about to expire, and a failed mint that does not take the spawn down.
- New end-to-end assertions in `server/services/authGate.test.js`: a token minted by `resolveAgentApiEnv()` passes the real gate, and the empty bearer an unprovisioned agent would send is still refused (the `:-` default is not a bypass).
- Spawn-site coverage at all three sites (`agentCliSpawning`, `agentTuiSpawning`, `agentLifecycle.spawnViaRunner`) asserting the variable reaches the child env / the runner's env delta.
- Prompt assertions for the auth header on both `curl` commands and for the 401 → bridge fallback; the review-loop file snapshots are regenerated.


atomantic:
## Review loop status: `review-blocked` — intentionally left open, not merged

Configured reviewers for this run were `provider:opencode-zen-cli` and `provider:nvidia-nim` (both required — neither carried `~opt`).

| Reviewer | Result | Detail |
|---|---|---|
| `provider:nvidia-nim` | reviewed | 2 blocking + 2 recommended findings; 1 applied, 3 rejected with reasons below |
| `provider:opencode-zen-cli` | **unavailable (config fault)** | `REVIEWER_UNAVAILABLE`: "This provider has no enforced tool-free review transport. Select its API mode or a supported reviewer harness." |

`opencode-zen-cli` is a `type: "cli"` provider, so it fails `supportsPublicReviewProvider()` and can never return a verdict through the tool-free transport — a configuration fault (the class #7660 made distinguishable from a transient outage), not a retryable failure. A required reviewer with no verdict means the required-review publication rule applies: publish, record, and stop before merge. **This PR is deliberately not merged and needs a human to either fix that reviewer's configuration (select its API mode) or merge it explicitly.**

### `provider:nvidia-nim` findings and disposition

1. **Applied** — `server/lib/agentApiToken.js`: document that `AGENT_API_AUTH_CURL_ARG` is shell text (`${VAR:-}` is parameter expansion), not a header value to hand an HTTP client.
2. **Rejected (false positive)** — "`isUsable` can throw a TypeError on a null cache entry": `!!entry && entry.expiresAt - …` short-circuits on `null`, and a record missing `expiresAt` yields `NaN > x` → `false`, which falls through to a fresh mint. No property access on `null` is reachable.
3. **Rejected (verified correct)** — "`isPublicReviewRestrictedProfile` may be inverted": it returns `true` exactly when a profile maps to a public-review posture, and `null` (an ordinary task) maps to `null` → `false`. The carve-out excludes public-content stages and nothing else.
4. **Rejected (repo convention)** — "emit a structured error instead of `console.error`": this runs outside the Express request lifecycle, where `AGENTS.md` prescribes the single-line emoji `console.error` form.

### Other gates

- Full server suite green before and after the rebase onto current `main` (2302 files / 46,130 tests).
- A four-angle `/simplify` pass ran before the PR: it moved the token coalescer onto the shared `createSingleFlight()` helper and gave the review-bridge path a single home (`server/lib/localReviewBridge.js`), which `server/services/cosTaskPrompts.js` now reads instead of rebuilding it.

atomantic:
### CI is green; the PR is still held on the required-reviewer gate

Follow-up to the review-loop comment above. CI went red twice after the first push — neither was flaky, and both are fixed here:

1. **`lib/importScoping.test.js` (the suite-wide import budget).** CI measured 114,253 against a 114,200 ceiling. Attribution before raising it: `origin/main` alone already measured **114,119** (81 of headroom), and this branch's own share is **134** — `lib/agentApiToken.js` on the 22 closures that reach a prompt builder, `lib/localReviewBridge.js` on 75 (mostly via `services/cosTaskPrompts.js`, where it replaces the same open-coded `join(PATHS.root, …)`), and `services/agentApiAuth.js` on the 12 that reach the TUI/runner spawn sites. No existing widely-reached module gained an eager edge into a subtree. The one that would have — the direct-CLI spawner, reached by far more of the suite — takes the service through an `await import()` instead, which is what `server/AGENTS.md` prescribes. The ceiling then moved 114,200 → 114,650 with the measured note the file's own convention asks for, restoring the ~400 of headroom recent entries carry.
2. **Windows-only snapshot mismatch (18 of them).** The 401-fallback line names the review bridge through `join()` + `shellQuote()`; on Windows that renders backslashes *and* gets quoted, so every committed POSIX snapshot differed. The matrix test now rebuilds that exact fragment with the same two functions and normalizes it before snapshotting — verified by mocking `path` to `path.win32`, which passes with the normalizer and reproduces all 18 mismatches without it. The prompt still hands a Windows agent the Windows path its shell needs.

Run [35412731694](https://github.com/atomantic/PortOS/actions/runs/35412731694) is green across every shard, including the three Windows ones.

**Still not merged**, for the reason in the previous comment only: `provider:opencode-zen-cli` was configured as a required reviewer and can never return a verdict through the tool-free transport. Fix that reviewer's configuration (select its API mode) or merge explicitly.

PR #7673, opened 2026-09-19
Age a quota meter against its own window, server-side

### LINKED ISSUE #7662: Age a quota meter against its own window, server-side, instead of a flat client-side hour
## Problem

A quota meter is captioned as stale on a flat one-hour threshold, decided on the client, and only ever for a reading that came from a federated peer. All three parts are the wrong shape:

1. **The threshold does not scale with the window.** `STALE_READING_MS = 60 * 60 * 1000` in `client/src/components/usage/UsageMeter.jsx` treats every window alike. A *weekly* meter read 70 minutes ago is perfectly good and now carries a "read 1h ago on <machine>" caption; a *5-hour* session meter read 55 minutes ago is ~18% of its window out of date and carries none. Tolerable age is a fraction of the window's period, not a constant.

2. **The client owns a policy it lacks the inputs for.** `server/lib/quotaWindows.js` already classifies a window's period, and adapters emit `periodHours` on a limit — but `sanitizeLimit` in `server/lib/fleetQuotas.js` drops it from the federated wire shape, and there is no client mirror of the classifier. So expressing the rule properly on the client means duplicating that classifier; the server has the inputs and already owns the card contract (`fetchedAt` semantics, `pending`, `error`).

3. **Only a stand-in reading is ever aged.** `readingAttribution` returns null unless `limit.readBy` is set, so a *local* card served from the agy/grok 5-minute scrape cache has the same "older than it looks" problem and says nothing. Staleness is a property of any reading, not a federation special case.

Found while fixing the federated-quota staleness bug (PR that added `readAt`/`readBy` captions, expiry pruning, and the sticky `pending` flag).

Move the judgment to the server and keep the client purely in formatting:

- Carry `periodHours` through the federated wire shape: add it to `sanitizeLimit` in `server/lib/fleetQuotas.js` (additive, back-compatible — an older peer simply omits it, and the existing scope-word classifier in `server/lib/quotaWindows.js` stays the fallback).
- Stamp `readAt` on **every** limit in `server/services/providerUsage.js`, not just merged ones, so a local cached reading is aged by the same rule.
- Have the server compute and stamp the verdict on each limit: `stale: true` when `now - readAt` exceeds a fraction of the window's period. Start at **10% of the window** (30 min for a 5h window, ~17h for a weekly one), floored at 5 minutes so a fast-moving window is not permanently captioned.
- `client/src/components/usage/UsageMeter.jsx` then renders the caption from `limit.stale` + `readAt` + `readByName` and holds no threshold of its own; delete `STALE_READING_MS`.

10% is the chosen number because it keeps a weekly meter quiet for most of a day (the federated case that motivated this) while catching a session meter inside one cache generation. It is a constant in one place and trivially tunable.

## Affected files

- `server/lib/fleetQuotas.js` — `sanitizeLimit`, `unifyLimits`
- `server/services/providerUsage.js` — stamp `readAt`/`stale` on local cards
- `server/lib/quotaWindows.js` — period classification (already exists; reuse)
- `client/src/components/usage/UsageMeter.jsx` — `readingAttribution`, drop `STALE_READING_MS`

## Acceptance criteria

- A limit's staleness verdict is computed once, server-side, from the window's own period.
- A locally-cached reading past the threshold is captioned the same way a peer's is.
- A peer that omits `periodHours` still merges and still gets a verdict via the scope-word fallback.
- `client/src/components/usage/UsageMeter.jsx` holds no time threshold.
- Tests: server unit coverage for the period-relative verdict (5h vs weekly, both sides of the threshold) and for the `periodHours`-absent fallback; the client test asserts rendering from the flag rather than from a clock.

atomantic:
## Summary

`STALE_READING_MS` in `UsageMeter.jsx` used one flat hour threshold for every quota window, so a weekly meter read 70 minutes ago captioned itself stale while a 5-hour session meter read 55 minutes ago (~18% out of date) said nothing. The client also had no `periodHours` on federated peer limits (so a merge fell back to scope-word classification even when the peer stated it exactly), and a purely local reading was never judged for staleness at all — only a federated stand-in with `readBy` set was.

Moves the judgment server-side, purely in formatting on the client:

- `sanitizeLimit` in `server/lib/fleetQuotas.js` carries `periodHours` through the federated wire shape (additive — an older peer simply omits it, and the existing scope-word classifier in `server/lib/quotaWindows.js` stays the fallback).
- `server/lib/quotaWindows.js` gains `staleThresholdMs`/`isLimitStale`: a reading is stale once it's older than 10% of its own window's period, floored at 5 minutes.
- `server/services/providerUsage.js` stamps `readAt` + the stale verdict on every limit of every card — not just merged ones — so a purely local (or single-instance) install ages its cached reading by the same rule.
- `unifyLimits` in `fleetQuotas.js` recomputes the verdict at merge time against the winning contribution's own `readAt`, since a peer's reading can come from a different clock than the local one.
- `client/src/components/usage/UsageMeter.jsx` renders `readingAttribution` purely from `limit.stale` + `readAt` + `readByName`; `STALE_READING_MS` is gone.

Closes #7662

## Test plan

- `server/node_modules/.bin/vitest run server/lib/quotaWindows.test.js server/lib/fleetQuotas.test.js server/services/providerUsage.test.js` — new coverage for the threshold function, wire-shape passthrough, and per-card stamping.
- `cd client && npm test -- src/components/usage/` and the full client suite — `UsageMeter.test.jsx` now asserts `readingAttribution` renders from the flag, not a clock; `UsagePage.test.jsx`'s fixture updated to carry the server-computed `stale` flag.
- Full `server` and `client` suites both green.

**Review note:** both configured local reviewers (`opencode-zen-cli`, `nvidia-nim`) were unavailable this run — one reports a configuration fault (`REVIEWER_UNSUPPORTED`), the other aborted the request on two attempts. Left the PR open per the required-review protocol rather than substituting a self-review.


atomantic:

PR #7674, opened 2026-09-19
Label the agent's loopback session so Settings > Security can show and revoke it independently

### LINKED ISSUE #7668: Label the agent's loopback session so Settings → Security can show and revoke it independently
## Context

PR #7667 gives a PortOS-spawned agent a loopback API credential: `server/services/agentApiAuth.js` mints a session with `createSession()` and injects it as `PORTOS_API_TOKEN`. That works, but it reuses a primitive shaped for a human browser session.

`server/services/auth.js` stores a session as `{ tokenHash, expiresAt }` and nothing else, so:

- the agent credential is **indistinguishable** from the user's own browser session in `data/auth-sessions.json`;
- it is revocable only by rotating the instance password or logging out everywhere (`revokeAllSessions()`), which also kicks the human's tabs — `agentApiAuth.js` listens for `sessions:revoked-all` precisely because that is the only lever;
- **Settings → Security** cannot show that an agent credential exists at all.

## The work

1. Carry a label on the persisted session record — `{ tokenHash, expiresAt, label }` — threading it through `readSessions()` / `writeSessions()` in `server/services/auth.js`. Note that `readSessions()` currently drops unknown fields on restart, so a record written by a newer install and read by an older one must stay valid (skip-on-missing-`tokenHash` is already the rule; an unknown `label` must not invalidate a record either way).
2. `createSession({ label })` (default `null` = an ordinary browser session); `agentApiAuth.js` passes `'agent'`.
3. Expose a read-only listing (`GET /api/auth/sessions` → count + label + `expiresAt`, never the hash) and a scoped revoke (`DELETE` by label, or by an opaque id), gated exactly as the rest of `/api/auth/*`.
4. Surface both in the Security settings page: "1 agent session, expires <date>" with a Revoke button that does not sign the user out.

## Decisions already made (do not re-open)

- The label is a free-text marker for display and scoped revocation, **not** an authorization dimension. The gate keeps treating every valid session identically; PortOS is single-user, and an agent already runs shell on the machine that owns the session store.
- Keep one shared agent token rather than one per run — the reuse contract in `agentApiAuth.js` exists so the store does not grow a record per spawn.

## Value

Today the only way to answer "does something other than my browser hold a session?" is to read `auth-sessions.json` and count hashes. This makes the agent credential visible and individually revocable without disrupting the human's own session.

atomantic:
## Summary

Sessions in `auth-sessions.json` were indistinguishable from a human's browser session, so Settings > Security couldn't show that an agent's loopback credential existed, and revoking it required rotating the instance password or logging out everywhere (which also kicks the human's own tabs).

- Added an optional `label` and an opaque `id` to the persisted session record, threaded through `createSession()` / `readSessions()` / `writeSessions()` with cross-version compatibility — a record missing either field stays valid, and an older install ignores fields it doesn't recognize.
- `agentApiAuth.js` now mints its loopback token with `label: 'agent'`.
- Added `GET /api/auth/sessions` (label + expiresAt only, never the token or its hash) and `DELETE /api/auth/sessions/:id` for a scoped revoke that only removes the targeted session — unlike `DELETE /api/auth/password`, which clears every session and kicks all sockets.
- Settings > Security now shows any labeled ("agent") sessions with a Revoke button.

## Test plan

- `server/services/auth.test.js` — label/id persistence, cross-version tolerance for older records missing `label`/`id`, `listSessions()` never exposing the token/hash, `revokeSessionById()` scoped to one session.
- `server/routes/auth.test.js` — `GET /api/auth/sessions` and `DELETE /api/auth/sessions/:id`, including that revoking the agent session leaves the caller's own browser session intact.
- `client/src/components/settings/SecurityTab.test.jsx` — renders the agent session with a Revoke button, revoking it removes only that entry, and the panel is hidden while auth is off.
- Ran the full server and client Vitest suites; two failures observed (`lib/validationSplit.test.js`, `PostSessionLauncher.test.jsx`) are both pre-existing flakes unrelated to this change — confirmed passing in isolation.

Closes #7668


atomantic:

PR #7715, opened 2026-09-19
fix: resolve all review feedback before merge

atomantic:
## Summary
- Treat every Do:Review finding, including non-blocking feedback and nits, as in-scope for Resolve & merge.
- Require all review threads from the posted review to be resolved before merging.

## Test plan
- server/node_modules/.bin/vitest run server/services/promptSections/reviewLifecycle.test.js server/services/promptSections/reviewLifecycleMatrix.test.js server/services/prRemediationFollowUp.test.js
- git diff --check


atomantic:
CI blocker: both the initial run and one failed-job rerun stop on the repository import-budget test, before the aggregate gate can pass. `server/lib/importScoping.test.js:861` reports 116,027 static module instantiations against a 115,900 limit. This patch adds no imports and the focused review-loop tests pass; the PR is left open for the repository-level budget failure.

PR #7753, opened 2026-09-20
feat: stop the dedicated Qwen host, and report who is using it

atomantic:
## Why

Enabling a dedicated model host from **AI Providers → Model host setup** arms four separate things, and there was no way to undo any of them:

1. the shared API queue listener,
2. `PORTOS_FLEET_LLM_ENABLED=1`, which brings it back on the next restart,
3. a Windows login-recovery scheduled task,
4. a container written with `restart: unless-stopped`, so stopping Docker does not stop it.

The host also appeared on no runtime list, so **Models → Runtimes** showed every local server you were *not* running and no sign of the one you were.

Separately, the host could report a queue *depth* but never *who* was spending the GPU.

## What changed

**Stopping.** A new **Stop hosting on this machine** action closes the queue, clears the enable marker, removes the login task, and stops **and removes** the container — removes, because a merely stopped one returns the next time the Docker engine starts. The prepared image and the ~20 GB of weights stay on disk. If Docker is not answering, the first three steps still happen and the result says which half did not. It is offered whenever there is anything to turn off, including the container-up/model-still-loading state where the host felt unstoppable. Two-click on the setup page ("Confirm stop — disconnect peers now"); one-click on the Runtimes row, like every other runtime there.

**Usage reporting.** A bounded per-client ledger at the gateway records requests, observed token counts, failures and daily totals, plus the last 50 completed requests. The panel shows what is generating right now, one row per calling machine, and a live indicator while it is busy. Callers are keyed by address (every client authenticates with the one shared host key, so the wire carries no per-peer identity) and named from the peer list or the tailnet; an address matching neither is flagged **Unrecognized** rather than hidden. Token counts come only from the model's own reply, so a row can show requests with no counts — the panel states how many carried them instead of implying the rest were free. Nothing from a request or response body is recorded.

**Runtimes.** The dedicated Qwen host now has a row on **Models → Runtimes** alongside llama.cpp, MTPLX, Slotstream, Ollama and LM Studio.

## Storage

`data/fleet-host-usage.json` — `file-primary`, bounded, machine-local and never federated (it describes who connected to *this* machine). Schema version 1, refused rather than half-read on any other version, so no migration or seed is needed. Documented in `docs/STORAGE.md` and `docs/features/fleet-llm-host.md`.

## Tests

- `server/lib/fleetHostUsage.test.js` — the usage parser against both wire shapes (including a stream whose every earlier frame carries `"usage": null`), retention, eviction-under-load, and the persistence round-trip.
- `server/services/fleetLlmGateway.test.js` — the ledger closes every entry it opens at the HTTP boundary: a completed generation, a client that hangs up while queued, and `/v1/models` polls that must not count as generations.
- `server/services/fleetLlmUsage.test.js` — peer-over-tailnet naming, the unrecognized-caller flag, restart hydration, and that the persisted document holds no request content.
- `server/services/fleetLlmHost.test.js` — the stop clears all four arming mechanisms and still disables the host when Docker refuses.
- Route and component tests for the new endpoints, the two-step stop, the usage panel and the Runtimes row.

`server/lib/importScoping.test.js` budget raised 116,100 → 116,500; measured 116,066 → 116,111 (+45), no new subtree.


atomantic (commented):
Reviewed by /do:review — 0 critical findings, 0 improvements, 0 nits.

## Highlights

The stop lifecycle, bounded inbound usage ledger, streaming accounting, persistence, route wiring, and runtime/setup UI were checked against the PR claims and changed-file contracts. No concrete defect was verified from the published diff.

## Coherence check

The implementation and tests cover the advertised stop controls, usage attribution, retention/bounding, restart hydration, and runtime visibility.

## Delegated review

Requested OpenCode pass: inconclusive because the local reviewer could not access the temporary diff file in its sandbox; no delegated finding was used.

_Generated by /do:review_

Issue #7783, opened 2026-09-20
Review coverage can silently reach zero when every configured reviewer is both optional and config-faulting


atomantic:
## Problem

`~opt` and `REVIEWER_CONFIG_FAULT_CODES` each solve a real problem, and their composition has a hole.

- `~opt` (slashdo grammar, emitted by `buildReviewWithArgs` from `optionalReviewers`) means "an inconclusive verdict from this reviewer must not block the merge." Correct for a reviewer that *usually* answers.
- `REVIEWER_CONFIG_FAULT_CODES` (`NO_MODEL`, `REVIEWER_UNAVAILABLE`, `REVIEWER_UNSUPPORTED`, added by #7660) distinguishes a reviewer that can **never** answer from one that merely failed this time, so the pipeline stops waiting out an outage that will not pass.

Compose them and an install whose **every** configured reviewer is both `~opt` and config-faulting merges every PR with **zero external review**, reporting success. `hasRequiredReviewer()` returns false, so no gate has anything to assert; each individual pass is correctly excused; the aggregate is correctly `clean`. Nothing in the run summary, the PR, or system health distinguishes "every reviewer was clean" from "no reviewer ran at all."

This is not hypothetical composition — a single-reviewer configuration is the common case, and a reviewer that is structurally unable to return a verdict on a real diff (no model pinned for a local backend; a provider with no tool-free review transport) is a state this repo already has vocabulary for. `claim-issue` v30 deliberately stopped posting a PR comment when a review is unavailable (correct: that comment was noise on a human's PR), which removed the last surface where a human would have noticed.

## Decision

**Zero-reviewer-coverage is a distinct outcome from clean, and it is loud.** It does not block the merge — `~opt` is the user's explicit instruction and must keep meaning what it says — but it must never be silent, and it must be visible *before* the next run rather than discovered by reading a diff nobody reviewed.

Rejected: making it blocking (overrides an explicit `~opt`, and strands the pipeline on exactly the outage `~opt` exists to survive). Rejected: auto-substituting a reviewer PortOS was not configured with (the executed reviewer list must never contain one the user did not ask for — slashdo holds the same line in `multi-reviewer-loop.md` step 5).

## Approach

[...]

Issue #7782, opened 2026-09-20
pregate plans from committed work only, so a mid-implementation run is a false green


atomantic:
`npm run pregate` (added alongside this issue) answers "what will CI run on this branch?" by spawning `scripts/ci-test-plan.js` and feeding its plan to `run-ci-lint.js` / `run-ci-tests.js`. Spawning the real planner is deliberate — it guarantees the local gate and CI never form two drifting opinions about which tests matter.

The cost of that choice: the planner derives its own changed-file list from `git diff <base>...HEAD`, so **it sees only committed work**. An agent that runs the gate mid-implementation gets a near-empty plan and a green result that proves nothing about the edits still in its working tree. Today pregate prints a warning and tells you to commit first, which is honest but is exactly the kind of "the tool was technically right" footgun that gets ignored under automation.

## Decision

Teach the planner to accept an explicit changed-file list, and have pregate pass the union of committed and working-tree changes. Chosen over the alternative (pregate computing the plan in-process via `buildCiTestPlan`) because that would require reimplementing `main()`'s `pathContractTests` `git grep` pass locally — a second copy of CI's selection logic, which is the drift this design exists to avoid.

## Approach

1. In `scripts/ci-test-plan.js`, extract the changed-files / tracked-files / `pathContractTests` / `appDiff` collection out of `main()` into an exported `collectPlanInputs({ baseSha, forceFull, changedFiles })`. When `changedFiles` is supplied, skip the `git diff` and use it verbatim; everything downstream (the basename `git grep`, `isRouteOnlyAppDiff`) runs against that list unchanged. `main()` calls it with no override, so CI behavior is byte-identical.
2. In `scripts/pregate.js`, build the changed-file list as `git diff --name-only --diff-filter=ACMRD <base>...HEAD` ∪ `git status --porcelain` (added/modified/renamed; deletions included, untracked-but-unignored included — an untracked new guard is exactly the case `repo-scan-guards.test.js` fails on). Pass it through. Keep the dirty-tree line, but change it from a warning to a statement of what was included.
3. `appDiff` currently reads `git diff <base>...HEAD -- client/src/App.jsx`; when App.jsx is modified but uncommitted that diff is empty and `isRouteOnlyAppDiff` would wrongly report a route-only change. Read the working-tree diff for App.jsx when it is in the override list.

## Acceptance criteria

[...]

PR #7784, opened 2026-09-20
add `npm run pregate` so CI's own test plan runs before the push, not after

atomantic:
## Summary

Adds `npm run pregate`, a local pre-push gate that runs **CI's own test plan** on the current branch.

The recurring CI failures that cost the most rounds are the ones no local command reached: an import-budget overshoot that only appears after a rebase, a server→client import-purity violation, a doubled catalog row from a union merge, a tree-wide guard that no changed file imports. Each needed a push and a ~20-minute CI round to discover, and each is static and answerable in seconds locally.

CI already decides what to run from the diff (`scripts/ci-test-plan.js`), but nothing could ask it that question outside Actions. `pregate` spawns that same planner and hands its plan to the same runners (`run-ci-lint.js`, `run-ci-tests.js`) — deliberately, so the local gate and CI can never form two drifting opinions about which tests matter. On a scoped diff it runs lint plus the selected suites in a few seconds.

Scope it is honest about:

- Plans from **committed** work only, and says so when the tree is dirty. Including the working tree needs a changed-file override in the planner — filed as #7782.
- A diff big enough to force CI's full suite runs the **always-run guards** here instead; `--full` opts into everything.
- Never runs the DB suites, the Windows job, the client build, or the boot smoke. It **names** the ones the diff implicated rather than implying it covered them.

`AGENTS.md` documents it as the pre-push gate, and the push pattern now runs it on **both sides of the rebase** — the rebase is precisely when a clean branch starts failing the budget, and the first run cannot have seen it.

Filed alongside, not in this PR: #7782 (plan from the working tree), #7783 (review coverage can silently reach zero when every configured reviewer is both optional and config-faulting), and atomantic/slashdo#264 (swarm reviewer preflight).

## Test plan

- `scripts/pregate.test.js` covers the plan→stage mapping (scope selection, selector pass-through, lint ordering, `--skip-lint`), the full-plan downgrade including a guard this checkout does not track, `--base` argument parsing, a drift guard that every `UNCOVERED_SUITES` key is a boolean field the planner still emits, and an end-to-end `--plan-only` run against this checkout.
- `npm run pregate` run against this branch: green (lint + 22 server guard files, ~4s).
- Registry guards re-run green: `direct-invocation-drift`, `repo-scan-guards`, `pre-install-entrypoints`.
- `scripts/pregate.js` is not in `FULL_TRIGGER_RULES`, so adding it does not force a full CI run; it never executes in CI.


atomantic:
CI is red on this branch from two unrelated flakes, not from the diff. Recording the evidence rather than re-running again — the one conservative re-run on the unchanged SHA (`64688b8`) is spent.

**Run 1** — `Client tests and build (1/3)` → `BeeperSettingsPanel.test.jsx > keeps Save token disabled until something is typed`:
`React state updated outside act(...) in: AttachmentMirrorCard` thrown from `client/src/test/setup.js:100`. `AttachmentMirrorCard` is not the component under test in that case — the warning is collected in a shared array and thrown in whichever `afterEach` happens to be running when it lands, which is order- and timing-dependent by construction.

**Run 2** (same SHA) — `Windows server unit tests (3/3)` → `services/videoGen/reactor.test.js > opens a portrait session for a portrait frame and uploads it pre-cropped`:
`AssertionError: promise resolved "Stats{...}" instead of rejecting` — a temp-file cleanup assertion racing Windows file-handle release.

Two runs, two different failures, in two different jobs, on two surfaces this branch does not touch. The diff is four files: `AGENTS.md`, one line in root `package.json`, and two new files under `scripts/` that never execute in CI.

Locally, in this worktree:
- `BeeperSettingsPanel.test.jsx` in isolation — 39/39 pass.
- The whole failing shard, `vitest run --shard=1/3` — 336 files, 3810 tests, 0 failures.

The full suite only runs here because root `package.json` changed; a normal scoped PR would not have exercised either surface. Both look worth filing on their own, separately from this PR.

PR #7807, opened 2026-09-20
fix(review): surface zero reviewer coverage (#7783)

### LINKED ISSUE #7783: Review coverage can silently reach zero when every configured reviewer is both optional and config-faulting

atomantic:
## Summary

- distinguish zero reviewer coverage from a clean review when every configured reviewer has a configuration fault
- persist real reviewer configuration faults, clear them after a successful review, and surface them in health details and Code Reviewers settings
- document the run-summary state across local-review prompts and preserve prompt-version integrity

## Verification

- focused server suites: 283 passed
- focused route suites: 56 passed
- focused client suites: 102 passed
- pregate: client lint and 22 selected server guard suites, 219 tests passed

The configured OpenCode reviewer is optional (`~opt`) and was unavailable because its API could not be reached; focused self-review, tests, and pregate are complete, so this does not block publication or merge under the configured policy.

Closes #7783


atomantic:
Required CI is red on the current head: the server suite's import-budget guard reports 116,547 static module instantiations against the 116,500 cap. A clean checkout of the current `origin/main` reports the same pre-existing failure, so this is not caused by #7783; the PR remains open for the baseline gate to be repaired before merge.

